PR #50, opened 2025-09-09
feature: Reduce allocations & improve performance

RicoSuter:


RicoSuter:
CleanShot 2025-09-09 at 23 14 03

PR #60, opened 2025-10-10
feature: Add high performance PropertyChangedChannel [dead lock?]

RicoSuter:
image


RicoSuter:
closed in favor of the faster queue implementation #78

PR #66, opened 2025-10-19
feature: Add support for private and protected properties in interceptor subjects

RicoSuter:
- Fixed and improved generator with property reflection for private and protected properties
- Add IsPublic and PropertyInfo? properties to subject metadata

CleanShot 2025-10-20 at 00 32 57


RicoSuter:
CleanShot 2025-10-19 at 23 51 56

PR #77, opened 2025-10-27
performance: Use object pool for registered subjects and properties

RicoSuter:
if Microsoft's DefaultObjectPool is slower in the PR #77 benchmarks, it might be because of the overhead of the IPooledObjectPolicy<T> abstraction and the Return method calling policy.Return(obj) on every return (which does a virtual dispatch).

mac: 
CleanShot 2025-11-08 at 17 48 06

win:
image


RicoSuter:
CleanShot 2025-11-08 at 17 40 42

PR #79, opened 2025-10-28
fix: Subject update with incremental changes and references

RicoSuter:
image


RicoSuter:
CleanShot 2025-11-04 at 18 29 00

PR #92, opened 2025-11-02
feature: Add property read/write lock to avoid partial reads/writes

RicoSuter:
image

----

image

----
macos
CleanShot 2025-11-04 at 15 45 29

----

Performance degradation: 

- OPC UA benchmark: 2.5% reduced throughput
- Some benchmarks: up to 20% slower


RicoSuter:
@copilot please review this PR.

Added read/write lock around low-level property field reads/writes.. does this make sense? What does this avoid? Stale reads? Dead locks?

RicoSuter:
image

PR #94, opened 2025-11-03
performance: Allocation free derived properties

RicoSuter:
Replace synchronized collections with **lock-free, copy-on-write design** using optimistic concurrency control.

## Key Changes

- **Version-based tracking** - detects concurrent modifications using monotonic counter
- **Dual-mode updates:**
  - **Sequential writes:** Replace dependencies (exact tracking, removes stale)
  - **Concurrent writes:** Merge dependencies (conservative, keeps all discovered)
- **Copy-on-write semantics** - readers get stable snapshots, no locks needed
- **Allocation-free steady state** - early exit when dependencies unchanged

## Thread Safety

- ✅ **Lock-free** using CAS (Compare-And-Swap) operations
- ✅ **ABA-safe** via version counter (wraps after 584 years @ 1B ops/sec)
- ✅ **No deadlocks** (no locks used)
- ✅ **Eventually consistent** in concurrent scenarios

## Correctness
- Prevents stale dependencies (fixes memory leak scenario)
- Handles sequential writes with exact tracking
- Handles concurrent writes with conservative merge
- Eventually consistent
- All unit tests passing

## Benchmarks

Baseline (master): 

image

image

----

image

----

### Results (see below comment):

- 40% reduced GC pressure in OPC UA benchmark
- 30% performance improvement in some benchmarks
- Up to 14x memory allocation reduction in derived property benchmark

### DIFF
--- a/src/Namotion.Interceptor.Tracking/Change/DerivedPropertyChangeHandler.cs
+++ b/src/Namotion.Interceptor.Tracking/Change/DerivedPropertyChangeHandler.cs
@@ -23,7 +23,6 @@ public class DerivedPropertyChangeHandler : IReadInterceptor, IWriteInterceptor,
     {
         if (change.Property.Metadata.IsDerived)
         {
-            // Record dependencies during initial evaluation
             StartRecordingTouchedProperties();
             var result = change.Property.Metadata.GetValue?.Invoke(change.Subject);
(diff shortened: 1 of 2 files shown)


RicoSuter:
image

RicoSuter:
image

PR #114, opened 2025-12-04
feature: Add subject and source transactions

RicoSuter:
**1. Transaction System** (`Namotion.Interceptor.Tracking`)
- `SubjectTransaction` - Captures property changes and commits atomically
- Two failure modes: `BestEffort` (apply what succeeds) and `Rollback` (all-or-nothing)
- Two locking modes: `Exclusive` (serialize all) and `Optimistic` (lock only at commit)
- Conflict detection with `FailOnConflict` behavior
- `SingleWrite` requirement for strict atomicity
- Commit timeout (default 30s) - user cancellation ignored during commit

**2. Source Transaction Writer** (`Namotion.Interceptor.Connectors`)
- Writes transaction changes to external sources (OPC UA, MQTT, etc.)
- Parallel writes to multiple sources
- Best-effort rollback on partial failures
- `WriteResult` struct for zero-allocation success path

**3. Source Ownership** (`Namotion.Interceptor.Connectors`)
- `SourceOwnershipManager` - Manages property-to-source associations with lifecycle cleanup
- `SetSource`/`RemoveSource` - Thread-safe atomic operations
- Automatic cleanup when subjects detach from object graph

### DIFF
--- a/src/Namotion.Interceptor.Connectors/Transactions/SourceTransactionWriteException.cs
+++ b/src/Namotion.Interceptor.Connectors/Transactions/SourceTransactionWriteException.cs
@@ -7,14 +7,12 @@ namespace Namotion.Interceptor.Connectors.Transactions;
 /// </summary>
 /// <remarks>
 /// <para>
-/// This exception is collected during the commit process. The transaction continues attempting to
-/// write to all sources even if some fail, maximizing the number of successful writes.
-/// A <see cref="Namotion.Interceptor.Tracking.Transactions.TransactionException"/> is thrown at the
-/// end containing all successful and failed changes.
+/// This exception is collected during the commit process and wrapped in a
+/// <see cref="Namotion.Interceptor.Tracking.Transactions.TransactionException"/> thrown at the end.
 /// </para>
 /// <para>
-/// Changes that failed to write are NOT applied to the in-process model to maintain consistency
-/// with the external system. Successfully written changes are applied even if other sources fail.
+/// In <c>BestEffort</c> mode: Successfully written changes are applied even if other sources fail.
+/// In <c>Rollback</c> mode: All successful source writes are reverted on any failure.
 /// </para>
 /// </remarks>
 public class SourceTransactionWriteException : Exception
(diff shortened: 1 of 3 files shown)


docs/sources.md
```diff
@@ -103,29 +103,137 @@ builder.Services.AddOpcUaSubjectClient<Sensor>("opc.tcp://localhost:4840", "opc"
 
 Properties with `[SourcePath("opc", "NodeName")]` map to OPC UA nodes in the external server's address space.
 
+## Source Ownership Model
+
+Each property can have **at most one source** at any time. This is the single source of truth model:
+
+- Sources claim ownership of properties by calling `SetSource(this)` during initialization
+- The `SubjectSourceBackgroundService` dispatches changes only to the owning source
```
RicoSuter: can we just run one background service (instead of per source) and this one just groups by source and calls write?

RicoSuter: each source add calls TryAddSingleton?

RicoSuter: too complicated for now

RicoSuter: to complex

src/Namotion.Interceptor.Tracking/Change/SubjectPropertyChange.cs
```diff
@@ -3,7 +3,7 @@
 
 namespace Namotion.Interceptor.Tracking.Change;
 
-public readonly struct SubjectPropertyChange
+public readonly record struct SubjectPropertyChange
```
RicoSuter: needed? and problematic for performance or even better for perf?

src/Namotion.Interceptor.Benchmark/SubjectSourceBenchmark.cs
```diff
@@ -161,7 +160,7 @@ public ValueTask WriteChangesAsync(ReadOnlyMemory<SubjectPropertyChange> changes
                 _signal.Set();
             }
 
-            return ValueTask.CompletedTask;
+            return new ValueTask<WriteResult>(WriteResult.Success());
```
RicoSuter:
```suggestion
            return new ValueTask<WriteResult>(WriteResult.Success);
```

PR #183, opened 2026-02-02
fix: ISubjectPropertyInitializer not called for dynamic properties [PLAN]

### LINKED ISSUE #181: ISubjectPropertyInitializer is not called for dynamically added properties
# Issue & repro

Having a registered `ISubjectPropertyInitializer` service as:

```cs
// Subject property initializer
public class AdditionalInfoFactory : ISubjectPropertyInitializer
{
    public void InitializeProperty(RegisteredSubjectProperty property)
    {
        if (property is { IsAttribute: false, IsSubjectReference: false, IsSubjectCollection: false, IsSubjectDictionary: false })
        {
            property.AddAttribute("isReadOnly",
                typeof(bool),
                _ => !property.HasSetter || property.Reference.Metadata.PropertyInfo?.SetMethod?.IsPublic == false,
                null);
        }
    }
}

// Register
context.AddService(new AdditionalInfoFactory());
```

and upon adding a property dynamically via `ILifecycleHandler`:

```cs

// Lifecycle handler that adds a property dynamically
public class MySubjectInitializer : ILifecycleHandler
{
    public void HandleLifecycleChange(SubjectLifecycleChange change)
    {
        if (change.IsContextAttach)
        {
            var registeredSubject = change.Subject.TryGetRegisteredSubject()
                ?? throw new InvalidOperationException("Subject not registered");

            
            if (change.Subject is IMySubjectContract)
            {
                FooType value = new();
                registeredSubject.AddProperty("myDynamicallyAddedProperty", typeof(FooType), _ => value , null);
            }
        }
    }
}

// Register
context.AddService(new MySubjectInitializer());
```

the `AdditionalInfoFactory.InitializeProperty` is not called for the dynamically added `myDynamicallyAddedProperty`.

# Expected behavior
All registered `ISubjectPropertyInitializer` implementations should be called for dynamically attached properties, just as for the statically generated properties.

RicoSuter:
Fixes #181

### Problem

When a property is added dynamically (e.g., via `ILifecycleHandler` during context attach), registered `ISubjectPropertyInitializer` services are not called for the new property, even though they are called for statically generated properties.

### Root Cause Hypothesis

The flow for dynamic properties is:
1. `RegisteredSubject.AddProperty()` calls `Subject.AddProperties()`
2. `AddPropertyInternal()` calls `Subject.AttachSubjectProperty()`
3. `AttachSubjectProperty()` calls `IPropertyLifecycleHandler.AttachProperty()` on all handlers
4. `SubjectRegistry.AttachProperty()` should call initializers

**Likely Issue:** In `SubjectRegistry.AttachProperty()`, the `TryGetRegisteredProperty()` lookup may fail for dynamic properties because the property isn't yet visible in the registry's `_knownSubjects` dictionary at the time of the call.

---

### Plan

This PR contains a detailed investigation and fix plan at `docs/plans/2026-02-02-issue-181-initializer-bug-fix.md`:

1. **Task 1:** Create regression tests that reproduce the bug
2. **Task 2:** Investigate root cause by tracing the property registration flow
3. **Task 3:** Implement fix (options: reorder registration, different lookup, or pass property directly)
4. **Task 4:** Verify all tests pass

### Files

**New:**
- `docs/plans/2026-02-02-issue-181-initializer-bug-fix.md` - Investigation and fix plan

**To be modified (based on investigation):**
- Likely `SubjectRegistry.cs` or `RegisteredSubject.cs`

### DIFF
--- a/docs/plans/2026-02-02-issue-181-initializer-bug-fix.md
+++ b/docs/plans/2026-02-02-issue-181-initializer-bug-fix.md
@@ -1,0 +1,179 @@
+# Issue #181: ISubjectPropertyInitializer Not Called for Dynamic Properties
+
+## Overview
+
+Fix the bug where `ISubjectPropertyInitializer` implementations are not invoked when properties are added dynamically via `AddProperty()` or `AddDerivedProperty()`.
+
+**GitHub Issue:** https://github.com/RicoSuter/Namotion.Interceptor/issues/181
+
+## Problem
+
+When a property is added dynamically (e.g., via `ILifecycleHandler` during context attach), registered `ISubjectPropertyInitializer` services are not called for the new property, even though they are called for statically generated properties.
+
+## Root Cause Hypothesis
+
+The flow for dynamic properties is:
+1. `RegisteredSubject.AddProperty()` calls `Subject.AddProperties()`
+2. `AddPropertyInternal()` calls `Subject.AttachSubjectProperty()`
+3. `AttachSubjectProperty()` calls `IPropertyLifecycleHandler.AttachProperty()` on all handlers
+4. `SubjectRegistry.AttachProperty()` should call initializers
+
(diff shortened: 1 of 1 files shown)


RicoSuter:
not an issue

PR #197, opened 2026-02-20
feature: Stable subject ID-based update protocol with structural mutation support

RicoSuter:
Reworks the subject-update protocol to reference subjects by **stable base62 IDs** instead of positional indices, adds full **structural mutation** support (collection / dictionary / ObjectRef insert, remove, move, replace), and makes WebSocket sync **reliable in both directions** so a server and multiple clients converge under concurrent modification, disconnects, and process kills. `ConnectorTester` is extended into a chaos/load harness that validates these scenarios.

## What this PR does

- **Identity-based update protocol.** Collection / dictionary / ObjectRef operations and item updates reference subjects by stable base62 ID (`Id` / `AfterId` / `Key`) instead of positional integers. The applier resolves items by ID and keeps, replaces, or creates instances by identity rather than CLR reference or position, so structural mutations converge correctly across participants even when applied concurrently or out of order. The legacy index-based operation model is removed.

- **Full structural mutations.** Collection insert/remove/move, dictionary insert/remove/replace, and ObjectRef set/clear, all serialized and applied by stable ID.

- **Reliable bidirectional sync.** Server to client was already reliable (monotonic per-broadcast `Sequence`, client gap detection, reconnect plus Welcome resync). This PR makes client to server symmetric: a per-connection sequence (`ConnectionSequenceTracker`), a `Resync` message, and a `ClientHeartbeat` carrying the client's last-sent sequence. The server detects in-stream and trailing-idle gaps and asks the client to re-push its complete owned state. The WebSocket client source inherits `SubjectSourceBase` (consistent with OPC UA and MQTT), which provides the receive pump and the `WriteRetryQueue` that retransmits buffered writes across reconnects.

- **Content-drop resilience under churn.** Two race-tolerant fixes prevent permanent divergence when a subject is momentarily unresolvable during heavy structural churn: an inbound bounded per-root **pending-apply buffer** recovers updates that cannot yet be resolved instead of dropping them, and outbound serialization **mints a stable ID on demand** for an unregistered subject's change instead of dropping it.

- **Quiescent-consistency backstop.** Server and client heartbeats carry a deterministic, value-aware, timestamp-insensitive **state digest** computed on demand from the live graph. A mismatch at idle routes to the existing recovery (reconnect/Welcome server to client, `Resync` client to server). It is O(1) on the wire, computed only at idle, false-positive-safe, and replaces an earlier unbounded structural-hash shadow.

- **Correct subject lifecycle during apply.** The lifecycle batch scope keeps a subject attached and registered while it moves between structural properties within a single update, so no transient detach can drop a concurrent write, and fallback context inheritance is preserved.

## Protocol model changes

```
SubjectCollectionOperation (removed; folded into item updates)

SubjectPropertyItemUpdate:
  - Index (object)  -> Id (string)     // stable subject ID (required)
  - Id (string?)    -> Key (string?)   // dictionary key (null for collections)

SubjectUpdate:
  - Root (string)   -> Root (string?)  // nullable; null = partial update without a root
```

## Code map

| Project | Key code | Purpose |
|---|---|---|
| **Connectors** | `Updates/SubjectUpdate`, `SubjectPropertyItemUpdate`, `SubjectPropertyUpdate(Kind)` | ID-based protocol model (replaces index-based ops) |
| | `Updates/Internal/SubjectUpdateApplier`, `SubjectItemsUpdateApplier`, `SubjectUpdateBuilder`, `SubjectUpdateFactory`, `SubjectUpdateApplyContext` | resolve and serialize subjects by stable ID; lazy-mint IDs for unregistered subjects |
| | `Updates/Internal/PendingApplyBuffer` (new) | recover momentarily-unresolvable inbound updates instead of dropping them |
| | `Updates/SubjectUpdateExtensions` | per-subject apply entry point and diagnostic counters |
| | `ChangeQueueProcessor` | shutdown-flush fix |
| | removed: `SubjectCollectionOperation(Type)`, `Internal/CollectionDiffBuilder` | legacy positional-index diffing |
| **WebSocket** | `Client/WebSocketSubjectClientSource` | inherits `SubjectSourceBase` (receive pump + write-retry queue); outbound sequence stamping; digest compare to reconnect |
| | `Server/ConnectionSequenceTracker` (new) | per-connection inbound sequence and gap detection |
| | `Server/WebSocketClientConnection`, `WebSocketSubjectHandler`, `WebSocketSubjectServer` | gap to `Resync`; idle heartbeat carrying the state digest; register-before-Welcome |
| | `Protocol/ResyncPayload` (new), `ClientHeartbeatPayload` (new), `HeartbeatPayload`, `MessageType` | `Resync` and `ClientHeartbeat` messages; `Digest` field on heartbeats |
| | `Internal/StateDigest` (new) | deterministic value-aware on-demand digest backstop |
| | `Serialization/JsonWebSocketSerializer` | expose JSON options so digest hashing matches the wire format |
| **Tracking** | `Lifecycle/LifecycleInterceptor`, `ContextInheritanceHandler` | batch scope keeps subjects attached and registered while moving within one apply; preserves context inheritance |
| **ConnectorTester** | `Engine/Verification/*`, `Snapshot/*`, `Logging/CycleLoggerProvider`, `appsettings.websocket-chaos.json` | chaos/load harness: independent value/structural mutation rates, per-cycle CSV recording, ID-aware JSON snapshot diffing on convergence failure |
| **Tests** | Connectors.Tests (`StableIdApplyTests`, `StableIdCollectionTests`, `ReconnectionConvergenceTests`, `DetachedSubjectUpdateDropTests`, `PendingApplyBufferTests`, verified snapshots); WebSocket.Tests (`ConnectionSequenceTrackerTests`, `StateDigestTests`, `ClientDigestRecoveryTests`, `SequenceNumberTests`); `Tracking.Tests/Lifecycle/BatchScopeTests`; `Registry.Tests/ContextRetentionLeakTests` | unit and integration coverage for ID-based apply, structural convergence, bidirectional reliability, digest recovery, batch-scope lifecycle, and retention (weak-reference liveness) |
| **docs/** | `connectors-subject-updates.md`, `connectors-websocket.md`, `registry.md` | stable-ID protocol, reliability and state-digest net, churn robustness, registry model |

## Validation

Full unit suite green; WebSocket suite 143/143. A sustained `websocket-chaos` run (85+ cycles across all profiles including process kills and forced disconnects, ~58k value and ~47k structural mutations per cycle) converges every cycle with a flat managed heap. The state-digest backstop stays idle because the race-tolerant layer keeps both sides converged on its own.

## Follow-up (optional)

Layered-PR split (core apply pipeline / protocol model / WebSocket transport plus tester) if a smaller review surface is preferred.


RicoSuter:
# Benchmark Comparison

**Date:** 2026-02-22 01:53:27
**Filter:** *SubjectUpdate*
**Base branch:** master
**Compared branch:** feature/websocket-structural-mutations

---

## feature/websocket-structural-mutations (current):

```

BenchmarkDotNet v0.15.5, macOS 26.3 (25D125) [Darwin 25.3.0]
Apple M4 Max, 1 CPU, 16 logical and 16 physical cores
.NET SDK 10.0.102
  [Host]     : .NET 9.0.10 (9.0.10, 9.0.1025.47515), Arm64 RyuJIT armv8.0-a
  DefaultJob : .NET 9.0.10 (9.0.10, 9.0.1025.47515), Arm64 RyuJIT armv8.0-a


```
| Method               | Mean     | Error     | StdDev    | Gen0   | Gen1   | Allocated |
|--------------------- |---------:|----------:|----------:|-------:|-------:|----------:|
| CreateCompleteUpdate | 3.721 μs | 0.0130 μs | 0.0115 μs | 0.7477 | 0.0153 |   6.13 KB |
| CreatePartialUpdate  | 1.028 μs | 0.0037 μs | 0.0035 μs | 0.3109 | 0.0019 |   2.55 KB |



---

## master branch:

```

BenchmarkDotNet v0.15.5, macOS 26.3 (25D125) [Darwin 25.3.0]
Apple M4 Max, 1 CPU, 16 logical and 16 physical cores
.NET SDK 10.0.102
  [Host]     : .NET 9.0.10 (9.0.10, 9.0.1025.47515), Arm64 RyuJIT armv8.0-a
  DefaultJob : .NET 9.0.10 (9.0.10, 9.0.1025.47515), Arm64 RyuJIT armv8.0-a


```
| Method               | Mean     | Error     | StdDev    | Gen0   | Gen1   | Allocated |
|--------------------- |---------:|----------:|----------:|-------:|-------:|----------:|
| CreateCompleteUpdate | 3.546 μs | 0.0097 μs | 0.0091 μs | 0.7057 | 0.0114 |   5.77 KB |
| CreatePartialUpdate  | 1.158 μs | 0.0033 μs | 0.0030 μs | 0.3166 | 0.0019 |   2.59 KB |

RicoSuter:
# Benchmark Comparison

**Date:** 2026-02-22 01:59:49
**Filter:** *Registry*
**Base branch:** master
**Compared branch:** feature/websocket-structural-mutations

---

## feature/websocket-structural-mutations (current):

```

BenchmarkDotNet v0.15.5, macOS 26.3 (25D125) [Darwin 25.3.0]
Apple M4 Max, 1 CPU, 16 logical and 16 physical cores
.NET SDK 10.0.102
  [Host]     : .NET 9.0.10 (9.0.10, 9.0.1025.47515), Arm64 RyuJIT armv8.0-a
  DefaultJob : .NET 9.0.10 (9.0.10, 9.0.1025.47515), Arm64 RyuJIT armv8.0-a


```
| Type                   | Method                  | Type        | Mean            | Error         | StdDev        | Gen0      | Gen1      | Gen2     | Allocated  |
|----------------------- |------------------------ |------------ |----------------:|--------------:|--------------:|----------:|----------:|---------:|-----------:|
| **SubjectSourceBenchmark** | **WriteToRegistrySubjects** | **?**           |  **2,050,818.1 ns** |   **2,337.94 ns** |   **2,186.91 ns** |         **-** |         **-** |        **-** |          **-** |
| **RegistryBenchmark**      | **AddLotsOfPreviousCars**   | **interceptor** | **33,561,774.1 ns** | **322,192.01 ns** | **301,378.61 ns** | **3266.6667** | **1400.0000** | **400.0000** | **24065313 B** |
| RegistryBenchmark      | IncrementDerivedAverage | interceptor |      4,403.6 ns |      10.81 ns |      10.12 ns |    0.0305 |         - |        - |      256 B |
| RegistryBenchmark      | Write                   | interceptor |        187.2 ns |       0.16 ns |       0.15 ns |         - |         - |        - |          - |
| RegistryBenchmark      | Read                    | interceptor |        246.7 ns |       0.35 ns |       0.33 ns |         - |         - |        - |          - |
| RegistryBenchmark      | DerivedAverage          | interceptor |        165.3 ns |       0.34 ns |       0.32 ns |         - |         - |        - |          - |
| RegistryBenchmark      | ChangeAllTires          | interceptor |      7,705.6 ns |      14.38 ns |      12.75 ns |    2.1057 |    0.1373 |        - |    17664 B |



---

## master branch:

```

BenchmarkDotNet v0.15.5, macOS 26.3 (25D125) [Darwin 25.3.0]
Apple M4 Max, 1 CPU, 16 logical and 16 physical cores
.NET SDK 10.0.102
  [Host]     : .NET 9.0.10 (9.0.10, 9.0.1025.47515), Arm64 RyuJIT armv8.0-a
  DefaultJob : .NET 9.0.10 (9.0.10, 9.0.1025.47515), Arm64 RyuJIT armv8.0-a


```
| Type                   | Method                  | Type        | Mean            | Error         | StdDev        | Gen0      | Gen1      | Gen2     | Allocated  |
|----------------------- |------------------------ |------------ |----------------:|--------------:|--------------:|----------:|----------:|---------:|-----------:|
| **SubjectSourceBenchmark** | **WriteToRegistrySubjects** | **?**           |  **1,996,130.8 ns** |   **2,394.03 ns** |   **2,239.38 ns** |         **-** |         **-** |        **-** |          **-** |
| **RegistryBenchmark**      | **AddLotsOfPreviousCars**   | **interceptor** | **33,498,115.8 ns** | **323,478.62 ns** | **302,582.10 ns** | **3312.5000** | **1437.5000** | **437.5000** | **24065372 B** |
| RegistryBenchmark      | IncrementDerivedAverage | interceptor |      4,439.7 ns |       8.37 ns |       7.83 ns |    0.0305 |         - |        - |      256 B |
| RegistryBenchmark      | Write                   | interceptor |        190.1 ns |       0.54 ns |       0.50 ns |         - |         - |        - |          - |
| RegistryBenchmark      | Read                    | interceptor |        253.0 ns |       0.41 ns |       0.38 ns |         - |         - |        - |          - |
| RegistryBenchmark      | DerivedAverage          | interceptor |        160.3 ns |       0.35 ns |       0.32 ns |         - |         - |        - |          - |
| RegistryBenchmark      | ChangeAllTires          | interceptor |      7,139.0 ns |      17.86 ns |      16.71 ns |    2.1057 |    0.1297 |        - |    17664 B |

src/Namotion.Interceptor.Connectors/ChangeQueueProcessor.cs
```diff
@@ -40,6 +40,42 @@ public class ChangeQueueProcessor : IDisposable
 
     private readonly PropertyChangeQueueSubscription _subscription;
 
+    // Tracks when the last flush sent changes (for idle detection)
```
RicoSuter: revert these changes (diag only)

src/Namotion.Interceptor.Tracking/Lifecycle/LifecycleInterceptor.cs
```diff
@@ -1,13 +1,12 @@
 using System.Collections;
 using System.Runtime.CompilerServices;
-using System.Runtime.InteropServices;
 using Namotion.Interceptor.Interceptors;
 
 namespace Namotion.Interceptor.Tracking.Lifecycle;
 
 public class LifecycleInterceptor : IWriteInterceptor, ILifecycleInterceptor
```
RicoSuter, line 8: is this here all needed? can we revert that?

src/Namotion.Interceptor.Tracking/Lifecycle/ContextInheritanceHandler.cs
```diff
@@ -20,7 +20,7 @@ public void HandleLifecycleChange(SubjectLifecycleChange change)
             {
                 change.Subject.Context.AddFallbackContext(change.Property.Value.Subject.Context);
             }
-            else if (change is { ReferenceCount: 0, IsPropertyReferenceRemoved: true })
+            else if (change is { IsContextDetach: true, IsPropertyReferenceRemoved: true })
```
RicoSuter, line 23: no regression risk? is this correct?

RicoSuter:
Open design item found while closing #475: **structural mutations cannot be serialized from the transactional source-write phase.** This branch will hit it on rebase, and it needs a design decision rather than a merge-conflict resolution.

## Why it lands here and not on master

On master, `ClaimPropertyOwnership` claims only `!p.CanContainSubjects` properties. `CreatePartialUpdateFromChanges` re-reads the model only in the subject reference, collection and dictionary branches of `ApplyPropertyChangeToUpdate`, so the owned set is the exact complement of the branches that read: the client source never receives a structural change, and never walks the graph during a write.

This branch deliberately removes that filter:

> Get all properties (including structural), filtered by PathProvider if configured. Structural properties (Collection, Dictionary, ObjectRef) must be claimed so that client-side structural mutations are forwarded to the server.

plus an `OnSubjectAttached` claim with no structural filter. So here the client source does own structural properties, `WriteChangesAsync` does receive structural changes, and the serializer does walk the graph.

The branch is 36 commits behind master and still carries the pre-#468 guard, with no `EnterCommitModelAccess` machinery. After rebase, a transaction whose commit writes a structural change to this source throws `InvalidOperationException` from `SubjectTransactionInterceptor.ReadProperty`.

Nothing exercised this yet because no Connector Tester profile sets `UseTransactions`, so the chaos runs never combined transactions with structural mutation.

## The constraint is ordering, not the guard

The commit writes sources first and applies locally second, deliberately, so a source rejection can leave the model untouched (Rollback) or partially applied (BestEffort). Two consequences during `WriteToSourcesAsync`:

1. The model still holds pre-transaction values, so any read contradicts what the same call is writing. That is what the guard rejects, as specified in `ITransactionWriter` and `docs/tracking-transactions.md`.
2. More fundamentally, a subject assigned inside the transaction **is not attached or registered yet**. `WriteProperty` captures the change and returns without calling `next`, so the lifecycle interceptor and the registry never see the attach until the local apply.

So relaxing the guard does not unblock this branch, it corrupts it silently. `ProcessSubjectComplete` bails at `SubjectUpdateFactory.cs:76` when `TryGetRegisteredSubject()` is null, but only after `GetOrCreateId` minted an id at line 69 and before `GetOrCreateProperties` at line 78: the walk emits an `ObjectRef` pointing at an id with no entry in `Subjects`. `BuildPathToRoot` starts from the same null and emits no path, so the change is not anchored under the root either. Minting a stable id on demand covers the id but cannot invent registry structure or parent links.

Pre-capturing the payload at write time does not rescue it either: at that moment the new subject is still unattached, so it has no parents and no path to root.

A scalar change is self-describing (property, old, new, timestamp, origin). A structural change is not: it carries a reference, and serializing it needs the subtree contents and the path to root, neither of which is in the snapshot.

## Two directions

**A. Make structural serialization self-sufficient from `changes`.** Keeps structural writes transactional, including all-or-nothing source semantics. Hard: the change carries a reference, not the subtree, so the snapshot would have to be extended to carry enough of the graph to serialize without walking it, and rollback would need the old subtree under the same constraint.

**B. Keep structural properties out of the transactional source-write path.** Apply them locally and let the post-commit outbound queue push them, which is what the non-transactional path already does: apply, notify, queue, serialize from an attached, registered, path-resolvable graph.

I would take B. Apply-then-publish is the only ordering in which a structural payload can be built at all, and this protocol is convergence-based (per-connection sequences, `Resync`, digest backstop) rather than two-phase, so it does not need transactional writes to converge. The cost is that structural mutations lose the transaction's all-or-nothing source guarantee, which is arguably not available for graph edits across a wire in any case: structural rollback via `ToRollbackChange` would need to re-serialize the old subtree under exactly the same constraint.

One semantic to pin down under B: a transaction mixing a scalar write and a structural mutation splits into source-first for the scalar and apply-then-queue for the structural. Less odd than it sounds, since scalars inside the newly attached subtree were never part of the transaction to begin with.

## Related

- #475 closed as not planned, with the master-side analysis.
- #476 merged: the guard's message now names the source-write boundary instead of blaming cross-thread use, `ISubjectSource.WriteChangesAsync` documents the constraint, and transactional WebSocket writes have end-to-end coverage. Worth rebasing onto, since the new message is what this case will surface.
- Whichever direction is taken, a Connector Tester profile with `UseTransactions` enabled would keep it covered.

PR #262, opened 2026-04-13
feature: Expose OPC UA client and server source surface (diagnostics, registration, current session)

RicoSuter:
Opens up the OPC UA client and server source surface so that consumers can observe runtime state and, where the connector does not cover the use case natively, reach the underlying session directly.

### Client (`IOpcUaSubjectClientSource`)

- `OpcUaClientDiagnostics Diagnostics { get; }`: connection state, subscription / monitored-item counts, reconnection history (`TotalReconnectionAttempts` = `SuccessfulReconnections` + `FailedReconnections` + `AbandonedReconnections`), polling fallback stats, read-after-write metrics. Live facade; resolve once and poll repeatedly.
- `ISession? CurrentSession { get; }`: direct access to the underlying OPC UA session for advanced scenarios the connector does not cover natively (OPC UA Method calls, Alarms & Conditions subscriptions).
- `event EventHandler<OpcUaCurrentSessionChangedEventArgs>? CurrentSessionChanged`: raised on every transition of the underlying session, including transitions to and from `null` during reconnection.
- `bool TryGetNodeId(PropertyReference, out NodeId)`: resolve the OPC UA NodeId bound to a tracked property.

### Server (`IOpcUaSubjectServer`)

- Equivalent `OpcUaServerDiagnostics`, `StandardServer? CurrentServer { get; }`, and `bool TryGetVariableNode(PropertyReference, out BaseDataVariableState)` for symmetry with the client side.

### DI Registration API

Aspire-style `Add*`/`AddKeyed*` extension methods returning `IServiceCollection`. Unnamed overloads register a non-keyed singleton; named overloads register a keyed singleton under the caller's name. Duplicate registration guards throw `InvalidOperationException`. The `OpcUaClientRegistration`/`OpcUaServerRegistration` handle types have been removed.

```csharp
// Unnamed (singleton)
builder.Services.AddOpcUaSubjectClientSource<Machine>(
    serverUrl: "opc.tcp://plc:4840", sourceName: "opc");
var source = serviceProvider.GetRequiredService<IOpcUaSubjectClientSource>();

// Named (keyed, for multiple registrations)
builder.Services.AddKeyedOpcUaSubjectClientSource<Machine>(
    name: "server1", serverUrl: "opc.tcp://plc:4840", sourceName: "opc");
var source = serviceProvider.GetRequiredKeyedService<IOpcUaSubjectClientSource>("server1");
```

### PublicApiGenerator coverage

Reconfigured all snapshot tests to include types in the `Microsoft.Extensions.DependencyInjection` namespace so that DI extension methods are now covered by public API snapshots.

## Retrieving the source

**Via standard DI:**

```csharp
var source = serviceProvider.GetRequiredService<IOpcUaSubjectClientSource>();
```

**From any property via `TryGetSource`:**

```csharp
if (property.TryGetSource(out var subjectSource) &&
    subjectSource is IOpcUaSubjectClientSource opcUaSource)
{
    ISession? session = opcUaSource.CurrentSession;
}
```

## CurrentSession lifecycle contract

- May return `null` at any time during reconnection. Always null-check before use.
- The underlying `ISession` instance changes after a manual reconnect, after a transferred-subscription failure, or after stall detection forces a reset. Never cache the reference.
- Read `CurrentSession` immediately before each use.
- For long-running operations on the session, subscribe to `CurrentSessionChanged` to recreate session-bound state on swap.

## CurrentSessionChanged

- Raised on every transition: new and previous session values are passed via `OpcUaCurrentSessionChangedEventArgs` (either may be `null`).
- Use `PreviousSession` only for synchronous local cleanup. Its transport may already be closed.
- Fires synchronously from the connector's own thread, but outside the reconnection lock and in transition order. For async work, use fire-and-forget (`_ = Task.Run(...)`) inside the handler.
- Handler exceptions are caught and logged so a misbehaving handler cannot destabilize the connector.

```csharp
opcUaSource.CurrentSessionChanged += (_, args) =>
{
    if (args.PreviousSession is not null) DisposeMyAlarmsSubscription(args.PreviousSession);

    if (args.CurrentSession is not null)
    {
        var newSession = args.CurrentSession;
        _ = Task.Run(async () =>
        {
            try { await RecreateMyAlarmsSubscriptionAsync(newSession); }
            catch (Exception ex) { /* log */ }
        });
    }
};
```

## Reconnection counters

`TotalReconnectionAttempts` counts every attempt initiated. Once attempts resolve, the total equals the sum of three terminal outcomes:

- `SuccessfulReconnections`: attempt produced a usable session.
- `FailedReconnections`: attempt threw an exception.
- `AbandonedReconnections`: attempt completed without an exception but the result was unusable.

## Documentation

Updated `docs/connectors-opcua.md`, `docs/connectors-opcua-client.md`, and `docs/connectors-opcua-server.md` for the new DI registration patterns.


src/Namotion.Interceptor.OpcUa.Tests/Integration/Testing/OpcUaTestClient.cs
```diff
@@ -28,9 +27,8 @@ public class OpcUaTestClient<TRoot> : IAsyncDisposable
 
     public IInterceptorSubjectContext Context => _context ?? throw new InvalidOperationException("Client not started.");
 
-    /// <summary>
-    /// Gets the client diagnostics, or null if not started.
-    /// </summary>
+    public IOpcUaSubjectClientSource? Source { get; private set; }
+
     public OpcUaClientDiagnostics? Diagnostics { get; private set; }
```
RicoSuter: remove and access via source?

src/Namotion.Interceptor.OpcUa.Tests/Integration/Testing/OpcUaTestServer.cs
```diff
@@ -25,7 +25,9 @@ public class OpcUaTestServer<TRoot> : IAsyncDisposable
     private int _disposed;
 
     public TRoot? Root { get; private set; }
-    public string BaseAddress => _baseAddress;
+
+    public IOpcUaSubjectServer? Server { get; private set; }
+
     public OpcUaServerDiagnostics? Diagnostics { get; private set; }
```
RicoSuter: remove and access via server?

src/Namotion.Interceptor.OpcUa.Tests/OpcUaCurrentSessionChangedTests.cs
```diff
@@ -0,0 +1,74 @@
...
+    public void WhenInvoked_ThenEventArgsCarryPreviousAndCurrent()
+    {
+        // Arrange
+        var source = CreateSource();
+        OpcUaCurrentSessionChangedEventArgs? captured = null;
+        source.CurrentSessionChanged += (_, args) => captured = args;
+
+        // Act - drive the event with both sides null; the contract is that the handler
+        // sees exactly the values the connector passed.
+        source.OnCurrentSessionChanged(null, null);
```
RicoSuter: shouldnt we pass in actual objects here?

src/Namotion.Interceptor.OpcUa/Client/Connection/SessionManager.cs
```diff
@@ -556,24 +556,34 @@ internal async Task DisconnectTransportAsync(CancellationToken cancellationToken
...
+    /// consumers perform synchronous local cleanup on the previous session reference.
+    /// </remarks>
     private void AbandonCurrentSession()
     {
         Debug.Assert(Monitor.IsEntered(_reconnectingLock), "AbandonCurrentSession must be called inside _reconnectingLock.");
 
-        var session = Volatile.Read(ref _session);
-        if (session is not null)
+        var oldSession = Volatile.Read(ref _session);
+        SetSession(null);
```
RicoSuter: correct to set it null here already? doesnt this break other places?

src/Namotion.Interceptor.OpcUa/Client/IOpcUaSubjectClientSource.cs
```diff
@@ -0,0 +1,75 @@
+using System.Diagnostics.CodeAnalysis;
+using Microsoft.Extensions.Hosting;
+using Opc.Ua;
+using Opc.Ua.Client;
+
+namespace Namotion.Interceptor.OpcUa.Client;
+
+/// <summary>
```
RicoSuter, line 8: maybe we should compact the xml docs (it's a bit too much) while not losing information

src/Namotion.Interceptor.OpcUa.Tests/Integration/Testing/OpcUaTestClient.cs
RicoSuter: Done. Removed the `Diagnostics` property; callers now use `client.Source!.Diagnostics`.

src/Namotion.Interceptor.OpcUa.Tests/Integration/Testing/OpcUaTestServer.cs
RicoSuter: Done. Removed the `Diagnostics` property; callers now use `server.Server!.Diagnostics`.

src/Namotion.Interceptor.OpcUa.Tests/OpcUaCurrentSessionChangedTests.cs
RicoSuter: Done. The test now uses `Mock<ISession>` instances and `Assert.Same` to verify the handler receives the exact references the connector passed.

src/Namotion.Interceptor.OpcUa/Client/Connection/SessionManager.cs
RicoSuter: Reverted to masters ordering (kill first, null last). The reordering was originally needed so the synchronous event handler could observe a live transport on the previous session. Since the event now fires outside the lock via `DrainPendingSessionTransitions`, that constraint is gone and the master ordering is the safer minimal-diff choice.

src/Namotion.Interceptor.OpcUa/Client/IOpcUaSubjectClientSource.cs
RicoSuter: Compacted the XML docs across the three new public types (`IOpcUaSubjectClientSource`, `IOpcUaSubjectServer`, `OpcUaCurrentSessionChangedEventArgs`) and the markdown sections (Diagnostics, Direct Session Access, session swap guidance). Kept the load-bearing details: Flow B/C cross-references, why method-call consumers do not need the event, the fire-and-forget guidance for async work, and the lock/ordering contract.

PR #283, opened 2026-05-04
fix: abandon SDK-transferred OPC UA sessions after server restart [BLOCKED - investigating root cause]

RicoSuter:
## Summary

- Always abandon SDK-transferred OPC UA sessions after server restart instead of trusting the transferred subscription
- The health check loop (5s interval) triggers a full manual reconnection with fresh subscriptions

## Problem

The OPC UA SDK's `SessionReconnectHandler` transfers subscriptions to a new session after server restart. Under concurrent write load (chaos testing), the transferred session's response dispatcher confuses `PublishResponse` with `WriteResponse`:

```
System.InvalidCastException: Unable to cast object of type 'Opc.Ua.PublishResponse' to type 'Opc.Ua.WriteResponse'.
   at Namotion.Interceptor.OpcUa.Client.OutboundWriter.WriteChangesAsync(...)
```

This causes:
1. **Write failures** — outbound writes get cast exceptions
2. **Silent notification loss** — publish responses are consumed by the wrong handler, subscription callbacks never fire

Clients appear connected but never receive updated values. Convergence fails after the full 5-minute timeout. Observed after 331 successful chaos cycles (server-only profile, ~0.3% failure rate).

## Investigation Status — BLOCKED

Investigating whether this is an OPC UA SDK regression in 1.5.378.134 vs 1.5.378.65.

### Possible causes

| Factor | 1000+ cycles (working) | 332 cycles (failing) |
|--------|----------------------|---------------------|
| OPC UA SDK | **1.5.378.65** | **1.5.378.134** |
| SubjectPropertyWriter race (PR #213) | Present (accidental mitigation) | Fixed |
| SDK reconnect handler | Old behavior | **PR #3471** changes `RecreateAsync` path |

### Key findings

**SDK version changed:** The 1000+ cycle runs used SDK **1.5.378.65**. Current code uses **1.5.378.134**. Between these versions, [OPCFoundation/UA-.NETStandard#3471](https://github.com/OPCFoundation/UA-.NETStandard/pull/3471) ("Fix session reconnect handler") changed the `RecreateAsync` call path. When `transportChannel` is null, it now calls `RecreateAsync(current)` without the transport channel, creating a brand new channel. This may introduce the response dispatcher confusion under concurrent writes.

**PR #213 race fix removed accidental mitigation:** Before PR #213, concurrent subscription notifications could interleave with the buffer replay, accidentally correcting stale values. The race fix blocks concurrent writes during replay, removing this accidental correction.

**SDK 1.5.378.65 has a memory leak** that's fixed in 1.5.378.134. We can't stay on the old SDK long-term.

### Verification steps

Chaos tester (`--launch-profile opcua-chaos --configuration Release`), clients at 500 mutations/sec.

| # | What | SDK version | Result |
|---|------|-------------|--------|
| 1 | Old SDK, 100/sec clients | 1.5.378.65 | 262 cycles passed, 0 failures (stopped) |
| 2 | Old SDK, 500/sec clients | 1.5.378.65 | **249 cycles passed, 0 failures** (stopped — memory leak in old SDK) |
| 3 | New SDK, 500/sec clients | 1.5.378.134 | ⏳ Running |
| 4 | This fix + new SDK | 1.5.378.134 | 39 cycles passed, stopped early |

**Combined old SDK results: 511 cycles, 0 failures.** If step 3 fails quickly, the SDK upgrade is confirmed as the cause.

### Next steps

- If step 3 confirms SDK regression → file issue on [OPCFoundation/UA-.NETStandard](https://github.com/OPCFoundation/UA-.NETStandard) with repro
- Consider whether to stay on this workaround (abandon transferred sessions) or fix the SDK interaction
- May need to use local SDK source reference (`UseLocalOpcUaProjects=true`) to patch the issue

## Test plan

- [x] All 4 `ServerRestart*` integration tests pass with fix
- [x] Full OPC UA test suite passes (240/240) with fix
- [ ] Confirm SDK regression with step 3
- [ ] Run final fix for 300+ cycles to validate

### DIFF
--- a/src/Namotion.Interceptor.OpcUa.Tests/Integration/OpcUaReconnectionTests.cs
+++ b/src/Namotion.Interceptor.OpcUa.Tests/Integration/OpcUaReconnectionTests.cs
@@ -129,6 +129,14 @@ public class OpcUaReconnectionTests
                     message: "Data should flow after restart");
                 logger.Log($"Client received: {client.Root.Name}");
 
+                // Wait for the manual reconnection to complete. The SDK reconnection
+                // is abandoned (transferred sessions are unreliable after server restart),
+                // and the health check triggers a fresh manual reconnection within ~5s.
+                await AsyncTestHelpers.WaitUntilAsync(
+                    () => client.Source!.Diagnostics.SuccessfulReconnections >= 1,
+                    timeout: TimeSpan.FromSeconds(30),
+                    message: "Manual reconnection should complete");
+
                 // Verify metrics
                 var finalAttempts = client.Source!.Diagnostics.TotalReconnectionAttempts;
                 var successfulReconnections = client.Source!.Diagnostics.SuccessfulReconnections;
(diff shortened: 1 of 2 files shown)


RicoSuter:
Closing: 768+ cycles at 500/sec client writes on master with SDK 1.5.378.134 show this is a rare event (~1 in 500-1000 cycles), not a regression. Created issue #284 to track the root cause investigation and a separate PR with diagnostic improvements for when it recurs.

Issue #287, opened 2026-05-05
chaos: Subscription silent death after OPC UA session transfer under chaos (InvalidCastException)


RicoSuter:
## Summary

Rare convergence failure (~1 in 500-1000 chaos cycles) where a client's subscription stops delivering notifications after an SDK session transfer following server restart. The client appears connected but never receives updated values, causing convergence timeout.

## Observed behavior

- **Cycle 332** of the OPC UA chaos tester (server-only chaos profile, Release mode) failed after 331 successful cycles
- Client-b had stale values across all 31 subjects (91/124 properties) after 5 minutes of convergence
- Client-a converged fine (it lost connection mid-convergence and did a clean manual reconnection)
- 61 `InvalidCastException: Unable to cast 'Opc.Ua.PublishResponse' to 'Opc.Ua.WriteResponse'` in the failing cycle

### Mutate phase (07:40:16 – 07:41:16)

Server-only chaos profile: 3 server kills during the 1-minute mutate phase.

| Time | Event |
|------|-------|
| 07:40:16 | Mutate phase starts. Server kill #1 (3.8s) |
| 07:40:26 | Both clients reconnect (sessions 1700652017, 1700652106). Full state syncs complete. |
| 07:40:40 | Server kill #2 (10.1s) |
| 07:40:49-50 | Both clients reconnect (sessions 1937852079, 1937852168). Full state syncs complete. |
| 07:40:50+ | **61 InvalidCastException** (`PublishResponse` → `WriteResponse`) in `OutboundWriter.WriteChangesAsync` — client writes fail and get queued for retry |
| 07:41:04 | Server kill #3 (8.9s), recovers ~07:41:13 |
| 07:41:14.439 | **Client A** reconnects → session 34667842, subscription transferred |
| 07:41:14.854 | **Client B** reconnects → session 34667931, subscription transferred |
| 07:41:15.390 | Client A full state sync completes (reads 124 nodes) |
| 07:41:16.240 | Last failed client write (queued for retry) |
| 07:41:16.299 | **Convergence phase starts** — mutations paused |
| 07:41:16.871 | Client B full state sync completes (reads 124 nodes) |

### Convergence phase (07:41:16 – 07:46:36)

| Time | Event |
|------|-------|
| 07:41:16 – 07:43:44 | **2.5 minutes of silence** — no log activity. Subscription should be delivering remaining notifications to both clients. |
[...]

PR #289, opened 2026-05-05
feature: Update OPC UA SDK package to upstream/master (post-1.5.378.134)

### LINKED ISSUE #285: chaos: OPC UA notification loss on client-b during rapid client-a reconnection
## Summary

After 317 successful chaos cycles, cycle 318 (client-a-only chaos profile) failed to converge. Client-b permanently lost OPC UA notifications from the server for 3 out of 4 properties on the root node, despite never being disconnected itself.

## Environment

- OPC UA SDK: 1.5.378.134
- Profile: opcua-chaos (1000 mutations/sec server, 100/sec per client, 2 clients)
- Chaos: client-a killed 3 times in rapid succession within a 60-second mutate phase

## Divergence

```
Root node comparison:
Property                 Server         Client-a         Client-b
------------------------------------------------------------------
StringValue            0120162a         0120162a         01200db3
DecimalValue          188800.96        188800.96        188800.96
IntValue               18880152         18880152         18877879
LongValue              18880138         18880138         18878862

Timestamps:
StringValue        08:17:01.688     08:17:01.688     08:16:59.620
DecimalValue       08:17:01.738     08:17:01.738     08:17:01.738
IntValue           08:17:01.792     08:17:01.792     08:16:59.625
LongValue          08:17:01.780     08:17:01.780     08:17:00.566
```

Client-a converged with server (full state read on reconnect). Client-b is 2167 mutations behind on StringValue, but DecimalValue matches exactly (same value + timestamp). This selective property loss suggests OPC UA monitored item queue overflow on the server side.

## Timeline

| Time | Event |
|------|-------|
| 08:16:01 | 1st kill on client-a, session closed |
| 08:16:02 | client-a reconnects, flushes 548 queued writes |
| 08:16:13 | client-a reconnects again, flushes 1000 queued writes |
| 08:16:24 | 2nd kill on client-a |
| 08:16:28 | Session created but immediately closed (failed restart) |
| 08:16:33 | client-a reconnects, flushes 947 queued writes |
| 08:16:43 | 3rd kill on client-a |
| 08:16:54 | Multiple failed restart attempts (session created + closed immediately) |
| 08:16:54+ | client-a write queue hits 1000 capacity, starts dropping writes |
| 08:16:59 | Another failed session creation (created + closed immediately) |
| 08:17:01 | Mutate phase ends, convergence starts |
| 08:17:04 | client-a finally reconnects, reads full state, session stable |
| 08:22:21 | 5-minute convergence timeout expires, FAIL |

## Key Observations

1. **Client-b was never disconnected** - no chaos, no reconnection, no errors logged for client-b
2. **Selective property loss** - DecimalValue arrived fine (08:17:01.738), but StringValue/IntValue/LongValue stopped at 08:16:59-08:17:00
3. **Server was busy** - 6 sessions created + 5 sessions closed in 60 seconds for client-a reconnections, while publishing 1000 mutations/sec
4. **Not recovered in 5 minutes** - the lost notifications never arrived, suggesting permanent loss (not just delay)

## Hypothesis

The OPC UA server's monitored item queue for client-b's subscription overflowed during the period of heavy session churn from client-a reconnections. The server was allocating resources for rapid session create/destroy cycles while simultaneously publishing at 1000 mutations/sec. Some publish responses for client-b were dropped or delayed past the queue limit.

The selective nature (DecimalValue arrived, others didn't) suggests per-item queue overflow rather than a full subscription failure.

## Relationship to PR #283

This is a **different issue** from PR #283 (which addresses SDK session transfer after server restart causing PublishResponse/WriteResponse cast confusion). Key differences:
- No server restart in this cycle
- No InvalidCastException
- Client-b (the affected client) was never disconnected
- Loss is on the receiving side (notifications from server), not the writing side

Both issues may share a common root in the OPC UA SDK's session management under concurrent load (SDK 1.5.378.134).

## Reproduction

```bash
dotnet run --project src/Namotion.Interceptor.ConnectorTester --launch-profile opcua-chaos --configuration Release
```

Run until failure (occurred at cycle 318 in this run, ~7 hours). The failure log is at `logs/cycle-0318-FAIL-2026-05-05T08-16-01.log`.

RicoSuter:
## Summary

- Adapt to breaking API changes in OPCFoundation/UA-.NETStandard master (ahead of 1.5.378.134)
- Originally proposed as the way to deliver the `#3734` notification-loss fix; this is now obsolete because `#331` ships the same fix on the release line without API churn.
- Uses UseLocalOpcUaProjects=true for now; will switch to NuGet when next SDK version is published

## Relationship to #331

`#331` bumps the release-line NuGet to `1.5.378.145`, which already contains the `#3734` fix and is API-compatible. `#331` supersedes this PR as the way to deliver `#3734` to users. This PR remains as the tracker for the larger move to master's newer API (`NodeId` struct, removed collection types, C# 14 extension syntax, `WrappedValue.AsBoxedObject`, `TelemetryContext` constructor). It stays blocked on #290 and on the underlying notification-loss family also seen in #285.

## Known Issue

Chaos testing on upstream/master (OPCFoundation/UA-.NETStandard@2806e35c) found a regression: tiny notification loss (1-3 properties, 25ms) at mutate/converge boundaries. Reproduced at cycle 133 and again at cycle 717 (see #290). The same scenario did not reproduce on NuGet 1.5.378.134 in 1050+ cycles. The release-line `1.5.378.145` soak gating `#331` did reproduce a broader #285-style failure once at cycle 698, but with flat heap throughout, so the master-only heap drift hypothesis still holds for #290's specific tiny-divergence variant.

## Key Changes

- Replace removed collection types with ArrayOf<T> and List<T> + .ToArrayOf()
- Use WrappedValue.AsBoxedObject() instead of DataValue.Value
- Convert SourceTimestamp with .ToDateTimeOffset()
- Pass TelemetryContext to ApplicationConfiguration constructor (prevents null ref on discovery)
- Handle NodeId as struct (.IsNull, default instead of null)
- Add LangVersion=preview for C# 14 extension syntax from SDK

## Test plan

Build / API surface:
- [x] Build succeeds with local SDK reference
- [ ] Re-verify build against a released SDK version (post-1.5.378.145) that contains the master changes
- [ ] Drop `UseLocalOpcUaProjects=true` and switch to a NuGet reference
- [ ] Drop `LangVersion=preview` if the SDK ships with stable C# language features

Chaos soak on the master-line SDK (run via `dotnet run --project src/Namotion.Interceptor.ConnectorTester --launch-profile opcua-chaos --configuration Release`; the `opcua-chaos` profile rotates through client-only, server-only, and full-chaos sub-profiles so a single run exercises all three regressions below):

- [ ] **#290** does not reproduce. This is the merge blocker for this PR. Prior repros at cycles 133 and 717. Target a clean run comparable to the 1050+ cycle baseline cleared by 1.5.378.134.
- [ ] **#285** does not reproduce at a rate worse than the release line. The release-line baseline is currently 2 known repros across 1.5.378.134 (cycle 318) and 1.5.378.145 (cycle 698); master must not be worse. Note: `#3734` was originally believed to fix #285 but does not (see #285 / #331).
- [ ] **#287** does not reproduce under the server-only sub-profile (subscription silent death after server-restart session transfer; observed once in 2200+ cycles, treat as a stability check rather than a blocker).
- [ ] HeapMB in `cycles.csv` does not drift upward across cycles. Master showed 88 MB to 216 MB over 717 cycles; release line is flat at 98.8 MB. Sustained growth is a merge blocker because it likely causes #290.

### DIFF
--- a/src/Namotion.Interceptor.OpcUa/Client/Connection/SessionManager.cs
+++ b/src/Namotion.Interceptor.OpcUa/Client/Connection/SessionManager.cs
@@ -173,7 +173,7 @@ internal sealed class SessionManager : IAsyncDisposable, IDisposable
         var endpointConfiguration = EndpointConfiguration.Create(application.ApplicationConfiguration);
         var serverUri = new Uri(configuration.ServerUrl);
 
-        EndpointDescriptionCollection endpoints;
+        ArrayOf<EndpointDescription> endpoints;
         try
         {
             using var discoveryClient = await DiscoveryClient.CreateAsync(
@@ -181,7 +181,7 @@ internal sealed class SessionManager : IAsyncDisposable, IDisposable
                 serverUri,
                 endpointConfiguration, ct: cancellationToken).ConfigureAwait(false);
 
-            endpoints = await discoveryClient.GetEndpointsAsync(null, cancellationToken).ConfigureAwait(false);
+            endpoints = await discoveryClient.GetEndpointsAsync(default, cancellationToken).ConfigureAwait(false);
         }
         catch (Exception ex) when (ex is not OperationCanceledException)
         {
@@ -212,7 +212,7 @@ internal sealed class SessionManager : IAsyncDisposable, IDisposable
             identity: configuration.CreateUserIdentity != null
                 ? await configuration.CreateUserIdentity(cancellationToken).ConfigureAwait(false)
                 : new UserIdentity(),
-            preferredLocales: null,
+            preferredLocales: default,
             cancellationToken).ConfigureAwait(false);
 
         var newSession = sessionResult as Session
--- a/src/Namotion.Interceptor.OpcUa/Client/Connection/SubscriptionManager.cs
+++ b/src/Namotion.Interceptor.OpcUa/Client/Connection/SubscriptionManager.cs
@@ -149,7 +149,7 @@ internal class SubscriptionManager : IAsyncDisposable
         }
     }
 
-    private void OnFastDataChange(Subscription subscription, DataChangeNotification notification, IList<string> stringTable)
+    private void OnFastDataChange(Subscription subscription, DataChangeNotification notification, ArrayOf<string> stringTable)
     {
         if (_shuttingDown)
         {
@@ -175,8 +175,8 @@ internal class SubscriptionManager : IAsyncDisposable
                     changes.Add(new PropertyUpdate
                     {
                         Property = property,
-                        Value = _configuration.ValueConverter.ConvertToPropertyValue(item.Value.Value, property),
-                        Timestamp = item.Value.SourceTimestamp
+                        Value = _configuration.ValueConverter.ConvertToPropertyValue(item.Value.WrappedValue.AsBoxedObject(), property),
+                        Timestamp = item.Value.SourceTimestamp.ToDateTimeOffset()
                     });
                 }
             }
--- a/src/Namotion.Interceptor.OpcUa/Client/OpcUaClientConfiguration.cs
+++ b/src/Namotion.Interceptor.OpcUa/Client/OpcUaClientConfiguration.cs
@@ -361,7 +361,7 @@ public class OpcUaClientConfiguration
         var host = System.Net.Dns.GetHostName();
         var applicationUri = $"urn:{host}:Namotion.Interceptor:{ApplicationName}";
 
-        var config = new ApplicationConfiguration
+        var config = new ApplicationConfiguration(TelemetryContext)
         {
             ApplicationName = ApplicationName,
             ApplicationType = ApplicationType.Client,
--- a/src/Namotion.Interceptor.OpcUa/Client/OpcUaClientDiagnostics.cs
+++ b/src/Namotion.Interceptor.OpcUa/Client/OpcUaClientDiagnostics.cs
@@ -26,7 +26,7 @@ public class OpcUaClientDiagnostics
     /// <summary>
     /// Gets the current session identifier, or null if not connected.
     /// </summary>
-    public string? SessionId => _source.SessionManager?.CurrentSession?.SessionId?.ToString();
+    public string? SessionId => _source.SessionManager?.CurrentSession?.SessionId.ToString();
 
     /// <summary>
     /// Gets the number of active OPC UA subscriptions.
--- a/src/Namotion.Interceptor.OpcUa/Client/OpcUaSubjectClientSource.cs
+++ b/src/Namotion.Interceptor.OpcUa/Client/OpcUaSubjectClientSource.cs
@@ -184,16 +184,18 @@ internal sealed class OpcUaSubjectClientSource : BackgroundService, IOpcUaSubjec
         for (var offset = 0; offset < itemCount; offset += batchSize)
         {
             var take = Math.Min(batchSize, itemCount - offset);
-            var readValues = new ReadValueIdCollection(take);
+            var readValuesList = new List<ReadValueId>(take);
 
             for (var i = 0; i < take; i++)
             {
-                readValues.Add(new ReadValueId
+                readValuesList.Add(new ReadValueId
                 {
                     NodeId = ownedProperties[offset + i].NodeId,
                     AttributeId = Opc.Ua.Attributes.Value
                 });
             }
+
+            var readValues = readValuesList.ToArrayOf();
 
             var readResponse = await session.ReadAsync(
                 requestHeader: null,
@@ -218,8 +220,8 @@ internal sealed class OpcUaSubjectClientSource : BackgroundService, IOpcUaSubjec
         {
             foreach (var (property, dataValue) in result)
             {
-                var value = _configuration.ValueConverter.ConvertToPropertyValue(dataValue.Value, property);
-                property.SetValueFromSource(this, dataValue.SourceTimestamp, null, value);
+                var value = _configuration.ValueConverter.ConvertToPropertyValue(dataValue.WrappedValue.AsBoxedObject(), property);
+                property.SetValueFromSource(this, dataValue.SourceTimestamp.ToDateTimeOffset(), null, value);
             }
 
             _logger.LogInformation("Updated {Count} properties with OPC UA node values.", itemCount);
@@ -504,7 +506,14 @@ internal sealed class OpcUaSubjectClientSource : BackgroundService, IOpcUaSubjec
         if (_configuration.RootName is not null)
         {
             var references = await BrowseNodeAsync(session, ObjectIds.ObjectsFolder, cancellationToken).ConfigureAwait(false);
-            return references.FirstOrDefault(reference => reference.BrowseName.Name == _configuration.RootName);
+            foreach (var reference in references)
+            {
+                if (reference.BrowseName.Name == _configuration.RootName)
+                {
+                    return reference;
+                }
+            }
+            return null;
         }
 
         return new ReferenceDescription
@@ -618,7 +627,7 @@ internal sealed class OpcUaSubjectClientSource : BackgroundService, IOpcUaSubjec
         }
     }
 
-    private async Task<ReferenceDescriptionCollection> BrowseNodeAsync(
+    private async Task<ArrayOf<ReferenceDescription>> BrowseNodeAsync(
         Session session,
         NodeId nodeId,
         CancellationToken cancellationToken)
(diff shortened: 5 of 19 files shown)


RicoSuter:
## Finding: Memory Observations on upstream/master SDK

### GC Segment Fragmentation (Not a Real Leak)

Chaos testing shows HeapMB (from `GC.GetTotalMemory(false)`) growing from 88 MB to 204 MB over 700 cycles, while NuGet 1.5.378.134 stays flat at ~84 MB.

However, a gcdump at cycle 210 revealed only **24 MB of live managed objects** on the heap. The 204 MB is committed-but-unused GC segments - fragmentation from the new Variant/immutable types allocation pattern, not a traditional memory leak.

| Metric | NuGet 1.5.378.134 (522 cycles) | Local master (700 cycles) |
|--------|-------------------------------|--------------------------|
| HeapMB (GC.GetTotalMemory) | ~84 MB (flat) | ~207 MB (growing) |
| Actual live managed heap (gcdump) | n/a | ~24 MB |

### Undisposed X509Certificate2 Instances

The gcdump at cycle 210 showed **634 `X509Certificate2`** instances with native `SafeSecCertificateHandle` handles (~3 per chaos cycle). These are likely created during secure channel establishment on each client reconnection and not disposed after the channel is closed.

```
432       634  System.Security.Cryptography.X509Certificates.AppleCertificatePal
168       634  System.Security.Cryptography.X509Certificates.X509Certificate2
 32       634  System.Security.Cryptography.X509Certificates.SafeSecCertificateHandle
```

While the managed footprint is tiny (432 bytes for 634 objects), each `X509Certificate2` holds a native OS certificate handle. On macOS these are `SecCertificateRef` handles that pin native memory. Over long-running tests (1000+ cycles), this could contribute to native memory growth and GC fragmentation since pinned objects prevent heap compaction.

This should be investigated separately - the certificates should be disposed when the secure channel is torn down during chaos recovery/reconnection.

RicoSuter:
## Finding: Potential Root Cause for Notification Loss (#290)

Deep investigation of all behavioral changes between NuGet 1.5.378.134 and upstream/master (2806e35c) identified the most likely root cause.

### MonitoredNode Lock Scope Reduction + ConcurrentDictionary Change

Two changes in `Libraries/Opc.Ua.Server/NodeManager/MonitoredItem/MonitoredNode.cs` (introduced in commits 8c008a17 and 588b7e92) work together to create a potential race during reconnection:

**Change 1 - Lock scope reduced:**

Old (`OnMonitoredNodeChanged`):
```csharp
lock (NodeManager.Lock)  // global lock, shared with subscription transfer
{
    for (int ii = 0; ii < DataChangeMonitoredItems.Count; ii++)
    {
        QueueValue(context, node, monitoredItem);
    }
}
```

New:
```csharp
lock (m_dataChangelock)  // per-node lock, NOT shared with subscription transfer
{
    foreach (var kvp in DataChangeMonitoredItems)
    {
        QueueValue(contextToUse, node, monitoredItem);
    }
}
```

**Change 2 - Collection type changed:**

Old: `List<IDataChangeMonitoredItem2> DataChangeMonitoredItems` (protected by `NodeManager.Lock`)
New: `ConcurrentDictionary<uint, IDataChangeMonitoredItem2> DataChangeMonitoredItems` (`foreach` provides a snapshot at enumeration start)

### Why This Is a Problem

In the old code, `NodeManager.Lock` was shared between:
- `OnMonitoredNodeChanged` (which reads the node value and calls `monitoredItem.QueueValue()`)
- Subscription transfer/creation operations (which add/remove MonitoredItems)

This meant notifications and subscription lifecycle were fully serialized. A value change notification would either see the complete old subscription state or the complete new subscription state - never a partial transition.

With the new `m_dataChangelock` (per-node, not shared with subscription operations), `OnMonitoredNodeChanged` and subscription transfer can now run concurrently on the same node. The `ConcurrentDictionary.foreach` takes a snapshot at enumeration start, but this snapshot can be stale by the time `QueueValue` actually reads the node value and queues it.

### Race Scenario During Server Chaos

1. Server is killed and restarts (server-only chaos profile)
2. Clients reconnect - old MonitoredItems being removed via `Remove()`, new MonitoredItems being added via `Add()`
3. Concurrently, a value change fires `OnMonitoredNodeChanged` on the server
4. The `ConcurrentDictionary` `foreach` snapshots the dictionary at enumeration start - might see the OLD MonitoredItem (about to be removed) and miss the NEW one (just being added)
5. Notification is queued to the old MonitoredItem, which is part of the old subscription being torn down
6. The new MonitoredItem never receives the notification
7. No more mutations arrive (converge phase starts) so the notification is permanently lost

### Why This Matches Our Symptoms

- Happens during chaos (reconnection) - both failures were during chaos profiles
- Affects both clients identically (server restart means both clients reconnect simultaneously)
- Loses only 1-3 notifications at the exact boundary (the narrow window during subscription transfer)
- Never recovers (the notification was queued to a dead MonitoredItem, no retry mechanism)
- Does NOT reproduce on NuGet 1.5.378.134 which uses the old `NodeManager.Lock` pattern

### Test Results

We tested by reverting the lock change in the local SDK:
```csharp
// Reverted from:
lock (m_dataChangelock)
// Back to:
lock (((CustomNodeManager2)NodeManager).Lock)
```

| Run | SDK | Lock | Cycles | Result |
|-----|-----|------|--------|--------|
| NuGet baseline | 1.5.378.134 | NodeManager.Lock (original) | 828+ | All PASS, heap flat at ~84 MB |
| Local master | upstream/master | m_dataChangelock (new) | 132 | FAIL at cycle 133 |
| Local master | upstream/master | m_dataChangelock (new) | 717 | FAIL at cycle 717 (gcdump may have triggered) |
| Local master | upstream/master | NodeManager.Lock (reverted) | 217+ | All PASS (stopped early) |

Results are **inconclusive** - the non-reverted run also went 700+ cycles before failing, so 217 cycles with the revert doesn't prove the fix works. Needs 1000+ cycles to compare fairly.

### Memory Finding (Corrected)

The apparent "memory leak" (HeapMB growing from 88 to 204 MB over 700 cycles) is **NOT a real leak**. A gcdump at cycle 210 showed only **24 MB of live managed objects**. The 204 MB reported by `GC.GetTotalMemory(false)` is committed-but-unused GC segments (fragmentation from the new Variant/immutable types allocation pattern). This affects both the reverted and non-reverted runs equally.

### Other Changes Investigated and Ruled Out

| Change | Risk | Assessment |
|--------|------|------------|
| `moreNotifications` fix (2806e35c, #3734) | N/A | Already included in tested commit |
| SessionPublishQueue lock consolidation (66f44396) | Low | Actually fixes an old race, not a regression |
| `IsReadyToTrigger`/`IsResendData` lock-free reads | Low | Volatile.Read is safe for single booleans |
| `ValueChanged` equality rewrite (c19e83b6) | Low | New code is more permissive (reports more changes, not fewer) |
| `DataChangeFilter.AreEqual` vs old `Equals` | Low | Semantically equivalent for our types |
| HiResClock timing change (562f71cf) | Low | Consistent monotonic clock |
| `value.Copy()` vs manual DataValue construction | Low | Functionally equivalent |
| Filter condition order swap (`!ignoreFilters` first) | None | Short-circuit optimization only |

### How to Reproduce and Verify

**Setup:**
1. Clone the SDK: `git clone https://github.com/OPCFoundation/UA-.NETStandard`
2. Checkout upstream/master: `git checkout upstream/master`
3. Apply the lock revert in `Libraries/Opc.Ua.Server/NodeManager/MonitoredItem/MonitoredNode.cs` line 240:
   ```csharp
   // Change from:
   lock (m_dataChangelock)
   // To:
   lock (((CustomNodeManager2)NodeManager).Lock)
   ```
4. In the Namotion.Interceptor repo on `feature/opcua-sdk-upgrade-master` branch, ensure `UseLocalOpcUaProjects=true` in `src/Namotion.Interceptor.OpcUa/Namotion.Interceptor.OpcUa.csproj`
5. Run: `dotnet run --project src/Namotion.Interceptor.ConnectorTester --launch-profile opcua-chaos --configuration Release`
6. Let it run for 1000+ cycles (about 22 hours)
7. Compare against a run WITHOUT the lock revert to see if the notification loss (#290) reproduces

**What to look for:**
- `logs/cycles.csv` for any FAIL rows
- `logs/cycle-*FAIL-*.log` for failure details
- HeapMB column for memory trends (expect ~200 MB from GC fragmentation, not a real leak)

PR #303, opened 2026-05-12
fix: Ensure property write timestamp matches published change event

RicoSuter:
## Summary
Fixes a within-write timestamp drift bug: the timestamp stored on a property differed from the timestamp published in change events, because storage and `PropertyChangeQueue` each called `GetTimestampFunction()` separately and got different `UtcNow` values. This caused OPC UA snapshot mismatches between server and client during the connector tester load run.

## Changes
- `PropertyWriteContext` carries a lazily-resolved, per-write timestamp cache. Storage, queue, observable, transaction capture, and the derived-property cascade all read the same cached value, so the drift is structurally impossible.
- Cascade dependents inherit the trigger's captured timestamp via parameter threading (no `SubjectChangeContext` scope push): the trigger's raw cache value is passed into the dependent's `PropertyWriteContext` via an internal ctor, pre-populating the cache. A side-effect write inside a derived getter's recalc body still resolves against the outer scope, not the trigger's value.
- Encoding: one `long` carries three states (`0` uninitialized, `> 0` real ticks, `< -1` negated ticks for an explicit-null scope). The negative case lets connectors that require a concrete timestamp still receive one while property storage keeps the never-written sentinel.
- Removed `SubjectChangeContext.ChangedTimestamp` and `ChangedTimestampUtcTicks` from the public/internal surface (audit found zero external readers). Within a write chain, use `PropertyWriteContext.WriteTimestamp`.
- `PropertyWriteContext<T>` grew by 8 bytes (the cache field).
- Snapshot tests in `Namotion.Interceptor.Connectors.Tests` were updated because cascade timestamps now share the trigger's value (previously drifted microseconds apart).

## Perf
Final paired suite + isolated re-runs, `LaunchCount=5`, Linux Ubuntu 24.04 / i7-11700K / .NET 9.0.11. The bench suite was widened in #310 so the write surface is properly covered.

| Benchmark | Master | Branch | Δ % | Note |
|---|---:|---:|---:|---|
| `RegistryBenchmark.IncrementDerivedAverage` | 7,569 ns | 7,149 ns | **-4.8%** | non-overlapping CIs, real cascade improvement |
| `RegistryBenchmark.Write` (real leaf write) | 1,347 ns | 1,353 ns | +0.5% | within noise |
| `RegistryBenchmark.WriteWithTimestampScope` (connector import) | 1,238 ns | 1,257 ns | +1.5% | small structural cost (~5 ns/write) |
| `RegistryBenchmark.WriteNoOp` (was `Write`; equality short-circuit) | 478.5 ns | 488.0 ns | +2.0% | small structural cost (~2 ns/write) |

Per-write structural cost on real writes is **2-5 ns** from the 8-byte cache field on `PropertyWriteContext<T>`. Everything else in the suite (`WriteToRegistrySubjects`, `WriteToSource`, `LinearChain`, `MixedDependencies`, `NoDependencies`, `Read`, `DerivedAverage`, `ChangeAllTires`, `AddLotsOfPreviousCars`, `SubjectUpdate*`, `SourcePathProvider*`, etc.) falls within noise. Zero allocation deltas anywhere.

Full raw tables and isolated re-run data: see [the benchmark comment](https://github.com/RicoSuter/Namotion.Interceptor/pull/303#issuecomment-4468501255).

## Test plan
- [x] Full solution unit tests pass
- [x] OPC UA connector tester (load profile) converged after the fix
- [x] `WriteTimestampTests` covers no-scope, explicit-timestamp, explicit-null, queue, observable, multi-read stability, manual recalc, cascade side-effect independence


src/Namotion.Interceptor.Tracking/Change/DerivedPropertyChangeHandler.cs
```diff
@@ -165,19 +165,38 @@ public void WriteProperty<TProperty>(ref PropertyWriteContext<TProperty> context
...
-            for (var i = 0; i < usedByProperties.Length; i++)
+            var timestampUtcTicks = context.WriteTimestampForStorage;
+            // Share trigger's timestamp with cascade chain so derived recalc events match the
+            // trigger's stored property timestamp. Skip scope push for explicit null (ticks == 0)
+            // where cascade child retains its own behavior.
+            if (timestampUtcTicks > 0)
             {
-                var dependent = usedByProperties[i];
-                if (dependent == context.Property)
+                using (SubjectChangeContext.WithChangedTimestamp(timestampUtcTicks))
```
RicoSuter: should we maybe create scope only if there are actual dependents? might this improve performance (fast path)? ideally we avoid any cost when there are dependents somehow

src/Namotion.Interceptor/Interceptors/IWriteInterceptor.cs
```diff
@@ -54,6 +62,79 @@ public PropertyWriteContext(PropertyReference property, TProperty currentValue,
...
+    {
+        var scopeTicks = SubjectChangeContext.CurrentChangedTimestamp;
+        long result;
+        if (scopeTicks > 0)
+        {
+            result = scopeTicks;
+        }
+        else if (scopeTicks == SubjectChangeContext.NullTimestampTicks)
+        {
+            result = -SubjectChangeContext.GetUtcNowTicks(); // negative = explicit-null encoding
```
RicoSuter: do we need negative utc here? why not just 0?

src/Namotion.Interceptor/SubjectChangeContext.cs
```diff
@@ -20,13 +20,36 @@ public readonly struct SubjectChangeContext
     /// <summary>
     /// Timestamp was explicitly set to null (source had no timestamp).
     /// Distinct from <see cref="UndefinedTimestampTicks"/> which triggers a fallback to <see cref="GetTimestampFunction"/>.
+    /// Exposed as internal so <c>PropertyWriteContext</c> can recognize the null-scope state.
     /// </summary>
-    private const long NullTimestampTicks = -1;
+    internal const long NullTimestampTicks = -1;
```
RicoSuter: does this have to be internal? not private?

src/Namotion.Interceptor.Tracking/Change/DerivedPropertyChangeHandler.cs
RicoSuter: Already gated — the `using (SubjectChangeContext.WithChangedTimestamp(...))` is inside `if (usedByProperties.Length > 0)` (line 159), so the scope only fires when there actually are dependents. Properties with no dependents skip the scope entirely.

(For also eliminating the scope when dependents *do* exist, the only path I could find was a "pre-seeded cascade child" constructor + parallel internal `IInterceptorExecutor.SetPropertyValue` overload — I prototyped that and we evaluated it, but the ~80 ns cascade-only win wasn't worth the parallel write API surface and the extra internal coupling. Reverted in favor of the current shape.)

src/Namotion.Interceptor/Interceptors/IWriteInterceptor.cs
RicoSuter: Yes, the negative encoding is load-bearing. One `long` carries three states for the cache:

| State | Cache value | `WriteTimestampForStorage` | `WriteTimestampForPublishing` |
|---|---|---|---|
| Uninitialized | `0` (re-resolve sentinel) | n/a | n/a |
| Real timestamp | positive ticks | positive | positive |
| Explicit-null scope | **negative** ticks | `0` (never-written sentinel) | `\|negative\|` (positive ticks for consumers) |

If we stored `0` for the explicit-null case we couldn't distinguish "cache not yet populated" from "cache resolved and value was null" — every read would re-resolve via UtcNow, reintroducing the timestamp-drift bug for null-scope writes. Alternative encoding would be a separate `bool _isNullEncoded` field (8 more bytes for alignment). The sign-bit trick keeps it to one `long`.

I expanded the doc comment on `_writeTimestamp` to spell this out at the field declaration so future readers don't have to reverse-engineer it. See `d72f2d6`.

src/Namotion.Interceptor/SubjectChangeContext.cs
RicoSuter: Needs to be `internal` because `PropertyWriteContext.ResolveAndCacheWriteTimestamp` (in `IWriteInterceptor.cs`, same assembly) reads `CurrentChangedTimestamp` and compares it against `NullTimestampTicks` to decide whether to encode the cache as negative (explicit-null scope) or positive (real ticks). Two structs in the same assembly need to agree on the sentinel value.

Alternatives I considered:
- Move the sentinel check + encode logic *into* `SubjectChangeContext` (e.g., a method `ResolveCurrentChangedTimestampForCache()` that returns the negative-encoded long). That pushes `PropertyWriteContext`'s cache-encoding contract back into `SubjectChangeContext` — coupling in the wrong direction.
- Return a `(long ticks, bool isNullScope)` tuple. Hides the sentinel but adds tuple overhead on the lazy-resolve hot path.

Keeping `internal` is the cleanest of the three: one const leak vs reverse coupling or tuple cost.

RicoSuter:
## Updated benchmark results vs master (with realistic write coverage)

After [merging #310](https://github.com/RicoSuter/Namotion.Interceptor/pull/310) the suite now covers three distinct write paths. The old `Write` (no-op fast path) has been renamed `WriteNoOp` and two new ones added: `Write` (real leaf write with varying value) and `WriteWithTimestampScope` (connector import path).

Paired suite via `scripts/benchmark.ps1 -LaunchCount 5 -Stash`. Verified deltas (full suite + isolated re-runs):

| Benchmark | Master | Branch | Δ ns | Δ % | Per-write |
|---|---:|---:|---:|---:|---:|
| `RegistryBenchmark.IncrementDerivedAverage` | 7,569 ns | **7,149 ns** | **-360** | **-4.8%** | n/a (cascade win) |
| `RegistryBenchmark.WriteNoOp` (was `Write`) | 478.5 ns | 488.0 ns | +9.5 | +2.0% | +2.4 ns/write |
| `RegistryBenchmark.Write` (real leaf write) | 1,346.7 ns | 1,353.1 ns | +6.4 | +0.5% | +1.6 ns/write (within noise) |
| `RegistryBenchmark.WriteWithTimestampScope` (connector import) | 1,238.2 ns | 1,257.1 ns | +18.9 | +1.5% | +4.7 ns/write |

### What changed in the picture

Previous comments showed `Write` (the historical bench) as the only measured number, and the latest reading there was +22 ns / +4.6%. That bench is now correctly named `WriteNoOp` — every measured invocation is short-circuited by the equality-check interceptor, so it measures interceptor-chain overhead with no terminal store. It's a legitimate fast path (connectors that republish unchanged values), but it's not a typical write.

The new `Write` and `WriteWithTimestampScope` measure the real terminal-write path. The structural cost there is **2-5 ns per write** — meaningfully smaller than what `WriteNoOp` suggested. The cascade improvement (-4.8% on `IncrementDerivedAverage`) more than compensates in any code that exercises derived properties.

### Cascade-heavy code is faster

`IncrementDerivedAverage`: 7,568.8 → 7,208.7 ns (-360 ns / -4.8%), CIs non-overlapping. This is the only bench with non-trivial cascade work and it's the largest win.

### Everything else within noise

`WriteToRegistrySubjects` 2,908 vs 2,958 us, `WriteToSource` 1,975 vs 2,000 us, `LinearChain` 1,347 vs 1,333 ns, `MixedDependencies` 1,358 vs 1,354 ns, `NoDependencies` 1,226 vs 1,232 ns, `VisitPropertiesFromPaths` 730 vs 744 ns, `Read` 626 vs 610 ns, `DerivedAverage` 395 vs 392 ns, `ChangeAllTires` 17,289 vs 17,226 ns, `AddLotsOfPreviousCars` 67.69 vs 67.38 ms, `CreateCompleteUpdate` 7,821 vs 7,789 ns, `CreatePartialUpdate` 2,638 vs 2,646 ns, `WithFirstLastGroups` 1,666 vs 1,662 ns, `TryGetPropertyFromPath` 357.78 vs 356.63 ns, `TryGetPath` 286.45 vs 287.32 ns, `TryGetPropertyFromSegment` 48.83 vs 48.24 ns, `GetOrAddSubjectId` 27.27 vs 27.59 ns, `GenerateSubjectId` 1,028 vs 1,028 ns. Zero allocation deltas anywhere.

### Net assessment

- Cascade-heavy code: clear -4.8% win (`IncrementDerivedAverage`, non-overlapping CIs).
- Real writes: +1-3% structural cost from the 8-byte cache field on `PropertyWriteContext<T>` (one extra long store per write).
- Connector import path: +1.5% (single-digit ns per write).
- Equality-short-circuit fast path: +2% (single-digit ns per write).
- Everything else: noise.

Tradeoff is good: 2-5 ns per real write in exchange for the timestamp-drift bug being structurally impossible, plus a real cascade improvement.

<details>
<summary>Raw output (current branch, full suite)</summary>

```
.NET SDK 10.0.100
  [Host]     : .NET 9.0.11 (9.0.11, 9.0.1125.51716), X64 RyuJIT x86-64-v4
  Job-PTHBVM : .NET 9.0.11 (9.0.11, 9.0.1125.51716), X64 RyuJIT x86-64-v4

LaunchCount=5
```

| Type                          | Method                    | Mean             | Error         | StdDev         | Median           | Allocated  |
|-------------------------------|---------------------------|-----------------:|--------------:|---------------:|-----------------:|-----------:|
| ServiceOrderResolverBenchmark | NoDependencies            |      1,232.01 ns |      7.612 ns |      19.098 ns |      1,230.64 ns |     1632 B |
| SourcePathProviderBenchmark   | TryGetPropertyFromPath    |        356.63 ns |      0.834 ns |       2.046 ns |        356.96 ns |      632 B |
| SubjectSourceBenchmark        | WriteToRegistrySubjects   |  2,957,683.57 ns | 44,375.526 ns | 107,171.890 ns |  2,909,680.96 ns |          - |
| SubjectUpdateBenchmark        | CreateCompleteUpdate      |      7,789.42 ns |     22.721 ns |      56.582 ns |      7,790.13 ns |     5904 B |
| ServiceOrderResolverBenchmark | LinearChain               |      1,332.67 ns |      7.224 ns |      17.584 ns |      1,332.74 ns |     2640 B |
| SourcePathProviderBenchmark   | VisitPropertiesFromPaths  |        743.54 ns |      9.302 ns |      22.641 ns |        755.72 ns |     1280 B |
| SubjectSourceBenchmark        | WriteToSource             |  1,999,784.12 ns | 21,088.232 ns |  52,124.890 ns |  1,993,642.97 ns |          - |
| SubjectUpdateBenchmark        | CreatePartialUpdate       |      2,646.10 ns |     10.347 ns |      25.380 ns |      2,631.14 ns |     2656 B |
| ServiceOrderResolverBenchmark | MixedDependencies         |      1,353.69 ns |      7.801 ns |      19.428 ns |      1,356.11 ns |     2064 B |
| SourcePathProviderBenchmark   | TryGetPropertyFromSegment |         48.24 ns |      1.727 ns |       4.269 ns |         47.18 ns |          - |
| ServiceOrderResolverBenchmark | WithFirstLastGroups       |      1,662.34 ns |      9.782 ns |      23.995 ns |      1,661.71 ns |     2576 B |
| SourcePathProviderBenchmark   | TryGetPath                |        287.32 ns |      2.411 ns |       6.003 ns |        286.26 ns |      264 B |
| RegistryBenchmark             | AddLotsOfPreviousCars     | 67,382,035.39 ns | 98,392.648 ns | 239,501.321 ns | 67,367,977.88 ns | 23856664 B |
| RegistryBenchmark             | IncrementDerivedAverage   |      7,208.68 ns |     23.046 ns |      53.869 ns |      7,224.31 ns |      128 B |
| RegistryBenchmark             | WriteNoOp                 |        493.86 ns |      7.302 ns |      17.353 ns |        502.98 ns |          - |
| RegistryBenchmark             | Write                     |      1,366.88 ns |     17.696 ns |      43.409 ns |      1,345.58 ns |          - |
| RegistryBenchmark             | WriteWithTimestampScope   |      1,263.71 ns |     13.954 ns |      34.492 ns |      1,266.52 ns |          - |
| RegistryBenchmark             | Read                      |        609.83 ns |      4.234 ns |      10.386 ns |        604.99 ns |          - |
| RegistryBenchmark             | DerivedAverage            |        392.09 ns |      2.789 ns |       6.894 ns |        395.39 ns |          - |
| RegistryBenchmark             | ChangeAllTires            |     17,225.66 ns |     57.917 ns |     143.156 ns |     17,224.58 ns |    17216 B |
| RegistryBenchmark             | GetOrAddSubjectId         |         27.59 ns |      0.228 ns |       0.550 ns |         27.42 ns |          - |
| RegistryBenchmark             | GenerateSubjectId         |      1,027.50 ns |      0.316 ns |       0.751 ns |      1,027.28 ns |          - |

</details>

<details>
<summary>Raw output (master, full suite)</summary>

```
.NET SDK 10.0.100
  [Host]     : .NET 9.0.11 (9.0.11, 9.0.1125.51716), X64 RyuJIT x86-64-v4
  Job-PTHBVM : .NET 9.0.11 (9.0.11, 9.0.1125.51716), X64 RyuJIT x86-64-v4

LaunchCount=5
```

| Type                          | Method                    | Mean             | Error          | StdDev         | Median           | Allocated  |
|-------------------------------|---------------------------|-----------------:|---------------:|---------------:|-----------------:|-----------:|
| ServiceOrderResolverBenchmark | NoDependencies            |      1,226.18 ns |       1.922 ns |       4.715 ns |      1,226.15 ns |     1632 B |
| SourcePathProviderBenchmark   | TryGetPropertyFromPath    |        357.78 ns |       1.845 ns |       4.491 ns |        358.19 ns |      632 B |
| SubjectSourceBenchmark        | WriteToRegistrySubjects   |  2,908,569.62 ns |     727.621 ns |   1,812.030 ns |  2,909,252.21 ns |          - |
| SubjectUpdateBenchmark        | CreateCompleteUpdate      |      7,821.41 ns |      25.660 ns |      62.944 ns |      7,817.05 ns |     5904 B |
| ServiceOrderResolverBenchmark | LinearChain               |      1,346.55 ns |       7.609 ns |      18.666 ns |      1,355.14 ns |     2640 B |
| SourcePathProviderBenchmark   | VisitPropertiesFromPaths  |        730.25 ns |      11.896 ns |      28.957 ns |        713.13 ns |     1280 B |
| SubjectSourceBenchmark        | WriteToSource             |  1,974,779.73 ns |  13,244.983 ns |  33,229.075 ns |  1,978,447.09 ns |          - |
| SubjectUpdateBenchmark        | CreatePartialUpdate       |      2,637.72 ns |       6.879 ns |      17.131 ns |      2,645.32 ns |     2656 B |
| ServiceOrderResolverBenchmark | MixedDependencies         |      1,358.15 ns |       5.633 ns |      14.133 ns |      1,358.77 ns |     2064 B |
| SourcePathProviderBenchmark   | TryGetPropertyFromSegment |         48.83 ns |       1.086 ns |       2.664 ns |         47.52 ns |          - |
| ServiceOrderResolverBenchmark | WithFirstLastGroups       |      1,665.90 ns |      11.884 ns |      29.814 ns |      1,682.60 ns |     2576 B |
| SourcePathProviderBenchmark   | TryGetPath                |        286.45 ns |       2.202 ns |       5.443 ns |        285.56 ns |      264 B |
| RegistryBenchmark             | AddLotsOfPreviousCars     | 67,691,325.77 ns | 148,862.795 ns | 362,352.643 ns | 67,662,199.38 ns | 23856654 B |
| RegistryBenchmark             | IncrementDerivedAverage   |      7,568.84 ns |      80.224 ns |     192.212 ns |      7,645.13 ns |      128 B |
| RegistryBenchmark             | WriteNoOp                 |        483.96 ns |       6.436 ns |      15.543 ns |        482.51 ns |          - |
| RegistryBenchmark             | Write                     |      1,333.72 ns |       9.330 ns |      22.710 ns |      1,316.93 ns |          - |
| RegistryBenchmark             | WriteWithTimestampScope   |      1,221.54 ns |      12.190 ns |      29.206 ns |      1,213.01 ns |          - |
| RegistryBenchmark             | Read                      |        626.22 ns |      17.568 ns |      43.751 ns |        601.46 ns |          - |
| RegistryBenchmark             | DerivedAverage            |        395.18 ns |       3.029 ns |       7.315 ns |        395.63 ns |          - |
| RegistryBenchmark             | ChangeAllTires            |     17,288.68 ns |      43.154 ns |     107.468 ns |     17,265.76 ns |    17216 B |
| RegistryBenchmark             | GetOrAddSubjectId         |         27.27 ns |       0.040 ns |       0.099 ns |         27.24 ns |          - |
| RegistryBenchmark             | GenerateSubjectId         |      1,027.91 ns |       0.617 ns |       1.455 ns |      1,027.39 ns |          - |

</details>

<details>
<summary>Isolated write re-runs (LC=5, focused filter)</summary>

Current branch:
| Method | Mean | Error | StdDev |
|---|---:|---:|---:|
| WriteNoOp | 488.0 ns | 3.52 ns | 8.51 ns |
| Write | 1,353.1 ns | 11.13 ns | 27.10 ns |
| WriteWithTimestampScope | 1,257.1 ns | 9.38 ns | 22.82 ns |

Master:
| Method | Mean | Error | StdDev |
|---|---:|---:|---:|
| WriteNoOp | 478.5 ns | 5.95 ns | 14.72 ns |
| Write | 1,346.7 ns | 17.85 ns | 41.36 ns |
| WriteWithTimestampScope | 1,238.2 ns | 10.61 ns | 25.63 ns |

</details>

PR #304, opened 2026-05-12
refactor: Restructure ConnectorTester into focused components and fix 7 bugs

RicoSuter:
## Summary

Strangler-fig refactor of `Namotion.Interceptor.ConnectorTester` into focused, individually-testable components, plus 7 bug fixes from the design spec and one DI cycle fix uncovered during smoke tests. 35 commits, 80+ files touched. Public CSV output is byte-identical to master, pinned by golden-string snapshot tests.

## Bug fixes

- **Bug 1** Parallel value-mutation batches now observe the same timestamp. `BatchValueMutationStrategy.MutateBatchParallel` captures `batchTimestamp = DateTimeOffset.UtcNow` once on the orchestrator thread and re-enters `SubjectChangeContext.WithChangedTimestamp(batchTimestamp)` inside each `Parallel.For` worker (the previous outer scope was invisible to worker threads because `_current` is `[ThreadStatic]`).
- **Bug 2** `--participant` arg validated against the configured server + clients via `RunModeSelector`. Typos now exit with `Unknown participant '<name>'. Available participants: server, client.` instead of starting a participant with no model.
- **Bug 3** `ChaosConfiguration.Validate()` rejects unknown `Mode` values; only `kill`, `disconnect`, `both` (case-insensitive) are accepted.
- **Bug 4** `Program.cs` now wraps `host.RunAsync()` in `try/finally` so `PerformanceProfiler` instances are disposed even on exception / abnormal shutdown.
- **Bug 5** Convergence loop uses a `Stopwatch`-based timeout instead of `maxPolls = timeout / pollInterval`. A `ConvergenceTimeout` shorter than `SnapshotPollInterval` no longer causes the loop body to be skipped; at least one snapshot is always captured.
- **Bug 7** `ChaosEngine.RecoverActiveDisruptionAsync` (which only ever returned `Task.CompletedTask`) is now `RecoverActiveDisruption` returning `void`. Call site updated.
- **Bug 8** All three CSV outputs (`cycles.csv`, `chaos-events.csv`, `performance-*.csv`) flow through a single column-based `CsvFile<TRow>` with `CultureInfo.InvariantCulture` and a cached `StreamWriter`. Header and row formats can no longer drift; each schema has golden-string regression tests.
- **DI cycle** `IFaultTargetResolver` no longer resolves `IEnumerable<ISubjectConnector>` inside its DI factory (ChaosEngine is an `IHostedService` that depends on the resolver, and the resolver's input lives on the `IHostedService` graph; resolving connectors from inside the factory re-entered the ChaosEngine factory). Resolver is now constructed empty pre-Build and `Bind(participants, connectors)` is called after `host.Build()` enumerates the singletons. Also switched profile detection from `Environment.GetEnvironmentVariable("DOTNET_ENVIRONMENT")` to `builder.Environment.EnvironmentName` so `--environment` CLI args and `ASPNETCORE_ENVIRONMENT` are honored.

## Folder reorg

- `Engine/Mutation/` - `TickPlan`, `GlobalMutationCounter`, `MutationCounters`, `KnownNodeGraph`, `StructuralMutator`, `IValueMutationStrategy` + `Random`/`Batch` strategies, `MutationEngineHost`. Eliminates the old `MutationEngine`/`RandomMutationEngine`/`BatchMutationEngine` class hierarchy.
- `Engine/Chaos/` - `ChaosEngine`, `ChaosEventRecord`, `ChaosProfileRotator`.
- `Engine/Verification/` - `CycleResult`, `ICycleLifecycleNotifier`, `HeapSampler`, `FindingsLog`, `ConvergenceChecker`, `FailureDiagnostics`, `CycleStatistics`, slim `VerificationEngine` (~190 lines, was 500+).
- `Snapshot/` - `SnapshotComparer`, `SnapshotIdMap`, `SnapshotDiffer` (structured diff entries).
- `Reporting/` - `CsvFile<TRow>` + `CsvColumn<TRow>`, plus per-schema `CyclesCsv`, `ChaosEventsCsv`, `PerformanceCsv`.
- `Performance/` - `PerformanceProfiler`, `PerformanceWindow`, `PerformanceConsoleReporter`, `ReservoirSampler`.
- `Connectors/` - `ConnectorKind` enum, `IConnectorBindings` strategy with OPC UA / MQTT / WebSocket implementations, `ConnectorBindingsRegistry`, `IFaultTargetResolver` (built empty, bound post-`host.Build()`).
- `Hosting/` - `RunMode` + `RunModeSelection`, `RunModeSelector`, `ConnectorTesterHost.Build(args)` (owns `runDirectory` creation from `builder.Environment.EnvironmentName`). `Program.cs` is now ~20 lines.

## Smoke test results

| Profile | Cycles | Result | Notes |
| --- | --- | --- | --- |
| opcua-chaos | 5/5 PASS | all profiles | 28 chaos events recovered, 0.1s converge p99 |
| mqtt-chaos | 5/5 PASS | all profiles | 26 chaos events recovered, 0.0s converge |
| websocket-chaos | in progress | - | cycles 1-4 PASS; cycle 5 (full-chaos) underway |
| opcua-load | 5 min | PASS | steady at 20,000 published/s on server and client after 60 s warmup; P99 latency ~150-400 ms |
| mqtt-load | - | not yet run | - |
| websocket-load | - | not yet run | - |
| multi-process | - | not yet run | - |
| `--participant typo` | - | not yet run | - |

All 117 unit tests pass; build is clean with `TreatWarningsAsErrors=true` (was 26 tests on master, +91 new).

## Follow-ups (log-quality cleanup, not blockers)

- [ ] Bare exception lines without `[time] [level] [logger]` prefix appear after `WARN` messages in chaos cycles. Make the logger emit stack traces under the same prefix so log entries are line-grep-able.
- [ ] `MqttSubjectClientSource` (and the equivalent WebSocket source) logs under the same category for both `client-a` and `client-b`; add the participant name to the logger category or message prefix so client-a vs client-b is distinguishable in the per-cycle log.
- [ ] Collapse repetitive retry warnings (~4 lines per disruption `Failed to write N`, `Failed to flush`, `Successfully flushed N`) into a single per-disruption summary line: `Buffered N writes during X.Xs disconnection; flushed M on reconnect`.
- [ ] Echo the active chaos profile in the end-of-cycle statistics block, not only in the start marker.
- [ ] On PASS, log a brief snapshot-equality summary (`verified N nodes, M properties equal`) so a reader can see what was actually compared without trusting the absence of a diff dump.
- [ ] Optional: tag every log line within a cycle with a `[Cn]` prefix so console output during overlapping cycles stays readable.

## How to run

```
dotnet run --project src/Namotion.Interceptor.ConnectorTester --launch-profile opcua-chaos --configuration Release
dotnet run --project src/Namotion.Interceptor.ConnectorTester --launch-profile mqtt-chaos --configuration Release
dotnet run --project src/Namotion.Interceptor.ConnectorTester --launch-profile websocket-chaos --configuration Release
dotnet run --project src/Namotion.Interceptor.ConnectorTester --launch-profile opcua-load --configuration Release
dotnet run --project src/Namotion.Interceptor.ConnectorTester --launch-profile mqtt-load --configuration Release
dotnet run --project src/Namotion.Interceptor.ConnectorTester --launch-profile websocket-load --configuration Release
```

Multi-process: `-- --participant server` in one terminal, `-- --participant client` in another. Typo path: `-- --participant typo-name` exits immediately with `Unknown participant 'typo-name'. Available participants: server, client.` on stderr.


src/Namotion.Interceptor.ConnectorTester/Configuration/ConnectorTesterConfiguration.cs
```diff
@@ -1,20 +1,32 @@
...
 
 public class ConnectorTesterConfiguration
 {
     /// <summary>"opcua", "mqtt", or "websocket"</summary>
     public string Connector { get; set; } = "opcua";
 
+    /// <summary>Parsed connector kind based on <see cref="Connector"/>.</summary>
+    public ConnectorKind ConnectorKind => Connector?.ToLowerInvariant() switch
+    {
+        "opcua"     => Namotion.Interceptor.ConnectorTester.Connectors.ConnectorKind.OpcUa,
```
RicoSuter: full type name needed here?

src/Namotion.Interceptor.ConnectorTester/Connectors/Mqtt/MqttTickTimestampCodec.cs
```diff
@@ -0,0 +1,25 @@
+using System.Buffers.Text;
+
+namespace Namotion.Interceptor.ConnectorTester.Connectors.Mqtt;
+
+/// <summary>
+/// Tick-precision UTC timestamp serialization for MQTT source timestamps. Defaults to
+/// Unix milliseconds in production; the connector tester uses ticks so snapshot comparison
+/// observes the same timestamp on both sides without millisecond rounding loss.
+/// </summary>
+public static class MqttTickTimestampCodec
```
RicoSuter:
```suggestion
public static class MqttTickTimestampSerialization
```

src/Namotion.Interceptor.ConnectorTester/Configuration/ConnectorTesterConfiguration.cs
RicoSuter: can we use some sort of string to enum reflection here so that it just works for new enums without touching this method?

RicoSuter: Resolved in 191c55bc: the hardcoded switch (with the redundant fully-qualified names) was replaced by `Enum.TryParse<ConnectorKind>(Connector, ignoreCase: true, ...)`, so the full type-name prefix is gone.

src/Namotion.Interceptor.ConnectorTester/Connectors/Mqtt/MqttTickTimestampCodec.cs
RicoSuter: Done in 191c55bc: renamed the class and the file to `MqttTickTimestampSerialization`; both call sites in `MqttConnectorFactory` updated.

src/Namotion.Interceptor.ConnectorTester/Configuration/ConnectorTesterConfiguration.cs
RicoSuter: Done in 191c55bc: replaced the switch with `Enum.TryParse<ConnectorKind>(Connector, ignoreCase: true, out var kind)`, so new connector kinds parse automatically without touching this method. Kept an explicit throw for unknown values, now listing `Enum.GetNames<ConnectorKind>()`.

PR #313, opened 2026-05-18
performance: Batch OPC UA browse and read operations to reduce round-trips

RicoSuter:
## Summary

Reduces OPC UA load round-trips by batching browse and read operations across sibling nodes, adds resilience for servers that reject large batches or return per-NodeId bad statuses, and centralizes status-code classification across browse/read/write/subscribe paths.

### `Namotion.Interceptor.OpcUa`

**Loader pipeline (`OpcUaSubjectLoader`, `OpcUaLoadContext`, `OpcUaAttributeLoader`)**
- Restructure `OpcUaSubjectLoader` into a five-phase pipeline (`LoadSubjectsAsync`) that processes sibling nodes in bulk instead of recursing one-by-one
- Introduce `OpcUaLoadContext` (new) with a per-load browse cache to avoid re-browsing Object nodes already seen during type classification, plus a transactional apply/rollback so a failed load leaves no partially-attached subjects
- Extract attribute discovery into `OpcUaAttributeLoader` (new); cap traversal depth via the new `MaxAttributeTraversals` configuration (default 100) with per-pair `(property, parentNodeId)` dedup across rounds, and check cancellation between rounds (cache-only rounds do no awaited I/O)
- Fix `ResolveObjectTypesAsync` to `TryGetValue` on the browse result so a bad-status browse for a dynamic Object child no longer throws `KeyNotFoundException`; the missing entry is left unset and `TryCreateDynamicProperty` logs and skips the property, with the next load retrying via reconnect
- Reuse existing collection/dictionary children across loads: collections rebind by index position, dictionaries by key. Dictionary children are matched by the invariant string form of their key, so string- and int-keyed subject dictionaries rebind existing instances instead of recreating them on every load
- Deterministic outcome for graph-shaped address spaces: when the same `PropertyReference` is reachable through multiple paths with different NodeIds, `OpcUaLoadContext.QueueClaim` and `BatchLoadVariableNodesAsync` both keep the smaller NodeId so the load result is reproducible across browse orderings

**Session extensions (`OpcUaSessionExtensions`, new)**
- `BrowseNodesAsync`, `ReadNodesAsync`, and `DistinctByResolvedNodeId` extension methods
- `DistinctByResolvedNodeId` filters references with null or empty `BrowseName` at the session boundary so downstream loaders can dereference `Reference.BrowseName.Name` without per-callsite null checks
- Batch browse and read respecting `OperationLimits.MaxNodesPerBrowse` / `MaxNodesPerRead`, with a 256 soft cap when the server reports 0 (unbounded)
- Split-and-retry recovery for `BadTooManyOperations` / `BadEncodingLimitsExceeded` / `BadResponseTooLarge`
- Continuation-point paging across multi-node browses with a `MaxBrowseContinuations` safety bound (default 100); best-effort release on failure on a fresh 5-second-timeout `CancellationTokenSource` so release succeeds even when the caller's token has been cancelled, without blocking shutdown
- Non-conforming short responses do not silently drop nodes: `ReadNodesAsync` pads short responses to preserve positional alignment (the padded status is returned for the caller to classify), and `BrowseNodesAsync` surfaces any requested node missing from the response as transient, so the structure load retries instead of loading a subject with zero children
- Browse throws `OpcUaTransientServiceException` (new public type) on a per-NodeId transient bad status so an incomplete structure is never loaded; permanent statuses keep the prior log-and-skip behavior. `ReadNodesAsync` is a best-effort primitive that never classifies or throws, so each caller applies its own policy (type resolution aborts on a transient attribute read; the initial value load skips the property)

**Type resolver (`OpcUaTypeResolver`)**
- Replace per-node `TryGetTypeForNodeAsync` with batch `ResolveVariableTypesAsync` (one read for all dynamic variables' DataType + ValueRank) and a synchronous `ResolveObjectNodeType` that operates on already-browsed children
- Removed per-resolver type cache; per-load caching now lives in `OpcUaLoadContext`

**Status-code classifier (`OpcUaStatusCodeClassifier`, new)**
- Shared classifier with a unified permanent-codes whitelist used by browse, read, write, and subscription paths
- Permanent (logged + skipped, never retried): `BadNodeIdUnknown`, `BadNodeIdInvalid`, `BadAttributeIdInvalid`, `BadIndexRangeInvalid`, `BadTypeMismatch`, `BadUserAccessDenied`, `BadSecurityModeInsufficient`, `BadNotImplemented`, `BadNotReadable`, `BadNotWritable`, `BadWriteNotSupported`
- All other bad statuses treated as transient
- `OutboundWriter.IsTransientWriteError` and `SubscriptionHealthMonitor.IsRetryable` migrated to the shared classifier, and `SubscriptionManager` now reuses `SubscriptionHealthMonitor.IsUnhealthy` instead of an inline copy. Behavior changes vs the prior per-callsite whitelists:
  - Writes: newly classified as permanent (no more retry loops): `BadNodeIdInvalid`, `BadIndexRangeInvalid`, `BadSecurityModeInsufficient`, `BadNotImplemented`, `BadNotReadable`
  - Subscription health: newly classified as permanent (no more retry of these monitored items): `BadNodeIdInvalid`, `BadTypeMismatch`, `BadUserAccessDenied`, `BadSecurityModeInsufficient`, `BadNotImplemented`, `BadNotReadable`, `BadNotWritable`, `BadWriteNotSupported`

**Subject client source (`OpcUaSubjectClientSource`)**
- Use `ReadNodesAsync` for initial-state load so all properties are read in one chunked call instead of per-batch loops. The value load is best-effort: a not-ready or transient value (e.g. `BadWaitingForInitialData`) leaves that one property unset for the subscription to backfill, instead of aborting the whole load and forcing a reconnect
- Harden configured-RootPath resolution against references with a null `BrowseName` (root-path browse results bypass `DistinctByResolvedNodeId`), so a non-conforming server no longer throws during connect/reconnect
- Log elapsed time for subject loading + subscription setup

**Configuration (`OpcUaClientConfiguration`)**
- Add `MaxBrowseContinuations` and `MaxAttributeTraversals` safety bounds (both default 100), validated as positive in `Validate()` alongside the other numeric resilience settings
- Behavior change: `RetryTime` default corrected from 1s to 10s (matches the doc comment). Operators upgrading will see a longer first-retry delay after a transient error.

### `HomeBlaze.OpcUa`

- `OpcUaClient` now disposes the client source on stop. Detaching only stops the hosted service, so without this the source's lifecycle subscription and structure-lock semaphore leaked one per start/stop cycle (a fresh source is created on each start). The OPC UA session/subscriptions were already closed by the stop path; this releases the remaining managed state.

### `Namotion.Interceptor.OpcUa.Tests`

**New test files**
- `OpcUaSessionExtensionsTests`: transient throw vs permanent skip for browse and browse-next; read passes bad statuses through (best-effort, never throws), including a mixed good + not-ready `BadWaitingForInitialData` batch; short browse response surfaces as transient
- `OpcUaStatusCodeClassifierTests`: Theory tests covering the permanent codes, transient codes, Good, and Uncertain
- `SubscriptionHealthMonitorTests`: `IsUnhealthy` (not-created and bad-status branches) and `IsRetryable` (permanent vs transient codes via the shared classifier)
- `OpcUaSubjectLoaderFailureTests`: discovery-failure rollback, nested staged-subject rollback, no orphan claims, retry-after-failure, permanent-bad-status skip
- `OpcUaRootPathResolutionTests`: root-path segment matching skips references with a null `BrowseName`
- `OpcUaSubjectLoaderDictionaryReuseTests`: an int-keyed subject dictionary reuses its existing children across a load instead of recreating them

**Subject loader tests split by theme**
- `OpcUaSubjectLoaderTestsBase` (new): shared helpers (`CreateLoader`, mock-session/browse/read setups, reference-description factories) on a public base class with `private protected` members so internal types like `OpcUaSubjectLoader` flow through signatures without leaking outside the test assembly
- `OpcUaSubjectLoaderTests`: basic load, dynamic property type resolution, edge cases (object-browse failure, address-space cycle, null BrowseName)
- `OpcUaSubjectLoaderDedupTests` (new): multi-path subject reuse, same-NodeId-via-different-reference-types dedup, sibling subjects sharing a Variable NodeId, shared dynamic variable read dedup
- `OpcUaSubjectLoaderAttributeTests` (new): `MaxAttributeTraversals` cap, pre-registered attribute safety net, cross-round browse-cache fallback
- `OpcUaSubjectLoaderBatchingTests` (new): `MaxNodesPerBrowse` chunking, mixed-tree end-to-end, browse and BrowseNext split-and-retry, continuation paging, runaway-server cap, dynamic collections and dictionaries
- All tests use realistic `ReadAsync`/`BrowseAsync` mocks instead of mocking `OpcUaTypeResolver` directly

**Other updated test files**
- `OpcUaTypeResolverTests`: covers the new `ResolveObjectNodeType` / `ResolveVariableTypesAsync` surface; asserts the resolver aborts (throws) on a transient DataType/ValueRank read and skips (null type) on a permanent one; alignment-padding test confirms positional alignment
- `OpcUaClientConfigurationTests`: rejects non-positive `MaxBrowseContinuations` / `MaxAttributeTraversals`
- `VerifyChecksTests.PublicApi.verified.txt`: accepts the new `OpcUaTransientServiceException` public type and the `OpcUaTypeResolver` surface changes

## Test plan

- [x] All OPC UA tests pass (290 unit + 27 integration), including the new classifier, session-extension, subscription-health-monitor, failure, root-path, and dictionary-reuse tests
- [x] CI green on the latest commit (`test`, `test-opcua-integration`, `test-homeblaze-integration`)
- [x] OPC UA chaos connector tester runs 50+ cycles across all five rotating fault profiles (no-chaos / server-only / client-a-only / all-clients / full-chaos) with zero failures, zero findings, stable heap
- [x] Public API snapshot updated for the `OpcUaTypeResolver` surface change (`TryGetTypeForNodeAsync` removed; `ResolveObjectNodeType` and `ResolveVariableTypesAsync` added) and the new `OpcUaTransientServiceException`

## Follow-ups

- #320: defer root-subject value mutations for atomic observer view (Blazor flicker reduction; UX improvement, not correctness)
- #332: drop permanently-failed writes from the retry queue (the transient/permanent split is currently informational on the write path; pre-existing, not introduced here)

## Breaking changes

- **Configuration property renames**: `MaximumItemsPerSubscription`, `SubscriptionMaximumNotificationsPerPublish`, and `MaximumReferencesPerNode` were renamed to `MaxItemsPerSubscription`, `SubscriptionMaxNotificationsPerPublish`, and `MaxReferencesPerNode` for .NET naming consistency. Configuration binding (appsettings.json) is name-based and does not fail on unknown keys, so external consumers using the old names will silently fall back to defaults. Update your configuration keys accordingly.
- **`RetryTime` default changed from 1s to 10s**: the previous value contradicted its own XML documentation (which always said 10s) and the MQTT/WebSocket connectors (both 10s). Reconnects after a disconnect now happen 10x slower by default; set `RetryTime` explicitly if you tuned around the old 1s cadence.


src/Namotion.Interceptor.OpcUa.Tests/Client/SubscriptionHealthMonitorTests.cs
```diff
@@ -0,0 +1,123 @@
...
+
+        // Act
+        var isRetryable = SubscriptionHealthMonitor.IsRetryable(item);
+
+        // Assert
+        Assert.True(isRetryable);
+    }
+
+    [Fact]
+    public void WhenItemHasGoodStatus_ThenIsRetryableReturnsFalse()
```
RicoSuter: is this correct?

RicoSuter:
## Design note: why this PR is ~2,000 production LOC (lean-alternative evaluation)

The original goal was batch loading only, so before merging we evaluated whether a leaner design would have sufficed. Method: reconstruct the pre-PR loader from master, extract the behavioral requirements pinned by the current test suite (37 design-independent invariants), design a minimal alternative (old per-node walk plus level-by-level batch prefetch, with the simplest failure story meeting the reliability bar), then adjudicate it against all 37 requirements and attack it adversarially. Reference bar: retries must always converge to a correct synced model, no torn state, no orphaned subjects, no stale ownership claims, thread-safe.

### Findings

**A "just batch the browses" loader (~900 LOC) fails the bar outright.** It fails roughly 15 of 37 requirements, all in failure semantics: torn state on mid-load abort, leaked claims without monitored items, dynamic properties permanently lost on retry.

**The crude "full reset on failure, retry from zero" strategy is not viable and not even cheaper.** Three blockers: (1) releasing all claims on failure breaks ownership retention across reloads, leaving application writes unrouted (pinned by test); (2) in-place writes to pre-existing subjects cannot be undone by any reset; (3) dynamic properties cannot be removed (no RemoveProperty API). Cost if pursued anyway: ~90 LOC plus the ~200 LOC of per-attempt tracking it cannot avoid.

**The viable lean design converges onto this PR's architecture.** It reuses the wire layer unchanged (SessionExtensions, StatusCodeClassifier, TypeResolver discharge 14 of 37 requirements on their own) and re-derives the staged load context mechanics point for point: queued claims deduped smallest-NodeId-wins, deferred root ops, deepest-first rollback, this-attempt-only claim release. Test matrix: 36 of 37 pass, 1 needs test adaptation, 0 fail. Estimated saving: 200 to 250 lines, essentially comments and naming polish.

**Adversarial review erased even that saving.** The lean design assumed the source orchestration could stay as on master; four attack classes falsify this: (1) its own staged rollback deadlocks against master's lock-taking RemoveItemsForSubject on every failed load, hanging retry and dispose; (2) dispose during an in-flight load stalls without the load cancellation hook; (3) malformed root-path references crash-loop deterministically; (4) root-path browses ignore continuation points on large folders. Closing these requires the same ~150 to 200 orchestration lines this PR spent. Honest residual saving: about 100 lines of explanatory comments, under 5 percent, and stripping those from rollback-ordering code would be a net negative under this repo's correctness-first priority.

**Hindsight tax.** At least four load-bearing guards (duplicate sibling BrowseNames, dictionary key dedup, string-space dictionary matching, per-attempt claim scoping) exist only because review rounds found them. A from-scratch minimal build should be expected to ship without several and rediscover them as production incidents.

**The old loader did not meet the bar cheaply; it did not meet it at all.** Bad browse statuses silently produced zero children with no log, failed loads leaked claims with no monitored items behind them, and dynamic properties from failed attempts were permanently lost on retry. This PR surfaced and fixed silent failures rather than adding complexity to a working design.

### Conclusion

The ~2,000 production LOC is approximately the price of the reliability bar (full-retry convergence with an app-owned object graph), not gold-plating on top of batch loading. About two thirds of the PR is tests, which are the durable asset that made the review fixes verifiable. No lean rewrite is planned; the only remaining simplification candidate is replacing item-level subscription healing with reconnect-on-unhealthy, which predates this PR and belongs in its own PR.

RicoSuter:
## Measurement: this branch is the one that wins the IO

I instrumented the browse path (a temporary counter around `session.BrowseAsync` / `BrowseNextAsync`) and ran the same address space through master, this PR, and #357, so the comparison is measured rather than argued.

**Harness:** in-process `OpcUaTestServer`, 241 subjects (40 machines x 5 sensors), ~1000 value properties. Server advertises `MaxNodesPerBrowse = 4000` and `MaxBrowseContinuationPoints = 100`.

| variant | browse round-trips | nodes submitted | wall clock |
|---|---|---|---|
| master | 1,083 | 1,083 | 523 ms |
| **this PR (#313)** | **9** | 1,084 | 321 ms |
| #357 (four-phase loader) | 16 | 1,084 | 321 ms |
| #357 with the continuation cap disabled | 9 | 1,084 | 333 ms |

Wall clock is not a usable signal here: the transport is loopback, so RTT is near zero and the numbers are noise. Round-trips are the thing that scales with latency. At 5 ms RTT master is roughly 5.4 s of browse latency against 45 ms for this branch; at 20 ms WAN it is roughly 22 s against 180 ms.

**The 120x reduction belongs to this branch.** #357 does not add IO wins on top of it; it costs 7 extra round-trips, all of them from the unconditional continuation-point cap introduced in #397.

## What this means for the stack

The four-phase staged loader in #357 was built to give per-load structural atomicity: discover fully off-tree, then commit. Reviewing it against the registry APIs, that model is not reachable today. `RegisteredSubject.Children` is written only by `SubjectRegistry.HandleLifecycleChange`, so a fabricated detached view always has empty `Children`, and the loader reads `.Children` in three places to decide whether to reuse an existing child subject. Off-tree discovery therefore silently loses subject reuse and logs a misleading "the parent type must instantiate this property in its constructor" message.

Issue #320 had already scoped this out:

> **Out of scope:** structural atomicity for `DynamicSubject` roots (where dynamic properties are added during load). That requires a heavier "staging subject" design and should be tracked separately if needed.

and prescribed exactly what this branch already does: defer only the value-bearing root mutations, keep subjects on-tree during discovery so `.Children` resolves. That is `OpcUaLoadContext.QueueOrApplySetValue` plus the eager `AddFallbackContext` in `RegisterStagedSubject`.

So the plan is to carry this branch forward as the loader, after the two lower pieces of the #357 stack land:

1. **#398** (subscription gating) merges as-is plus a two-line fix. This branch has no `_callbacksEnabled` gate at all, and `SubscriptionManager` here admits its post-registration sweep leaves the window open.
2. **#397** (batched session primitives) merges with the continuation-point cap made conditional on `maxReferencesPerNode != 0`, plus an adaptive fallback on `BadNoContinuationPoints`. Today the cap is unconditional, which is what costs #357 the 7 round-trips. This branch has no cap at all, so it has no protection when `MaxReferencesPerNode` is configured non-zero. Neither extreme is right.
3. **#357** closes. Its model-independent tests (batching, dedup, attributes, dictionary reuse, root path resolution) get carried over here.
4. This PR is then refactored on top: the `Apply`/`Dispose` rollback bookkeeping goes away along with the failure-path tests that only exist to cover it, and the loader adopts #397's browse primitives and browse-completion contract instead of its own copies.

Net effect on this PR is mostly deletion. The eager-attach plus deferred-root-burst model stays exactly as it is.

RicoSuter:
Closing in favour of #414, which carries this branch's loader model forward.

#414 keeps what mattered here: eager attach during discovery so `.Children` resolves and subject reuse works, the deferred root value burst, the claims-before-root-assignments ordering at commit, the staged-subject rollback, and the three-file split. It verifies at the same 9 browse round-trips this branch achieved.

**Correction to my earlier comment on this PR.** I wrote that the refactor would be "mostly deletion". That was wrong, and it was based on misreading `OpcUaSubjectLoaderFailureTests` as scaffolding for the four-phase design. It is not: all eight tests cover this branch's own contract (no registry orphans after a failed load, pre-existing ownership retained across a mid-commit throw, clean retry, nested staged rollback, permanent-bad-status skip). They carried over intact.

What did change on the way to #414:

- The three protocol files here (`OpcUaSessionExtensions`, `OpcUaStatusCodeClassifier`, `OpcUaTransientServiceException`) were replaced by the reviewed versions from #397, which add split-and-retry and the browse-completion contract.
- Root path resolution gained a null guard: `ExpandedNodeId.ToNodeId` returns null when a segment's namespace URI is absent from the session's `NamespaceTable`, and master browsed that null NodeId instead of reporting the root as unresolvable.
- The branch was 18 commits behind master, which had since rewritten four files it touches, so #414 was built on current master rather than rebasing this.

PR #315, opened 2026-05-19
Enforce mutually exclusive subject-type classification (IInterceptorSubject wins)

RicoSuter:
## Summary

Tightens `Namotion.Interceptor.Tracking`'s subject-type classifiers so the three predicates `IsSubjectReferenceType`, `IsSubjectCollectionType`, and `IsSubjectDictionaryType` are mutually exclusive for any given type, and codifies the rule that an `IInterceptorSubject`-implementing type is always a Reference regardless of any container interfaces it also happens to implement.

Additionally, broadens runtime dispatch across the solution so read-only collection and dictionary types (`IReadOnlyList<T>`, `IReadOnlyCollection<T>`, `IReadOnlyDictionary<K,V>`, `ImmutableArray<T>`, custom IReadOnlyDictionary implementations, etc.) are fully supported in lifecycle tracking, registry, connectors, and JSON serialization. All connectors (OPC UA, MQTT, WebSocket, MCP) benefit automatically since they use `property.Children` from the registry.

### Rationale

Subject classes express their children through their declared `[InterceptorSubject]` partial properties; an incidental `IEnumerable<T>` or `IDictionary<,>` implementation should not change that. Previously a hybrid like `class Foo : IInterceptorSubject, IDictionary<string, Bar>` would classify as Dictionary, breaking downstream `else if` chains in connectors that depend on a single canonical category per type.

### Type classifier rules (mutually exclusive)

- **Reference**: `IInterceptorSubject` always wins. Otherwise: a plain interface or `object` (a single subject reference).
- **Collection**: an `IEnumerable<T>` where `T` is reference-shaped (interface/object/IInterceptorSubject) and not itself an enumerable.
- **Dictionary**: must implement `IDictionary`, `IDictionary<K,V>`, or `IReadOnlyDictionary<K,V>` AND its `IEnumerable<T>` element type must be `KeyValuePair<K, V>` with V reference-shaped. A bare `IEnumerable<KVP<,>>` or `List<KVP<,>>` is NOT dict-classified, because the runtime handler dispatches via `IDictionary` and would silently no-op such values. Classifier and handler agree.

Element-type checks short-circuit on `IInterceptorSubject` so a hybrid like `class Foo : IInterceptorSubject, IEnumerable<X>` still classifies as a valid element inside `List<Foo>` (the outer list is a subject collection).

### Runtime handlers

Hot paths use an inline switch-on-shape (master-equivalent perf, no abstraction overhead):
- `LifecycleInterceptor.FindSubjectsInProperty`
- `RegisteredSubjectProperty.BuildCollectionPositions`
- `SubjectItemsUpdateFactory.BuildDictionaryDiff` / `BuildDictionaryComplete` (accept `IDictionary` directly; zero-alloc passthrough for the common case)
- `SubjectValueConvert.ToSubjectList` / `ToSubjectMutableList` / `ToSubjectDictionary`

`SubjectValueLookup` (renamed from `SubjectValueVisitor`) is a small static helper class with `FindCollectionSubjectAt`, `FindDictionarySubjectAt`, and `TryGetKvpSubjectEntry`. Used by `PathExtensions` for keyed lookups and by `SubjectValueConvert` for the read-only-dict KVP fallback.

### Implementation notes

- `IsSubjectReferenceType` returns `true` for any `IInterceptorSubject`-derived type immediately; otherwise uses `CanDirectlyHoldSubject` (interface/object/IIS that is not IEnumerable) and ensures the two container predicates return `false`
- `IsSubjectCollectionType` / `IsSubjectDictionaryType` early-exit `false` for any `IInterceptorSubject`-derived type. For non-subjects: generic interfaces over candidate element content classify; non-generic `ICollection`/`IDictionary` fallback handles legacy types (`ArrayList`, `Hashtable`)
- `IsCandidateElementType` predicate short-circuits on `IInterceptorSubject` so hybrid container-subjects still qualify as collection elements; `CanDirectlyHoldSubject` retains the IEnumerable exclusion for self-classification
- `GetEnumerablesIncludingSelf` includes a bare `IEnumerable<T>` so interface-typed properties classify correctly
- `CollectionDiffBuilder.GetDictionaryChanges` accepts `IDictionary` directly and uses `dict.Contains(key)` plus the indexer for lookups, eliminating the prior intermediate List materialization
- `TryGetKvpSubjectEntry` uses cached `PropertyInfo.GetValue` delegates per closed `KeyValuePair<K,V>` type, with a TODO marker noting expression-tree compilation as the upgrade path if a real workload exercises this slow path
- `SubjectRegistryJsonExtensions` materializes the `OfType<IInterceptorSubject>` result once into a list and branches on `Count`, avoiding double-enumeration

**Call sites updated**:
- `LifecycleInterceptor.FindSubjectsInProperty`: inline switch over IInterceptorSubject / IDictionary / ICollection / string / IEnumerable
- `RegisteredSubjectProperty.BuildCollectionPositions`: inline IList fast path + ICollection + IEnumerable fallback
- `PathExtensions`: delegates to `SubjectValueLookup.FindCollectionSubjectAt` / `FindDictionarySubjectAt`
- `SubjectUpdateFactory` / `SubjectItemsUpdateFactory` / `SubjectItemsUpdateApplier`: use IDictionary direct dispatch
- `SubjectRegistryJsonExtensions`: uses `IEnumerable` with single-pass `OfType<>` materialization
- `SubjectPathResolver` / `SubjectPathResolverExtensions` (HomeBlaze): use shared classifier

### Tests

- 789 tests pass (Tracking 305 + Registry 110 + Connectors 374)
- Public API snapshot regenerated (`VerifyChecksTests.PublicApi`) to reflect the `SubjectValueVisitor` -> `SubjectValueLookup` rename and the `[NotNullWhen]` annotation on `TryGetKvpSubjectEntry`
- New tests cover mutual-exclusion classification, hybrid `IInterceptorSubject`-plus-container types (including `List<HybridSubjectGenericList>`), self-referential subject shapes, read-only dict wrappers (lifecycle), and an Apply-side connector round-trip with `ReadOnlyDictionaryWrapper`

### DIFF
--- a/src/Namotion.Interceptor.AspNetCore/Extensions/SubjectRegistryJsonExtensions.cs
+++ b/src/Namotion.Interceptor.AspNetCore/Extensions/SubjectRegistryJsonExtensions.cs
@@ -100,32 +100,7 @@ public static class SubjectRegistryJsonExtensions
         {
             // add static properties
             var propertyName = GetJsonPropertyName(subject, property.Value, jsonSerializerOptions);
-            var value = property.Value.GetValue?.Invoke(subject);
-            if (value is IInterceptorSubject childProxy)
-            {
-                obj[propertyName] = childProxy.ToJsonObject(jsonSerializerOptions);
-            }
-            else if (value is IEnumerable enumerable and not string)
-            {
-                // Materialize once so the Count check and the foreach share a single pass over
-                // potentially lazy / one-shot enumerables.
-                var subjectChildren = enumerable.OfType<IInterceptorSubject>().ToList();
-                if (subjectChildren.Count > 0)
-                {
-                    var children = new JsonArray();
-                    foreach (var item in subjectChildren)
-                        children.Add(item.ToJsonObject(jsonSerializerOptions));
-                    obj[propertyName] = children;
-                }
-                else
-                {
-                    obj[propertyName] = JsonValue.Create(value);
-                }
-            }
-            else
-            {
-                obj[propertyName] = JsonValue.Create(value);
-            }
+            obj[propertyName] = ValueToJsonNode(property.Value.GetValue?.Invoke(subject), jsonSerializerOptions);
         }
 
         var registeredSubject = subject.TryGetRegisteredSubject();
@@ -134,40 +109,39 @@ public static class SubjectRegistryJsonExtensions
             // add dynamic properties
             foreach (var property in registeredSubject
                 .Properties
-                .Where(p => p.HasGetter && 
+                .Where(p => p.HasGetter &&
                             subject.Properties.ContainsKey(p.Name) == false))
             {
                 var propertyName = property.GetJsonPropertyName(jsonSerializerOptions);
-                var value = property.GetValue();
-                if (value is IInterceptorSubject childProxy)
-                {
-                    obj[propertyName] = childProxy.ToJsonObject(jsonSerializerOptions);
-                }
-                else if (value is IEnumerable enumerable and not string)
-                {
-                    // Materialize once so the Count check and the foreach share a single pass
-                    // over potentially lazy / one-shot enumerables.
-                    var subjectChildren = enumerable.OfType<IInterceptorSubject>().ToList();
-                    if (subjectChildren.Count > 0)
-                    {
-                        var children = new JsonArray();
-                        foreach (var item in subjectChildren)
-                            children.Add(item.ToJsonObject(jsonSerializerOptions));
-                        obj[propertyName] = children;
-                    }
-                    else
-                    {
-                        obj[propertyName] = JsonValue.Create(value);
-                    }
-                }
-                else
-                {
-                    obj[propertyName] = JsonValue.Create(value);
-                }
+                obj[propertyName] = ValueToJsonNode(property.GetValue(), jsonSerializerOptions);
             }
         }
 
         return obj;
+    }
+
+    private static JsonNode? ValueToJsonNode(object? value, JsonSerializerOptions jsonSerializerOptions)
+    {
+        if (value is IInterceptorSubject childProxy)
+        {
+            return childProxy.ToJsonObject(jsonSerializerOptions);
+        }
+
+        if (value is IEnumerable enumerable and not string)
+        {
+            // Materialize once so the Count check and the foreach share a single pass over
+            // potentially lazy / one-shot enumerables.
+            var subjectChildren = enumerable.OfType<IInterceptorSubject>().ToList();
+            if (subjectChildren.Count > 0)
+            {
+                var children = new JsonArray();
+                foreach (var item in subjectChildren)
+                    children.Add(item.ToJsonObject(jsonSerializerOptions));
+                return children;
+            }
+        }
+
+        return JsonValue.Create(value);
     }
 
     private static string GetJsonPropertyName(IInterceptorSubject subject, SubjectPropertyMetadata property, JsonSerializerOptions jsonSerializerOptions)
--- a/src/Namotion.Interceptor.AspNetCore/Extensions/SubjectRegistryJsonExtensions.cs
+++ b/src/Namotion.Interceptor.AspNetCore/Extensions/SubjectRegistryJsonExtensions.cs
@@ -105,14 +105,22 @@ public static class SubjectRegistryJsonExtensions
             {
                 obj[propertyName] = childProxy.ToJsonObject(jsonSerializerOptions);
             }
-            else if (value is IEnumerable enumerable and not string && enumerable.OfType<IInterceptorSubject>().Any())
+            else if (value is IEnumerable enumerable and not string)
             {
-                var children = new JsonArray();
-                foreach (var arrayProxyItem in enumerable.OfType<IInterceptorSubject>())
+                // Materialize once so the Count check and the foreach share a single pass over
+                // potentially lazy / one-shot enumerables.
+                var subjectChildren = enumerable.OfType<IInterceptorSubject>().ToList();
+                if (subjectChildren.Count > 0)
                 {
-                    children.Add(arrayProxyItem.ToJsonObject(jsonSerializerOptions));
+                    var children = new JsonArray();
+                    foreach (var item in subjectChildren)
+                        children.Add(item.ToJsonObject(jsonSerializerOptions));
+                    obj[propertyName] = children;
                 }
-                obj[propertyName] = children;
+                else
+                {
+                    obj[propertyName] = JsonValue.Create(value);
+                }
             }
             else
             {
@@ -135,14 +143,22 @@ public static class SubjectRegistryJsonExtensions
                 {
                     obj[propertyName] = childProxy.ToJsonObject(jsonSerializerOptions);
                 }
-                else if (value is IEnumerable enumerable and not string && enumerable.OfType<IInterceptorSubject>().Any())
+                else if (value is IEnumerable enumerable and not string)
                 {
-                    var children = new JsonArray();
-                    foreach (var arrayProxyItem in enumerable.OfType<IInterceptorSubject>())
+                    // Materialize once so the Count check and the foreach share a single pass
+                    // over potentially lazy / one-shot enumerables.
+                    var subjectChildren = enumerable.OfType<IInterceptorSubject>().ToList();
+                    if (subjectChildren.Count > 0)
                     {
-                        children.Add(arrayProxyItem.ToJsonObject(jsonSerializerOptions));
+                        var children = new JsonArray();
+                        foreach (var item in subjectChildren)
+                            children.Add(item.ToJsonObject(jsonSerializerOptions));
+                        obj[propertyName] = children;
                     }
-                    obj[propertyName] = children;
+                    else
+                    {
+                        obj[propertyName] = JsonValue.Create(value);
+                    }
                 }
                 else
                 {
--- a/src/Namotion.Interceptor.AspNetCore/Extensions/SubjectRegistryJsonExtensions.cs
+++ b/src/Namotion.Interceptor.AspNetCore/Extensions/SubjectRegistryJsonExtensions.cs
@@ -105,10 +105,10 @@ public static class SubjectRegistryJsonExtensions
             {
                 obj[propertyName] = childProxy.ToJsonObject(jsonSerializerOptions);
             }
-            else if (value is ICollection collection && collection.OfType<IInterceptorSubject>().Any())
+            else if (value is IEnumerable enumerable and not string && enumerable.OfType<IInterceptorSubject>().Any())
             {
                 var children = new JsonArray();
-                foreach (var arrayProxyItem in collection.OfType<IInterceptorSubject>())
+                foreach (var arrayProxyItem in enumerable.OfType<IInterceptorSubject>())
                 {
                     children.Add(arrayProxyItem.ToJsonObject(jsonSerializerOptions));
                 }
@@ -135,10 +135,10 @@ public static class SubjectRegistryJsonExtensions
                 {
                     obj[propertyName] = childProxy.ToJsonObject(jsonSerializerOptions);
                 }
-                else if (value is ICollection collection && collection.OfType<IInterceptorSubject>().Any())
+                else if (value is IEnumerable enumerable and not string && enumerable.OfType<IInterceptorSubject>().Any())
                 {
                     var children = new JsonArray();
-                    foreach (var arrayProxyItem in collection.OfType<IInterceptorSubject>())
+                    foreach (var arrayProxyItem in enumerable.OfType<IInterceptorSubject>())
                     {
                         children.Add(arrayProxyItem.ToJsonObject(jsonSerializerOptions));
                     }
@@ -207,7 +207,7 @@ public static class SubjectRegistryJsonExtensions
 
                     if (subject.Properties.TryGetValue(nextSegment, out var property))
                     {
-                        var collection = property.GetValue?.Invoke(subject) as ICollection;
+                        var collection = property.GetValue?.Invoke(subject) as IEnumerable;
                         var child = collection?.OfType<IInterceptorSubject>().ElementAt(index);
                         if (child is not null)
                         {
--- a/src/Namotion.Interceptor.Connectors/Updates/Internal/CollectionDiffBuilder.cs
+++ b/src/Namotion.Interceptor.Connectors/Updates/Internal/CollectionDiffBuilder.cs
@@ -1,5 +1,3 @@
-using System.Collections;
-
 namespace Namotion.Interceptor.Connectors.Updates.Internal;
 
 /// <summary>
@@ -148,14 +146,14 @@ internal sealed class CollectionDiffBuilder
     /// <summary>
     /// Builds dictionary change operations.
     /// </summary>
-    /// <param name="oldDictionary">The old dictionary.</param>
-    /// <param name="newDictionary">The new dictionary.</param>
+    /// <param name="oldEntries">The old dictionary entries.</param>
+    /// <param name="newEntries">The new dictionary entries.</param>
     /// <param name="operations">Output: structural operations (Insert/Remove), or null if none.</param>
     /// <param name="newItemsToProcess">Output: items that are new and need full processing, or null if none.</param>
     /// <param name="removedKeys">Output: keys that were removed, or null if none.</param>
     public void GetDictionaryChanges(
-        IDictionary? oldDictionary,
-        IDictionary newDictionary,
+        IReadOnlyList<(object key, IInterceptorSubject subject)>? oldEntries,
+        IReadOnlyList<(object key, IInterceptorSubject subject)> newEntries,
         out List<SubjectCollectionOperation>? operations,
         out List<(object key, IInterceptorSubject item)>? newItemsToProcess,
         out HashSet<object>? removedKeys)
@@ -165,48 +163,41 @@ internal sealed class CollectionDiffBuilder
         removedKeys = null;
 
         _oldKeys.Clear();
-        if (oldDictionary is not null)
+        Dictionary<object, IInterceptorSubject>? oldLookup = null;
+        if (oldEntries is not null)
         {
-            foreach (var key in oldDictionary.Keys)
+            oldLookup = new Dictionary<object, IInterceptorSubject>(oldEntries.Count);
+            foreach (var (key, subject) in oldEntries)
+            {
                 _oldKeys.Add(key);
+                oldLookup[key] = subject;
+            }
         }
 
-        // Track removed keys - start with all old keys, remove ones that still exist with same value
         HashSet<object>? keysToRemove = _oldKeys.Count > 0 ? [.._oldKeys] : null;
 
-        foreach (DictionaryEntry entry in newDictionary)
+        foreach (var (key, newItem) in newEntries)
         {
-            var key = entry.Key;
-            if (entry.Value is not IInterceptorSubject newItem)
-                continue;
-
             if (_oldKeys.Contains(key))
             {
-                // Key exists in both - check if VALUE changed (different object reference)
-                var oldValue = oldDictionary![key];
+                var oldValue = oldLookup![key];
                 if (ReferenceEquals(oldValue, newItem))
                 {
-                    // Same object - key is not removed, not a new item
                     keysToRemove?.Remove(key);
                 }
                 else
                 {
-                    // Different object at same key - treat as Remove + Insert
-                    // Key stays in keysToRemove (will generate Remove)
-                    // Add to newItemsToProcess (will generate Insert)
                     newItemsToProcess ??= [];
                     newItemsToProcess.Add((key, newItem));
                 }
             }
             else
             {
-                // New key
                 newItemsToProcess ??= [];
                 newItemsToProcess.Add((key, newItem));
             }
         }
 
-        // Only return removedKeys if there are actually keys to remove
         if (keysToRemove is { Count: > 0 })
         {
             removedKeys = keysToRemove;
--- a/src/Namotion.Interceptor.Connectors/Updates/Internal/CollectionDiffBuilder.cs
+++ b/src/Namotion.Interceptor.Connectors/Updates/Internal/CollectionDiffBuilder.cs
@@ -14,6 +14,7 @@ internal sealed class CollectionDiffBuilder
     private readonly Dictionary<IInterceptorSubject, int> _oldCommonIndexMap = new();
     private readonly List<IInterceptorSubject> _oldCommonOrder = [];
     private readonly List<IInterceptorSubject> _newCommonOrder = [];
+    private readonly List<(object key, IInterceptorSubject item)> _commonDictionaryItems = [];
 
     /// <summary>
     /// Builds collection change operations.
@@ -192,6 +193,7 @@ internal sealed class CollectionDiffBuilder
                 if (ReferenceEquals(oldItem, newItem))
                 {
                     keysToRemove?.Remove(key);
+                    _commonDictionaryItems.Add((key, newItem));
                 }
                 else
                 {
@@ -227,6 +229,12 @@ internal sealed class CollectionDiffBuilder
     public IReadOnlyList<IInterceptorSubject> GetCommonItems() => _newCommonOrder;
 
     /// <summary>
+    /// Gets dictionary entries whose key and value (by reference) exist in both old and new.
+    /// Populated by <see cref="GetDictionaryChanges"/>.
+    /// </summary>
+    public IReadOnlyList<(object key, IInterceptorSubject item)> GetCommonDictionaryItems() => _commonDictionaryItems;
+
+    /// <summary>
     /// Clears the builder for reuse. Call before returning to pool.
     /// </summary>
     public void Clear()
@@ -236,5 +244,6 @@ internal sealed class CollectionDiffBuilder
         _oldCommonIndexMap.Clear();
         _oldCommonOrder.Clear();
         _newCommonOrder.Clear();
+        _commonDictionaryItems.Clear();
     }
 }
--- a/src/Namotion.Interceptor.Connectors/Updates/Internal/CollectionDiffBuilder.cs
+++ b/src/Namotion.Interceptor.Connectors/Updates/Internal/CollectionDiffBuilder.cs
@@ -1,3 +1,5 @@
+using System.Collections;
+
 namespace Namotion.Interceptor.Connectors.Updates.Internal;
 
 /// <summary>
@@ -12,7 +14,6 @@ internal sealed class CollectionDiffBuilder
     private readonly Dictionary<IInterceptorSubject, int> _oldCommonIndexMap = new();
     private readonly List<IInterceptorSubject> _oldCommonOrder = [];
     private readonly List<IInterceptorSubject> _newCommonOrder = [];
-    private readonly Dictionary<object, IInterceptorSubject> _oldLookup = new();
 
     /// <summary>
     /// Builds collection change operations.
@@ -144,16 +145,19 @@ internal sealed class CollectionDiffBuilder
     }
 
     /// <summary>
-    /// Builds dictionary change operations.
+    /// Builds dictionary change operations. Iterates <paramref name="oldDictionary"/> and
+    /// <paramref name="newDictionary"/> directly; non-subject entries are filtered inline so the
+    /// caller can pass any <see cref="IDictionary"/> (including passthrough of the user's value)
+    /// without a separate materialization step.
     /// </summary>
-    /// <param name="oldEntries">The old dictionary entries.</param>
-    /// <param name="newEntries">The new dictionary entries.</param>
+    /// <param name="oldDictionary">The old dictionary value, or null if none.</param>
+    /// <param name="newDictionary">The new dictionary value.</param>
     /// <param name="operations">Output: structural operations (Insert/Remove), or null if none.</param>
     /// <param name="newItemsToProcess">Output: items that are new and need full processing, or null if none.</param>
     /// <param name="removedKeys">Output: keys that were removed, or null if none.</param>
     public void GetDictionaryChanges(
-        IReadOnlyList<(object key, IInterceptorSubject subject)>? oldEntries,
-        IReadOnlyList<(object key, IInterceptorSubject subject)> newEntries,
+        IDictionary? oldDictionary,
+        IDictionary newDictionary,
         out List<SubjectCollectionOperation>? operations,
         out List<(object key, IInterceptorSubject item)>? newItemsToProcess,
         out HashSet<object>? removedKeys)
@@ -162,25 +166,30 @@ internal sealed class CollectionDiffBuilder
         newItemsToProcess = null;
         removedKeys = null;
 
-        _oldLookup.Clear();
-        if (oldEntries is not null)
+        // Track removed keys - start with all old subject-valued keys, remove ones that still exist
+        // with the same value below.
+        HashSet<object>? keysToRemove = null;
+        if (oldDictionary is not null)
         {
-            foreach (var (key, subject) in oldEntries)
+            foreach (DictionaryEntry entry in oldDictionary)
             {
-                _oldLookup[key] = subject;
+                if (entry.Value is IInterceptorSubject)
+                {
+                    keysToRemove ??= [];
+                    keysToRemove.Add(entry.Key);
+                }
             }
         }
 
-        // Track removed keys - start with all old keys, remove ones that still exist with same value
-        HashSet<object>? keysToRemove = _oldLookup.Count > 0 ? [.._oldLookup.Keys] : null;
-
-        foreach (var (key, newItem) in newEntries)
+        foreach (DictionaryEntry entry in newDictionary)
         {
-            if (_oldLookup.ContainsKey(key))
+            if (entry.Value is not IInterceptorSubject newItem) continue;
+            var key = entry.Key;
+
+            if (oldDictionary is not null && oldDictionary.Contains(key) &&
+                oldDictionary[key] is IInterceptorSubject oldItem)
             {
-                // Key exists in both - check if VALUE changed (different object reference)
-                var oldValue = _oldLookup[key];
-                if (ReferenceEquals(oldValue, newItem))
+                if (ReferenceEquals(oldItem, newItem))
                 {
                     keysToRemove?.Remove(key);
                 }
@@ -227,6 +236,5 @@ internal sealed class CollectionDiffBuilder
         _oldCommonIndexMap.Clear();
         _oldCommonOrder.Clear();
         _newCommonOrder.Clear();
-        _oldLookup.Clear();
     }
 }
--- a/src/Namotion.Interceptor.Connectors/Updates/Internal/CollectionDiffBuilder.cs
+++ b/src/Namotion.Interceptor.Connectors/Updates/Internal/CollectionDiffBuilder.cs
@@ -11,10 +11,10 @@ internal sealed class CollectionDiffBuilder
     // Reusable containers
     private readonly Dictionary<IInterceptorSubject, int> _oldIndexMap = new();
     private readonly Dictionary<IInterceptorSubject, int> _newIndexMap = new();
-    private readonly Dictionary<IInterceptorSubject, int> _oldCommonIndexMap = new();
-    private readonly List<IInterceptorSubject> _oldCommonOrder = [];
-    private readonly List<IInterceptorSubject> _newCommonOrder = [];
-    private readonly List<(object key, IInterceptorSubject item)> _commonDictionaryItems = [];
+    private readonly Dictionary<IInterceptorSubject, int> _oldRetainedIndexMap = new();
+    private readonly List<IInterceptorSubject> _oldRetainedOrder = [];
+    private readonly List<IInterceptorSubject> _newRetainedOrder = [];
+    private readonly List<(object key, IInterceptorSubject item)> _retainedDictionaryItems = [];
 
     /// <summary>
     /// Builds collection change operations.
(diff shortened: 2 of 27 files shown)


src/Namotion.Interceptor.AspNetCore/Extensions/SubjectRegistryJsonExtensions.cs
```diff
@@ -105,10 +105,10 @@ public static JsonObject ToJsonObject(this IInterceptorSubject subject, JsonSeri
             {
                 obj[propertyName] = childProxy.ToJsonObject(jsonSerializerOptions);
             }
-            else if (value is ICollection collection && collection.OfType<IInterceptorSubject>().Any())
+            else if (value is IEnumerable enumerable and not string && enumerable.OfType<IInterceptorSubject>().Any())
```
RicoSuter: do we need fast path here (check ICollection and IEnumerable) to keep performance?

src/Namotion.Interceptor.Connectors/Updates/Internal/CollectionDiffBuilder.cs
```diff
@@ -165,48 +163,41 @@ public void GetDictionaryChanges(
         removedKeys = null;
 
         _oldKeys.Clear();
-        if (oldDictionary is not null)
+        Dictionary<object, IInterceptorSubject>? oldLookup = null;
+        if (oldEntries is not null)
         {
-            foreach (var key in oldDictionary.Keys)
+            oldLookup = new Dictionary<object, IInterceptorSubject>(oldEntries.Count);
```
RicoSuter: More allocs now?

src/Namotion.Interceptor.Connectors/Updates/Internal/CollectionDiffBuilder.cs
```diff
@@ -165,48 +163,41 @@ public void GetDictionaryChanges(
...
-                var oldValue = oldDictionary![key];
+                var oldValue = oldLookup![key];
                 if (ReferenceEquals(oldValue, newItem))
                 {
-                    // Same object - key is not removed, not a new item
                     keysToRemove?.Remove(key);
                 }
                 else
                 {
-                    // Different object at same key - treat as Remove + Insert
```
RicoSuter, line 194: should we really remove all these comments in the class here?

src/Namotion.Interceptor.Connectors/Updates/Internal/SubjectItemsUpdateFactory.cs
```diff
@@ -47,17 +46,17 @@ internal static void BuildCollectionComplete(
...
     {
         update.Kind = SubjectPropertyUpdateKind.Collection;
 
-        if (newCollection is null)
+        if (newCollectionValue is null)
             return;
 
-        var oldItems = oldCollection as IReadOnlyList<IInterceptorSubject> ?? oldCollection?.ToList() ?? [];
-        var newItems = newCollection as IReadOnlyList<IInterceptorSubject> ?? newCollection.ToList();
+        var oldItems = oldCollectionValue is not null ? SubjectValueConverter.ToSubjectList(oldCollectionValue) : (IReadOnlyList<IInterceptorSubject>)[];
```
RicoSuter: Use some Collection.Empty type here?

src/Namotion.Interceptor.Connectors/Updates/Internal/SubjectValueConvert.cs
```diff
@@ -0,0 +1,64 @@
...
+
+internal static class SubjectValueConverter
+{
+    internal static IReadOnlyList<IInterceptorSubject> ToSubjectList(object value)
+    {
+        if (value is IReadOnlyList<IInterceptorSubject> readOnlyList)
+            return readOnlyList;
+
+        if (value is IEnumerable<IInterceptorSubject> typedEnumerable)
+            return typedEnumerable.ToList();
```
RicoSuter, line 38: will this introduce mem alloc regressions? in what paths?

src/Namotion.Interceptor.Connectors/Updates/Internal/SubjectValueConverter.cs
```diff
@@ -0,0 +1,64 @@
+using System.Runtime.CompilerServices;
+using Namotion.Interceptor.Tracking;
+
+namespace Namotion.Interceptor.Connectors.Updates.Internal;
+
+internal static class SubjectValueConverter
```
RicoSuter: Rename to SubjectValueConvert (similar to Convert class in .NET)?

src/Namotion.Interceptor.Registry/Paths/PathExtensions.cs
```diff
@@ -188,15 +188,11 @@ public static (RegisteredSubjectProperty Property, object? Index)? TryGetPropert
             return value as IInterceptorSubject;
         }
 
-        if (property.IsSubjectDictionary && value is IDictionary dictionary)
-        {
-            return dictionary[index] as IInterceptorSubject;
-        }
+        if (property.IsSubjectDictionary)
+            return SubjectValueVisitor.FindSubjectAt(value, isDictionaryType: true, index);
```
RicoSuter: would it make sense to have a method for dict and one for collection and avoid bool parameter?

src/Namotion.Interceptor.Tracking.Tests/SubjectPropertyTypeExtensionsTests.cs
```diff
@@ -0,0 +1,227 @@
...
+    [InlineData(typeof(List<object>),                                       false, true,  false)]
+    [InlineData(typeof(List<IInterceptorSubject>),                          false, true,  false)]
+    [InlineData(typeof(Person[]),                                           false, true,  false)]
+    [InlineData(typeof(List<Person>),                                       false, true,  false)]
+    [InlineData(typeof(IEnumerable<Person>),                                false, true,  false)]
+    [InlineData(typeof(ImmutableArray<Person>),                             false, true,  false)]
+    [InlineData(typeof(ImmutableArray<object>),                             false, true,  false)]
+    [InlineData(typeof(ImmutableArray<IComparable>),                        false, true,  false)]
+    [InlineData(typeof(ArrayList),                                          false, true,  false)]
+    [InlineData(typeof(IReadOnlyCollection<Person>),                         false, true,  false)]
```
RicoSuter: fix spacings

src/Namotion.Interceptor.Tracking/Lifecycle/LifecycleInterceptor.cs
```diff
@@ -370,7 +370,7 @@ newValue is not (null or IInterceptorSubject or ICollection or IDictionary))
 
                 // Refresh child index metadata for retained subjects whose
                 // positions may have shifted in the new collection.
-                if (newValue is ICollection && oldTouchedSubjects.Overlaps(newTouchedSubjects))
+                if (newValue is IEnumerable && oldTouchedSubjects.Overlaps(newTouchedSubjects))
```
RicoSuter, line 372: is switching to enumerable here safe?

src/Namotion.Interceptor.AspNetCore/Extensions/SubjectRegistryJsonExtensions.cs
RicoSuter: No fast path needed here. Both .OfType<IInterceptorSubject>().Any() and the iteration use LINQ, which calls GetEnumerator() via IEnumerable internally regardless of whether the source is ICollection or IEnumerable. An ICollection check before IEnumerable would not change codegen or perf for the LINQ operators.

src/Namotion.Interceptor.Connectors/Updates/Internal/CollectionDiffBuilder.cs
RicoSuter: Fixed: pooled the lookup dictionary as a reusable field on CollectionDiffBuilder (like the existing _oldKeys HashSet). Zero extra allocation now since the builder is already pooled via ChangeBuilderPool.

src/Namotion.Interceptor.Connectors/Updates/Internal/CollectionDiffBuilder.cs
RicoSuter: Restored all inline comments in 9f4b454.

src/Namotion.Interceptor.Connectors/Updates/Internal/SubjectItemsUpdateFactory.cs
RicoSuter: The `[]` collection expression already compiles to `Array.Empty<T>()` in modern C#, which is a cached singleton with zero allocation. The cast to `IReadOnlyList<>` is free via covariance. So this is already optimal.

src/Namotion.Interceptor.Connectors/Updates/Internal/SubjectValueConvert.cs
RicoSuter: No regression for common types. The method checks IReadOnlyList<IInterceptorSubject> first (zero-alloc, hits for T[], List<T>, ImmutableArray<T> via covariance), then IEnumerable<IInterceptorSubject> (.ToList() same as master), then falls back to the visitor only for non-generic collections like ArrayList. The master code had the same .ToList() call: `collection as IReadOnlyList<IInterceptorSubject> ?? collection.ToList()`. Benchmarks confirm zero regression.

src/Namotion.Interceptor.Connectors/Updates/Internal/SubjectValueConverter.cs
RicoSuter: Done in 9f4b454.

src/Namotion.Interceptor.Registry/Paths/PathExtensions.cs
RicoSuter: Done in 9f4b454. Split into VisitCollectionSubjects/VisitDictionarySubjects and FindCollectionSubjectAt/FindDictionarySubjectAt. Call sites are now self-documenting with no bool parameter.

src/Namotion.Interceptor.Tracking.Tests/SubjectPropertyTypeExtensionsTests.cs
RicoSuter: Fixed in 9f4b454.

src/Namotion.Interceptor.Tracking/Lifecycle/LifecycleInterceptor.cs
RicoSuter: Yes, safe. This gate only fires after two earlier guards: (1) CanContainSubjects<TProperty>() at line 293 filters out string-typed properties via JIT, (2) the early-exit at line 316 explicitly excludes strings with `|| lastProcessed is string`. A string value can never reach this line. The broadening from ICollection to IEnumerable is needed so read-only collections (IReadOnlyList<T>, ImmutableArray<T>) that don't implement ICollection still trigger position refresh for retained subjects.

src/Namotion.Interceptor.Connectors/Updates/Internal/SubjectItemsUpdateFactory.cs
```diff
@@ -229,13 +227,8 @@ internal static void BuildDictionaryDiff(
             }
 
             List<SubjectPropertyItemUpdate>? updates = null;
-            foreach (DictionaryEntry entry in newDict)
+            foreach (var (key, item) in newDict)
             {
-                var key = entry.Key;
-                if (entry.Value is not IInterceptorSubject item)
```
RicoSuter, line 235: check not needed?

src/Namotion.Interceptor.Tracking/SubjectPropertyTypeExtensions.cs
```diff
@@ -130,28 +154,82 @@ private static bool IsSubjectDictionaryTypeSlow(Type type)
...
+    // Non-recursive leaf check so self-referential types do not blow the stack.
+    private static bool IsElementSubjectReference(Type element)
+    {
+        if (!IsSubjectReferenceCandidate(element))
+            return false;
+
+        if (!typeof(IEnumerable).IsAssignableFrom(element))
+            return true;
+
+        foreach (var i in GetEnumerablesIncludingSelf(element))
```
RicoSuter: is this all needed everything returns false here, no? review the whole method? is it even needed or can it just be replaced by IsSubjectReferenceCandidate?

src/Namotion.Interceptor.Connectors/Updates/Internal/SubjectItemsUpdateFactory.cs
RicoSuter: The check is needed. This loop iterates all entries in the new dictionary and generates sparse property updates for existing items that have changes. Without the `newKeysSet` skip, items that are being inserted (already handled as Insert operations above) would also get a redundant sparse update, causing duplicate processing.

src/Namotion.Interceptor.Tracking/SubjectPropertyTypeExtensions.cs
RicoSuter: You're right. The foreach loop body (lines 195-208) only ever returns `false`, so the entire post-IEnumerable section collapses to a single check. The method can be simplified to:

```csharp
private static bool IsElementSubjectReference(Type element) =>
    IsSubjectReferenceCandidate(element) && !typeof(IEnumerable).IsAssignableFrom(element);
```

Traced all cases:
- `Person` (IIS): candidate=true, IEnumerable=false -> true (single reference)
- `IComparable`: candidate=true, IEnumerable=false -> true (could hold subject)
- `IEnumerable<Person>`: candidate=true, IEnumerable=true -> false (container, not ref)
- `IList<int>`: candidate=true, IEnumerable=true -> false (container)
- `int`: candidate=false -> false

Will simplify.

src/Namotion.Interceptor.Connectors/Updates/Internal/SubjectItemsUpdateFactory.cs
RicoSuter: Sorry, misunderstood the question. You're asking about the removed `if (entry.Value is not IInterceptorSubject item) continue;` check from master. Yes, safe to remove: `newDict` is now `IReadOnlyList<(object key, IInterceptorSubject subject)>` from `SubjectValueConvert.ToDictionaryEntries()`, which already filters to only subject entries upstream. The type system guarantees every `item` in the loop is an `IInterceptorSubject`.

src/Namotion.Interceptor.Connectors/Updates/Internal/CollectionDiffBuilder.cs
```diff
@@ -15,6 +13,7 @@ internal sealed class CollectionDiffBuilder
     private readonly List<IInterceptorSubject> _oldCommonOrder = [];
     private readonly List<IInterceptorSubject> _newCommonOrder = [];
     private readonly HashSet<object> _oldKeys = [];
```
RicoSuter: do we need to keep this old keys or can we use keys from look up? any performance gains from keeping it?

src/Namotion.Interceptor.Connectors/Updates/Internal/SubjectItemsUpdateFactory.cs
```diff
@@ -16,15 +15,15 @@ internal static class SubjectItemsUpdateFactory
     /// </summary>
     internal static void BuildCollectionComplete(
```
RicoSuter, line 17: any mem alloc or perf regressions here in this class?

src/Namotion.Interceptor.Connectors/Updates/Internal/CollectionDiffBuilder.cs
RicoSuter: Good catch. Removed _oldKeys and use _oldLookup.ContainsKey() + _oldLookup.Keys directly. Dictionary.ContainsKey() is the same O(1) hash lookup as HashSet.Contains(), and Dictionary.Keys returns a lightweight view (no allocation).

src/Namotion.Interceptor.Connectors/Updates/Internal/SubjectItemsUpdateFactory.cs
RicoSuter: No regression. SubjectValueConvert.ToSubjectList() checks IReadOnlyList<IInterceptorSubject> first (zero-alloc for T[], List<T>, ImmutableArray<T> via covariance), then IEnumerable<IInterceptorSubject> (.ToList(), same allocation as master), then falls back to the visitor only for non-generic types. ToDictionaryEntries() allocates a List<(object, IInterceptorSubject)> where master iterated IDictionary inline, but this runs only during connector update diffs (not the lifecycle hot path). Benchmarks confirmed: WriteToRegistrySubjects 2.0ms/0B, WriteToSource 1.14ms/0B, identical to master.

src/Namotion.Interceptor.Tracking/SubjectValueVisitor.cs
```diff
@@ -0,0 +1,96 @@
...
+using System.Runtime.CompilerServices;
+
+namespace Namotion.Interceptor.Tracking;
+
+/// <summary>
+/// Static helpers for locating subjects inside property values. Used by hot paths
+/// (LifecycleInterceptor, RegisteredSubjectProperty) which inline the dispatch switch
+/// themselves for best codegen, and by PathExtensions for keyed lookups.
+/// </summary>
+public static class SubjectValueVisitor
```
RicoSuter: Should rename class because it's not a visitor anymore, avoid utilities or helpers but this is what it is

src/Namotion.Interceptor.Connectors/Updates/Internal/CollectionDiffBuilder.cs
```diff
@@ -164,36 +167,37 @@ public void GetDictionaryChanges(
         newItemsToProcess = null;
         removedKeys = null;
 
-        _oldKeys.Clear();
+        // Track removed keys - start with all old subject-valued keys, remove ones that still exist
+        // with the same value below.
+        HashSet<object>? keysToRemove = null;
```
RicoSuter, line 172: Use _oldKeys so it's pooled like in master?

src/Namotion.Interceptor.Connectors/Updates/Internal/CollectionDiffBuilder.cs
```diff
@@ -224,6 +228,12 @@ public int GetNewIndex(IInterceptorSubject item) =>
     /// </summary>
     public IReadOnlyList<IInterceptorSubject> GetCommonItems() => _newCommonOrder;
 
+    /// <summary>
+    /// Gets dictionary entries whose key and value (by reference) exist in both old and new.
+    /// Populated by <see cref="GetDictionaryChanges"/>.
+    /// </summary>
+    public IReadOnlyList<(object key, IInterceptorSubject item)> GetCommonDictionaryItems() => _commonDictionaryItems;
```
RicoSuter: why is this called common? and do we need to expose it publicly?

src/Namotion.Interceptor.Connectors/Updates/Internal/SubjectValueConvert.cs
```diff
@@ -0,0 +1,86 @@
...
+    /// Returns the value as an <see cref="IDictionary"/> for keyed subject access.
+    /// Fast path: the value is already an <see cref="IDictionary"/> and is returned as-is (zero allocation).
+    /// Slow path: read-only types that implement only <see cref="IEnumerable{T}"/> of <c>KeyValuePair</c>
+    /// are materialized into a fresh <see cref="Dictionary{TKey,TValue}"/>. Callers must filter values
+    /// against <see cref="IInterceptorSubject"/> inline since the returned dictionary may contain
+    /// non-subject entries on the passthrough path.
+    /// </summary>
+    internal static IDictionary ToSubjectDictionary(object value)
+    {
+        if (value is IDictionary dict)
```
RicoSuter: rename dict to dictionary

src/Namotion.Interceptor.Tracking/SubjectValueVisitor.cs
RicoSuter: Already renamed to `SubjectLookup` in an earlier commit on this branch.

src/Namotion.Interceptor.Connectors/Updates/Internal/CollectionDiffBuilder.cs
RicoSuter: Leaving as-is. The local HashSet is lazily allocated (only when old dict has subject entries) and short-lived. The master code actually allocated more (always populated _oldKeys, then copied via [\.._oldKeys]). Marginal benefit from pooling given the builder itself is already pooled.

src/Namotion.Interceptor.Connectors/Updates/Internal/CollectionDiffBuilder.cs
RicoSuter: Renamed both to `GetRetainedItems()` and `GetRetainedDictionaryItems()` for clarity.

RicoSuter:
## LC=10 benchmark results vs `master`

Full sweep at `LaunchCount=10` across all 23 enabled benchmark methods, master vs branch on the same host (i7-11700K, .NET 9.0.11). Δ% computed as `(branch - master) / master * 100`. Both Mean and Median Δ% reported; Median is more robust against per-iteration outliers (GC pauses, scheduler hiccups).

**Allocations: 0% Δ across every benchmark.** No GC pressure change anywhere.

### Improvements (Median Δ ≤ -1%)

| Class | Method | Mean Δ% | **Median Δ%** |
|---|---|---:|---:|
| SourcePathProviderBenchmark | TryGetPropertyFromSegment | -14.51% | **-10.74%** |
| SourcePathProviderBenchmark | VisitPropertiesFromPaths | -3.55% | **-4.09%** |
| SubjectUpdateBenchmark | CreateCompleteUpdate | -1.02% | **-1.21%** |
| RegistryBenchmark | DerivedAverage | -1.01% | **-1.15%** |

### Regressions (Median Δ ≥ +1%)

| Class | Method | Mean Δ% | **Median Δ%** | Notes |
|---|---|---:|---:|---|
| SourcePathProviderBenchmark | TryGetPath | +6.36% | **+10.01%** | Cold path: only Mcp tool resolution + MQTT initial-state snapshot. ~30 ns absolute. |
| RegistryBenchmark | WriteNoOp | +2.21% | **+3.85%** | High inter-run variance (was -2.23% at LC=7); structural codegen. |
| ServiceOrderResolverBenchmark | MixedDependencies | +1.79% | **+2.04%** | **Untouched code.** Pure structural codegen variance. |
| SubjectUpdateBenchmark | CreatePartialUpdate | +2.47% | **+1.32%** | Borderline. Hot-path code is bit-identical to master. |

### Within noise (`|Median Δ%| < 1%`): 15 methods

`ChangeAllTires` +0.98%, `Read` +0.94%, `WriteWithTimestampScope` +0.91% (Mean said -2.41%, Median says noise), `WithFirstLastGroups` -0.78% (untouched), `IncrementDerivedAverage` -0.65%, `GetOrAddSubjectId` +0.48%, `AddLotsOfPreviousCars` +0.46%, `LinearChain` +0.40% (untouched), `NoDependencies` -0.39% (untouched), `WriteToSource` +0.21%, `KnownSubjectsSnapshot` +0.16%, `GenerateSubjectId` -0.10%, `Write` +0.06%, `WriteToRegistrySubjects` -0.03% (Mean said -2.60%, Median says noise), `TryGetPropertyFromPath` +0.01%.

### Hot-path audit for the two real regressions

For both `TryGetPath` and `CreatePartialUpdate`, every line of source code executed by the benchmark is **bit-identical to master**:

- `TryGetPath` walks `current.Parent.Parents[0]` via accessors that did not change. The `PathExtensions.cs` change is to `TryGetSubjectFromValue` (resolution down), not `TryGetPath` (build up); the benchmark hits the up direction only.
- `CreatePartialUpdate` processes 4 scalar changes (2 doubles, 2 strings). Only the scalar branch of `ApplyPropertyChangeToUpdate` runs, which is bit-identical to master. The dictionary/collection branches changed (typed `IDictionary` / `IEnumerable<IInterceptorSubject>` -> `object?`) are not executed.

The remaining timing differences are therefore **structural code-layout effects**: methods in the same assembly grew (`SubjectPropertyTypeExtensions`, `RegisteredSubjectProperty`, `SubjectUpdateFactory`, new `SubjectLookup` and `SubjectValueConvert` files), shifting the JIT-compiled code-heap layout and changing i-cache / branch-predictor behaviour on unchanged hot-path methods. Not an algorithmic regression.

### Net verdict

Branch is net positive on real workloads:

- 4 genuine algorithmic improvements (one at -10.7% on `TryGetPropertyFromSegment`)
- Zero allocation changes
- One real regression on a cold setup path (`TryGetPath`), absolute slowdown ~30 ns and called O(N) per MQTT client connection or per Mcp tool invocation, invisible vs network/IO costs
- One borderline regression on `CreatePartialUpdate` at +1.32% Median, well inside noise floor
- One structural-noise regression on `MixedDependencies` on code that did not change

Commit `8dc9fba0` (extract slow paths in `SubjectLookup` and `SubjectValueConvert` for JIT inlining) eliminated three prior LC=7 regression candidates and reduced `TryGetPath` further. The remaining residuals are code-layout side effects of unrelated assembly growth, not reachable by source-level changes.

<details>
<summary>Raw BenchmarkDotNet output (click to expand)</summary>

#### Branch (current)

```
.NET SDK 10.0.100
  [Host]     : .NET 9.0.11 (9.0.11, 9.0.1125.51716), X64 RyuJIT x86-64-v4
  Job-CKQBHW : .NET 9.0.11 (9.0.11, 9.0.1125.51716), X64 RyuJIT x86-64-v4

LaunchCount=10
```

| Type                          | Method                    | Type        | Mean             | Error          | StdDev         | Median           | P95              | Ratio    | RatioSD | Gen0      | Gen1      | Gen2     | Allocated  | Alloc Ratio |
|------------------------------ |-------------------------- |------------ |-----------------:|---------------:|---------------:|-----------------:|-----------------:|---------:|--------:|----------:|----------:|---------:|-----------:|------------:|
| **ServiceOrderResolverBenchmark** | **NoDependencies**            | **?**           |      **1,232.99 ns** |       **6.304 ns** |      **22.352 ns** |      **1,230.30 ns** |      **1,292.40 ns** |     **1.00** |    **0.03** |    **0.1945** |         **-** |        **-** |     **1632 B** |        **1.00** |
| SourcePathProviderBenchmark   | TryGetPropertyFromPath    | ?           |        358.10 ns |       0.576 ns |       2.027 ns |        358.30 ns |        360.94 ns |     0.29 |    0.01 |    0.0753 |         - |        - |      632 B |        0.39 |
| SubjectSourceBenchmark        | WriteToRegistrySubjects   | ?           |  2,911,419.05 ns |   2,277.468 ns |   8,075.051 ns |  2,910,314.92 ns |  2,930,460.90 ns | 2,362.02 |   41.84 |         - |         - |        - |          - |        0.00 |
| SubjectUpdateBenchmark        | CreateCompleteUpdate      | ?           |      7,774.60 ns |      14.858 ns |      52.488 ns |      7,771.63 ns |      7,864.02 ns |     6.31 |    0.12 |    0.7019 |    0.0153 |        - |     5904 B |        3.62 |
| ServiceOrderResolverBenchmark | LinearChain               | ?           |      1,347.18 ns |       7.716 ns |      27.656 ns |      1,341.69 ns |      1,394.31 ns |     1.09 |    0.03 |    0.3147 |         - |        - |     2640 B |        1.62 |
| SourcePathProviderBenchmark   | VisitPropertiesFromPaths  | ?           |        715.01 ns |       0.952 ns |       3.362 ns |        715.84 ns |        719.89 ns |     0.58 |    0.01 |    0.1526 |         - |        - |     1280 B |        0.78 |
| SubjectSourceBenchmark        | WriteToSource             | ?           |  1,981,756.35 ns |  15,051.757 ns |  54,913.466 ns |  1,990,135.74 ns |  2,067,242.46 ns | 1,607.79 |   52.57 |         - |         - |        - |          - |        0.00 |
| SubjectUpdateBenchmark        | CreatePartialUpdate       | ?           |      2,710.56 ns |      28.954 ns |     101.142 ns |      2,678.57 ns |      3,020.66 ns |     2.20 |    0.09 |    0.3166 |         - |        - |     2656 B |        1.63 |
| ServiceOrderResolverBenchmark | MixedDependencies         | ?           |      1,383.57 ns |       3.879 ns |      13.704 ns |      1,381.71 ns |      1,407.06 ns |     1.12 |    0.02 |    0.2460 |         - |        - |     2064 B |        1.26 |
| SourcePathProviderBenchmark   | TryGetPropertyFromSegment | ?           |         42.17 ns |       0.286 ns |       1.022 ns |         42.16 ns |         43.66 ns |     0.03 |    0.00 |         - |         - |        - |          - |        0.00 |
| ServiceOrderResolverBenchmark | WithFirstLastGroups       | ?           |      1,669.15 ns |       5.171 ns |      18.064 ns |      1,673.39 ns |      1,692.85 ns |     1.35 |    0.03 |    0.3071 |         - |        - |     2576 B |        1.58 |
| SourcePathProviderBenchmark   | TryGetPath                | ?           |        305.65 ns |       3.579 ns |      12.781 ns |        313.72 ns |        318.74 ns |     0.25 |    0.01 |    0.0315 |         - |        - |      264 B |        0.16 |
|                               |                           |             |                  |                |                |                  |                  |          |         |           |           |          |            |             |
| **RegistryBenchmark**             | **AddLotsOfPreviousCars**     | **interceptor** | **67,998,234.79 ns** | **120,072.326 ns** | **431,940.288 ns** | **67,995,843.12 ns** | **68,714,840.91 ns** |        **?** |       **?** | **3375.0000** | **2250.0000** | **625.0000** | **23856667 B** |           **?** |
| RegistryBenchmark             | IncrementDerivedAverage   | interceptor |      7,094.81 ns |      16.037 ns |      56.862 ns |      7,079.10 ns |      7,226.56 ns |        ? |       ? |    0.0153 |         - |        - |      128 B |           ? |
| RegistryBenchmark             | WriteNoOp                 | interceptor |        496.01 ns |       5.968 ns |      21.393 ns |        493.41 ns |        540.66 ns |        ? |       ? |         - |         - |        - |          - |           ? |
| RegistryBenchmark             | Write                     | interceptor |      1,368.86 ns |      12.276 ns |      43.365 ns |      1,359.85 ns |      1,447.19 ns |        ? |       ? |         - |         - |        - |          - |           ? |
| RegistryBenchmark             | WriteWithTimestampScope   | interceptor |      1,246.24 ns |       7.516 ns |      26.056 ns |      1,243.34 ns |      1,299.79 ns |        ? |       ? |         - |         - |        - |          - |           ? |
| RegistryBenchmark             | Read                      | interceptor |        594.74 ns |       2.454 ns |       8.700 ns |        593.29 ns |        615.59 ns |        ? |       ? |         - |         - |        - |          - |           ? |
| RegistryBenchmark             | DerivedAverage            | interceptor |        385.30 ns |       2.104 ns |       7.516 ns |        383.22 ns |        400.89 ns |        ? |       ? |         - |         - |        - |          - |           ? |
| RegistryBenchmark             | ChangeAllTires            | interceptor |     17,372.05 ns |      35.726 ns |     128.060 ns |     17,387.99 ns |     17,577.06 ns |        ? |       ? |    2.0447 |    0.1221 |        - |    17216 B |           ? |
| RegistryBenchmark             | GetOrAddSubjectId         | interceptor |         27.42 ns |       0.049 ns |       0.172 ns |         27.37 ns |         27.67 ns |        ? |       ? |         - |         - |        - |          - |           ? |
| RegistryBenchmark             | GenerateSubjectId         | interceptor |      1,022.34 ns |       0.753 ns |       2.581 ns |      1,021.47 ns |      1,027.41 ns |        ? |       ? |    0.0076 |         - |        - |       72 B |           ? |
| RegistryBenchmark             | KnownSubjectsSnapshot     | interceptor |  1,419,487.94 ns |   2,632.779 ns |   9,091.719 ns |  1,422,387.12 ns |  1,431,233.37 ns |        ? |       ? |   37.1094 |   11.7188 |        - |   320416 B |           ? |

#### master

```
11th Gen Intel Core i7-11700K 3.60GHz (Max: 2.43GHz), 1 CPU, 16 logical and 8 physical cores
.NET SDK 10.0.100
  [Host]     : .NET 9.0.11 (9.0.11, 9.0.1125.51716), X64 RyuJIT x86-64-v4
  Job-CKQBHW : .NET 9.0.11 (9.0.11, 9.0.1125.51716), X64 RyuJIT x86-64-v4

LaunchCount=10
```

| Type                          | Method                    | Type        | Mean             | Error          | StdDev         | Median           | P95              | Ratio    | RatioSD | Gen0      | Gen1      | Gen2     | Allocated  | Alloc Ratio |
|------------------------------ |-------------------------- |------------ |-----------------:|---------------:|---------------:|-----------------:|-----------------:|---------:|--------:|----------:|----------:|---------:|-----------:|------------:|
| **ServiceOrderResolverBenchmark** | **NoDependencies**            | **?**           |      **1,236.32 ns** |       **5.953 ns** |      **21.417 ns** |      **1,235.14 ns** |      **1,293.61 ns** |     **1.00** |    **0.02** |    **0.1945** |         **-** |        **-** |     **1632 B** |        **1.00** |
| SourcePathProviderBenchmark   | TryGetPropertyFromPath    | ?           |        358.28 ns |       0.710 ns |       2.508 ns |        358.26 ns |        362.44 ns |     0.29 |    0.01 |    0.0753 |         - |        - |      632 B |        0.39 |
| SubjectSourceBenchmark        | WriteToRegistrySubjects   | ?           |  2,989,119.49 ns |  70,341.904 ns | 246,643.087 ns |  2,911,258.88 ns |  3,752,794.38 ns | 2,418.45 |  202.98 |         - |         - |        - |          - |        0.00 |
| SubjectUpdateBenchmark        | CreateCompleteUpdate      | ?           |      7,854.50 ns |      20.117 ns |      72.369 ns |      7,866.64 ns |      7,938.84 ns |     6.35 |    0.12 |    0.7019 |    0.0153 |        - |     5904 B |        3.62 |
| ServiceOrderResolverBenchmark | LinearChain               | ?           |      1,335.40 ns |       5.054 ns |      17.788 ns |      1,336.32 ns |      1,359.82 ns |     1.08 |    0.02 |    0.3147 |         - |        - |     2640 B |        1.62 |
| SourcePathProviderBenchmark   | VisitPropertiesFromPaths  | ?           |        741.34 ns |       7.090 ns |      24.483 ns |        746.39 ns |        795.77 ns |     0.60 |    0.02 |    0.1526 |         - |        - |     1280 B |        0.78 |
| SubjectSourceBenchmark        | WriteToSource             | ?           |  1,971,110.93 ns |  22,496.457 ns |  81,502.670 ns |  1,986,021.69 ns |  2,122,159.05 ns | 1,594.80 |   70.98 |         - |         - |        - |          - |        0.00 |
| SubjectUpdateBenchmark        | CreatePartialUpdate       | ?           |      2,645.21 ns |       5.329 ns |      18.684 ns |      2,643.80 ns |      2,675.90 ns |     2.14 |    0.04 |    0.3166 |         - |        - |     2656 B |        1.63 |
| ServiceOrderResolverBenchmark | MixedDependencies         | ?           |      1,359.18 ns |       5.781 ns |      20.040 ns |      1,354.02 ns |      1,393.69 ns |     1.10 |    0.02 |    0.2460 |         - |        - |     2064 B |        1.26 |
| SourcePathProviderBenchmark   | TryGetPropertyFromSegment | ?           |         49.33 ns |       1.483 ns |       5.298 ns |         47.23 ns |         63.25 ns |     0.04 |    0.00 |         - |         - |        - |          - |        0.00 |
| ServiceOrderResolverBenchmark | WithFirstLastGroups       | ?           |      1,677.03 ns |       5.999 ns |      21.425 ns |      1,686.49 ns |      1,698.44 ns |     1.36 |    0.03 |    0.3071 |         - |        - |     2576 B |        1.58 |
| SourcePathProviderBenchmark   | TryGetPath                | ?           |        287.36 ns |       1.431 ns |       4.998 ns |        285.18 ns |        299.00 ns |     0.23 |    0.01 |    0.0315 |         - |        - |      264 B |        0.16 |
|                               |                           |             |                  |                |                |                  |                  |          |         |           |           |          |            |             |
| **RegistryBenchmark**             | **AddLotsOfPreviousCars**     | **interceptor** | **67,756,763.59 ns** | **123,276.906 ns** | **432,251.546 ns** | **67,683,452.25 ns** | **68,360,485.72 ns** |        **?** |       **?** | **3375.0000** | **2250.0000** | **625.0000** | **23856668 B** |           **?** |
| RegistryBenchmark             | IncrementDerivedAverage   | interceptor |      7,118.31 ns |      10.478 ns |      37.286 ns |      7,124.97 ns |      7,176.37 ns |        ? |       ? |    0.0153 |         - |        - |      128 B |           ? |
| RegistryBenchmark             | WriteNoOp                 | interceptor |        485.30 ns |       4.940 ns |      17.770 ns |        475.11 ns |        522.50 ns |        ? |       ? |         - |         - |        - |          - |           ? |
| RegistryBenchmark             | Write                     | interceptor |      1,361.53 ns |       7.609 ns |      27.373 ns |      1,359.07 ns |      1,411.63 ns |        ? |       ? |         - |         - |        - |          - |           ? |
| RegistryBenchmark             | WriteWithTimestampScope   | interceptor |      1,276.99 ns |      18.885 ns |      65.967 ns |      1,232.16 ns |      1,390.57 ns |        ? |       ? |         - |         - |        - |          - |           ? |
| RegistryBenchmark             | Read                      | interceptor |        590.59 ns |       3.249 ns |      11.479 ns |        587.77 ns |        607.60 ns |        ? |       ? |         - |         - |        - |          - |           ? |
| RegistryBenchmark             | DerivedAverage            | interceptor |        389.25 ns |       2.156 ns |       7.617 ns |        387.69 ns |        405.02 ns |        ? |       ? |         - |         - |        - |          - |           ? |
| RegistryBenchmark             | ChangeAllTires            | interceptor |     17,227.48 ns |      37.043 ns |     132.301 ns |     17,219.83 ns |     17,447.16 ns |        ? |       ? |    2.0447 |    0.1221 |        - |    17216 B |           ? |
| RegistryBenchmark             | GetOrAddSubjectId         | interceptor |         27.28 ns |       0.023 ns |       0.081 ns |         27.24 ns |         27.43 ns |        ? |       ? |         - |         - |        - |          - |           ? |
| RegistryBenchmark             | GenerateSubjectId         | interceptor |      1,022.71 ns |       0.270 ns |       0.932 ns |      1,022.46 ns |      1,024.46 ns |        ? |       ? |    0.0076 |         - |        - |       72 B |           ? |
| RegistryBenchmark             | KnownSubjectsSnapshot     | interceptor |  1,423,102.78 ns |   6,899.202 ns |  24,099.982 ns |  1,420,136.41 ns |  1,470,402.57 ns |        ? |       ? |   37.1094 |   11.7188 |        - |   320416 B |           ? |

Issue #328, opened 2026-05-31
Fluent mapping: add instance-level (ForPath) overrides


RicoSuter:
Follow-up to #325, which makes fluent mapping a complete code-based alternative to attributes at the type level (`ForType<T>().Map(...)` plus `.Configure(...)`). #325 ships type-level only on purpose, since that fully replaces attributes (attributes are themselves type-level only). This issue tracks adding instance-level overrides on top.

## What ForPath adds

Instance-specific configuration keyed by an absolute path from the connected root, for tuning a single location differently from the type-level default:

```csharp
fluent.ForPath(p => p.Motors[2].Speed,         b => b.SamplingInterval(250).QueueSize(10));
fluent.ForPath(p => p.Sensors["boiler"].Value, b => b.DeadbandType(DeadbandType.Absolute).DeadbandValue(0.5));
fluent.ForPath(p => p.Pump.Motor.Speed,        b => b.WithRetain(true)); // MQTT
```

## Design constraints (decided in #325)

- **Metadata-only.** ForPath sets delivery and monitoring fields (MQTT QoS and Retain; OPC UA sampling, queue, deadband), never the path or topic segment. Segments stay a type-level concern. This keeps the segment identical forward and reverse, so reverse lookup stays entirely in the existing path-provider mapper (the fluent metadata mapper returns null on reverse). Allowing an instance segment or topic override would require a separate instance-keyed reverse index, reintroducing the O(n) scan that #325 removes.
- **Leaf-only.** The selector must resolve to a value leaf; a subject, subject-collection, or subject-dictionary target throws at registration.

## Additive implementation (no changes to the #325 type-level design)

1. Add an `instancePath -> metadata` map to `FluentMappingRegistry`.
2. Prepend an instance-first check in `FluentMetadataMapper.TryGetMapping` (instance metadata merged over type-level, most-specific-wins).
3. Add a full-path-with-indexers entry point to `ExpressionPathHelper` (member chains plus `ArrayIndex` and `get_Item` for arrays, `List`, and `Dictionary`, evaluating the constant or captured index or key, emitting canonical bracket segments). Depends on the index- and `[InlinePaths]`-aware `PathExtensions.TryGetPath(string separator)` overload (done in #325) so instance lookup paths match.
4. Add `ForPath(...)` to the fluent builder.

Reverse lookup, the path provider, the composite wiring, and `ForType` are all unchanged.

## Out of scope (candidate for a later extension)

[...]

PR #329, opened 2026-05-31
Add code-based fluent mapping as a type-level alternative to attributes (#325)

### LINKED ISSUE #325: Make the fluent mapper a complete code-based alternative to attribute mapping (per-type, relative, array-aware)
Follow-up from #323 (property-mapper abstraction). Not blocking that PR; capturing the design here to pick up afterwards.

## Background

Connectors resolve property mappings from two sources today:

- **Attributes** (`[MqttTopic]` / `[Path]` / `[OpcUaNode]`) read by `AttributeBasedPathProvider` + the per-connector attribute mapper. These are keyed by `(declaringType, member)`, so a segment is defined **once on the type** and composed hierarchically wherever that type appears, including collection and dictionary elements (the structural path layer appends the index, e.g. `motors[2]/speed`). Reverse lookup works because `PathProviderBase.TryGetPropertyFromSegment` is segment-driven.
- **Fluent** (`MqttFluentMapper<T>`, `OpcUaFluentMapper<T>`) keyed by the **absolute instance path from the connected root** (`Pump.Motor.Speed`, `Fan.Motor.Speed`, ...).

## Limitation

Because the fluent mappers key by absolute instance path, they cannot express two things that attributes handle naturally:

1. **Reuse across locations.** A type used in several places (e.g. a `Motor` at `Pump.Motor` and `Fan.Motor`) must have each property redefined per location.
2. **Collection / array elements.** Selectors are member-access chains (`x => x.A.B`); an indexer (`x => x.Motors[0].Speed`) is rejected, and there is no per-element registration. So array/dictionary element properties cannot be configured fluently at all.

This affects **both** connectors. OPC UA's nested `.Map(r => r.Person, person => person.Map(...))` does not change this: it still produces absolute-path keys (`_basePath + "." + relativePath`); its value is configuring intermediate object nodes and per-level type safety, not per-type reuse.

So the fluent mapper is currently only suitable for instance-specific configuration, not as a standalone replacement for attributes.

## Goal

Make fluent a **complete, code-based alternative** to attribute mapping, so an external/runtime configuration can define everything without annotating the subject types (useful for third-party types, per-deployment variation, or keeping the domain model free of connector concerns). Attributes continue to work unchanged; the code-based source is a **parallel option**, and the two coexist through the existing composite mapper.

## Proposed approach

Reuse the existing path engine; only replace the **source of truth**.

- Keep `PathExtensions`, `PathProviderBase`, `*PathProviderMapper`, and the composite untouched. They already do hierarchical composition, index/array handling, `[InlinePaths]`, and reverse navigation.
- Add a `FluentPathProvider : PathProviderBase` that overrides only `IsPropertyIncluded` + `TryGetPropertySegment`, sourcing the relative segment from a `Dictionary<(Type declaringType, string member), segment>`. It is a drop-in replacement for `AttributeBasedPathProvider`, so it inherits forward composition, arrays, `[InlinePaths]`, and reverse lookup for free.
- Add a code-backed metadata mapper keyed by `(declaringType, member)` (drop-in for `MqttAttributeMapper` / `OpcUaAttributeMapper`) supplying protocol metadata: MQTT QoS/Retain; OPC UA BrowseName/NodeId/type definition/references/sampling/etc. The fluent `PropertyBuilder` already models these fields.
- The fluent builder becomes a way to **populate those registries** via single-member selectors scoped to a type, e.g. `ForType<Motor>().Map(m => m.Speed, s => s.WithSegment("speed"))`, rather than a standalone matcher keyed by absolute path. The single-member (no chain, no indexer) restriction is exactly right in this model.

The pattern is shared: redesign it once and both connectors get it, each supplying only its protocol-specific metadata builder.

## Open design decisions

- **Scoping and precedence (central question).** Support both type-level registration ("`Motor.Speed` everywhere") and path-level overrides ("the Motor at `Pump.Motor` specifically"), with most-specific-wins (exact instance path beats `(type, member)`). Attributes only do type-level; today's fluent only does path-level; the replacement needs both.
- **Polymorphism / inheritance.** A `(type, member)` lookup should walk base types and interfaces so a derived type (e.g. `ServoMotor : Motor`) inherits the base registration unless overridden.
- **Array index segment format.** `motors[2]/speed` (current bracket form) vs `motors/2/speed`. Decide and keep forward and reverse consistent.
- **Coexistence / migration.** Attributes and code-based sources merge via the composite, so adoption can be incremental; using the code-based source exclusively is opt-in.
- **Breaking API change.** The fluent API moves from absolute-path selectors to type-scoped single-member selectors. The current `Map(x => x.Pump.Motor.Speed, ...)` style would be replaced.

## Notes

- Reverse lookup likely gets faster: segment-guided navigation replaces the current fluent O(n) full-subtree scan per lookup.
- This validates the existing `ExpressionPathHelper` single-member restriction rather than relaxing it.

## Acceptance / parity checklist

- [ ] A type used in multiple locations and in collections/dictionaries is configured once in code and resolves everywhere (forward and reverse).
- [ ] Collection/array element properties resolve in both directions.
- [ ] Type-level plus path-level override precedence implemented and tested.
- [ ] Polymorphic (base type) registrations apply to derived types.
- [ ] Attributes continue to work and can be mixed with the code-based source via the composite.
- [ ] Shared implementation across MQTT and OPC UA (no per-connector duplication of the keying/composition logic).

RicoSuter:
## Summary

Makes code-based fluent configuration a complete type-level alternative to attribute mapping for the MQTT and OPC UA connectors, so external or runtime configuration can map an object graph without annotating the subject types. Closes #325.

Fluent becomes a second mapping source keyed by `(declaringType, member)`, exactly like attributes, reusing the same path engine and composite machinery instead of a separate engine. The old absolute-path fluent mappers are removed.

### Shared core (in Connectors, reusing the Registry path engine)
- `FluentMappingRegistry<TMetadata>`, an inheritance-aware source of truth: it resolves a member by walking the runtime type, its base classes, then its interfaces, most-derived first. It holds per-property segment plus metadata, and class-level (type) metadata.
- `FluentPathProvider<TMetadata> : PathProviderBase`, a drop-in for `AttributeBasedPathProvider` that reads segments from the registry (keeps index handling, `[InlinePaths]` support, and segment-guided reverse lookup).
- `ExpressionPathHelper.GetSingleMemberName` for strict single-member `Map` selectors.

### Fluent builder API

Each connector exposes a `*FluentMapperBuilder<TRoot>` that you configure per type and then `Build(pathSeparator)` into a mapper:

- MQTT: `MqttFluentMapperBuilder<TRoot>` with `ForType<T>().Map(...)` and a `MqttFluentPropertyBuilder` (`WithSegment`, `WithQualityOfService`, `WithRetain`). `Build('/')` returns an `MqttFluentMapper`.
- OPC UA: `OpcUaFluentMapperBuilder<TRoot>` with `ForType<T>().Map(...)` and `.Configure(...)` (class-level node metadata), plus an `OpcUaFluentPropertyBuilder` mirroring the attribute fields (`BrowseName` sets both the path segment and the node browse name). `Build('.')` returns an `OpcUaFluentMapper`.

Each `*FluentMapper` is a single class that subclasses the connector's `*PathProviderMapper` (constructed over a `FluentPathProvider`): a forward override layers the registry metadata on top of the resolved path/browse name (OPC UA also merges the class-level type-self metadata for subject-typed members), and reverse lookup is inherited from the base mapper. You compose the built mapper into the connector's `Mapper`, layered after the attribute mappers so fluent wins on conflicts; default, attribute, and fluent sources coexist.

```csharp
Mapper = new MqttCompositeMapper(
    new MqttPathProviderMapper(new AttributeBasedPathProvider("mqtt", '/')),
    new MqttAttributeMapper("mqtt"),
    new MqttFluentMapperBuilder<Plant>()
        .ForType<Motor>().Map(m => m.Speed, b => b.WithSegment("speed"))
        .Build('/'))
```

### OPC UA configuration ergonomics

The OPC UA connector now configures itself the way MQTT does. `OpcUaServerConfiguration.ValueConverter` and the client config's `ValueConverter` and `SubjectFactory` default instead of being required, and the DI-dependent fields are completed by the connector at registration time: `TelemetryContext` is nullable and, when left null, is filled with DI-backed telemetry (set it explicitly, including to `NullTelemetryContext.Instance`, to keep your own), and the client `TypeResolver` is created from the DI logger when unset. The consumer just writes `new OpcUaServerConfiguration { Mapper = ... }`. The previous `CreateDefault*Configuration` factories are removed.

### Path engine

The structural path computation (parent-chain walk with bracket indices such as `Items[1]` and `[InlinePaths]` support) lives in the path-provider overload `TryGetPath(pathProvider, rootSubject)`, which fluent reuses; a simple `TryGetPath(rootSubject)` convenience uses the default provider. This index-aware computation is also the prerequisite for the deferred instance-level `ForPath` feature.

### Scope
Type-level mapping only. Instance-level `ForPath` overrides are deferred to #328.

## Test plan
- [x] Full solution unit suite passes (`dotnet test src/Namotion.Interceptor.slnx --filter "Category!=Integration"`), zero failures across all projects.
- [x] MQTT connector tests pass, including integration (73/73).
- [x] OPC UA connector tests pass, including integration (259/259 on a clean run), which exercises the config auto-wiring end to end.
- [x] Unit tests cover: inheritance-aware resolution, forward and reverse lookup, type reuse across locations, collection elements with bracket indices (both directions), OPC UA class-level (type-self) merge, the DI auto-wiring (filled when unset, preserved when explicit), and fluent-wins-over-attribute precedence in production mapper order.
- [x] Public API snapshots updated for Registry, Connectors, MQTT, and OPC UA.
- [x] Connector documentation updated for the new fluent API and OPC UA configuration.


docs/connectors-mqtt.md
```diff
@@ -183,7 +183,7 @@ The built-in mapper types:
 |---|---|
 | `MqttPathProviderMapper` | Wraps a `PathProviderBase` to produce topics from `[Path]` attributes |
 | `MqttAttributeMapper` | Layers per-topic QoS and Retain from `[MqttTopic]` attributes onto the mapping (the topic segment itself is resolved by the path provider) |
-| `MqttFluentMapper<TSubject>` | Code-based per-property configuration via lambda expressions |
+| `MqttFluentMapping<TRoot>` | Code-based type-level configuration via `ForType<T>().Map(...)` |
```
RicoSuter: Should not be named mapping, mapping is the object with the information, mapper produces it, no?

docs/connectors-mqtt.md
```diff
@@ -235,18 +235,49 @@ Multi-level topics are built by annotating each level of the object graph (paren
...
+
+#### DI Registration
+
+The `AddMqttSubjectServer<T>` and `AddMqttSubjectClientSource<T>` overloads accept an optional `configureFluent` callback:
+
+```csharp
+services.AddMqttSubjectServer<Plant>(
+    connectorName: "mqtt",
+    brokerPort: 1883,
+    configureFluent: fluent =>
```
RicoSuter: probably should not be here, but just via the config lambda extension method, no?

PR #333, opened 2026-06-03
Drop permanent OPC UA write failures from retry queue

### LINKED ISSUE #332: OPC UA: permanently-failed writes are retried indefinitely
Surfaced while reviewing the OPC UA write path during PR #313 (pre-existing, not introduced by that PR).

## Problem

The transient vs permanent classification of failed writes is computed but does not influence retry behavior, so writes that can never succeed are retried forever.

`ProcessWriteResults` in `OutboundWriter` separates transient from permanent failed writes, but the resulting `WriteResult.PartialFailure` / `WriteResult.Failure` carries all failed changes together. `WriteChangesViaRetryQueueAsync` (`SubjectSourceBase`) then enqueues every failed change into the `WriteRetryQueue` regardless of classification, and the queue is re-applied on every subsequent write cycle.

The transient/permanent split currently only feeds the `OpcUaWriteException` count properties and a warning log line.

## Impact

A property that permanently fails to write (for example `BadNotWritable` or `BadUserAccessDenied`) is retried on every write cycle for the lifetime of the process. For 24/7 operation this is a slow, ring-buffer-bounded waste rather than a storm, but it never converges.

## Suggested fix (either)

- Only re-enqueue transient failures into the retry queue; drop permanently-failed changes (optionally surface them once via diagnostics/logging).
- Or document explicitly that the transient/permanent split is informational only and intended to be retried regardless.

## Relevant code

- `src/Namotion.Interceptor.Connectors/SubjectSourceBase.cs` (`WriteChangesViaRetryQueueAsync`, `ReapplyRetryQueue`)
- `src/Namotion.Interceptor.Connectors/WriteRetryQueue.cs`
- `src/Namotion.Interceptor.OpcUa/Client/OutboundWriter.cs` (`ProcessWriteResults`)
- `src/Namotion.Interceptor.OpcUa/Client/OpcUaStatusCodeClassifier.cs`
- `src/Namotion.Interceptor.OpcUa/Client/OpcUaWriteException.cs`

Note: the retry queue lives in the shared `SubjectSourceBase`, so the same behavior applies to other connectors (for example MQTT).

RicoSuter:
Fixes #332.

## Summary

- `OutboundWriter.ProcessWriteResults` now drops writes that fail with a permanent OPC UA status code (`BadAttributeIdInvalid`, `BadTypeMismatch`, `BadWriteNotSupported`, `BadNodeIdUnknown`, `BadUserAccessDenied`, `BadNotWritable`) from `WriteResult.FailedChanges`, so they never enter the retry queue. Each dropped NodeId is logged once per session at warning level.
- New `OpcUaClientConfiguration.DropPermanentWriteFailures` (default `true`) gates the behavior. Set to `false` to restore the pre-fix behavior (every failed write is retried forever).
- The reported-NodeIds set is cleared on every session change (`OpcUaSubjectClientSource.OnCurrentSessionChanged`), so a new session re-evaluates each NodeId fresh and operators get a new warning line if a previously-broken node is still broken after reconnection. Memory is bounded by `unique-permanently-failing-NodeIds-per-session` rather than per-process-lifetime.
- Throughput counter and read-after-write notifications were moved into `ProcessWriteResults` so they read the local `successCount` directly, instead of computing it from `result.FailedChanges.Length` which no longer reflects every failure.

## Why all six codes (not just three)

An earlier version of this PR only dropped the three schema/type-level codes and kept the three state-dependent codes (`BadNodeIdUnknown`, `BadUserAccessDenied`, `BadNotWritable`) in the retry queue on the basis that the OPC UA spec allows the underlying server state (address space membership, permissions, `AccessLevel`) to change at runtime.

In practice these state-dependent codes are static for the vast majority of deployments, and the cost of keeping them in the retry queue (one extra round-trip per write cycle for the lifetime of the connector) outweighs the rare benefit of silent auto-recovery. When state legitimately does change, the natural recovery is a fresh user-initiated write, which retries automatically. The application is still informed of every failure batch via `OpcUaWriteException`; only the *automatic re-attempt* of dropped writes is suppressed.

The config flag exists for the minority of deployments with dynamic AccessLevels or permissions that need the old behavior.

## Test plan

- [x] `dotnet build src/Namotion.Interceptor.slnx` clean (0 warnings, 0 errors)
- [x] `dotnet test src/Namotion.Interceptor.slnx --filter \"Category!=Integration\"` all green across 26 test projects
- [x] `WriteErrorClassificationTests` covers: `IsPermanentWriteError` for all 6 permanent codes, 6 transient codes, and non-bad codes (Good/Uncertain); `ShouldDropFailedWrite` gated by the config flag in both directions; default value of `DropPermanentWriteFailures` pinned to true
- [x] `VerifyChecksTests.PublicApi` snapshot updated for the new public `DropPermanentWriteFailures` property
- [x] Docs updated: `docs/connectors-opcua-client.md` configuration table and new "Permanent failure dropping" subsection

### DIFF
--- a/docs/connectors-opcua-client.md
+++ b/docs/connectors-opcua-client.md
@@ -518,6 +518,7 @@ For 24/7 production use, the default configuration provides robust resilience:
 | `OperationTimeout` | 30s | Timeout for individual OPC UA operations |
 | `SubscriptionHealthCheckInterval` | 5s | Interval for health checks and post-reconnection state sync |
 | `WriteRetryQueueSize` | 1000 | Updates buffered during disconnection |
+| `DropPermanentWriteFailures` | true | Drop writes that fail with a permanent status code instead of retrying every cycle (see [Write Error Handling](#write-error-handling)) |
 | `SessionDisposalTimeout` | 5s | Max wait for graceful session close |
 | `SubscriptionSequentialPublishing` | false | Process subscription messages in order (see Thread Safety) |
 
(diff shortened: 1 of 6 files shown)


RicoSuter:
Superseded by #448, which keeps this PR's goal and its status classification but changes the mechanism.

The reason for not rebasing this one: it removes permanently failed changes from `WriteResult.FailedChanges`. That set carries a completeness contract, and `SourceTransactionWriter` computes `written = sourceChanges \ failedSet`. A change removed from `FailedChanges` is therefore treated as having reached the source, so its rollback is skipped and the model silently keeps a value the source rejected.

That was very likely sound when this was written in June. The transaction work landed afterwards and invalidated it. #448 pins the problem with a test that injects this mechanism and asserts the refused change is then reverted at a source it never reached.

Two smaller reasons: this branch also moves `NotifyPropertiesWritten` to fire for every change, which is the opposite of the fix in #445 that schedules read-backs only for successful writes; and its six files are almost exactly the ones #445 rewrote, so a rebase would have been a rewrite.

What #448 does instead: `FailedChanges` stays complete and an additive `RefusedUntilReconnect` subset is added. The retry queue retains those changes but excludes them from flush attempts, and a flush whose only failures are refused changes reports success, which is what actually removes the stall. Nothing is dropped, so a reconnect retries them naturally, which matters because the note in this PR about a new session having different permissions is correct.

The classification here is reused directly, extended with `BadNodeIdInvalid` and `BadIndexRangeInvalid`, both decided by the request itself.

PR #317, opened 2026-05-20
HomeBlaze: Add history support [PLAN]

RicoSuter:
## Summary

Adds two planning documents for the upcoming time-series history feature in HomeBlaze.

- **`history-mvp.md`** — Design doc. Three packages (`HomeBlaze.History.Abstractions`, `HomeBlaze.History.InMemory`, `HomeBlaze.History.TimescaleDb`) with per-property column dispatch (`value_long` / `value_double` / `value_json`), per-bucket cross-store merge, `get_property_history` MCP tool, and a property history dialog using `MudTimeSeriesChart`. Companion to the architectural target at `architecture/design/history.md`; narrows it to a first shippable slice.
- **`history-mvp-impl-plan.md`** — Implementation plan, 43 TDD-style tasks across 9 phases with exact file paths, code skeletons, and verification steps. Phase 0 promotes `ThroughputCounter` to `Namotion.Interceptor.Connectors` so OPC UA and the history stores share one rate counter.

No code changes in this PR; planning only. Implementation will follow on a separate branch.

## Test plan

- [ ] Docs render correctly in the architecture site (markdown is well-formed).
- [ ] Cross-references to `architecture/design/history.md` resolve.
- [ ] Plan covers all design decisions; review for omitted scope.

PR #249, opened 2026-04-01
HomeBlaze: Implement history system [PLAN]

RicoSuter:
## Summary

Design and implementation plan for the HomeBlaze time-series history system.

### Architecture
- **`HistoryService`** — DI singleton BackgroundService that captures property changes via its own `ChangeQueueProcessor`, buffers them, and distributes to sink subjects discovered via lifecycle events
- **Two-stage buffering** — CQP deduplication (~1s) + per-sink flush intervals via cursor-based shared buffer (in-memory ~1s, SQLite ~10s)
- **Sinks are subjects** in the graph — `InMemoryHistorySink` for fast recent lookups/testing, `SqliteHistorySink` for long-term storage with partitioned databases

### Key Features
- **Move tracking** — detects path changes at runtime, records moves, supports query-time path chain resolution (`FollowMoves`)
- **Aggregation** — separate `QueryAsync` (raw) and `QueryAggregatedAsync` (bucketed) APIs with 7 aggregation types, native SQL in SQLite, cross-partition merging
- **Snapshot reconstruction** — periodic graph snapshots + change replay, backwards partition scan for O(1) lookup
- **Structural recording** — `[State]` properties holding subjects recorded as lightweight path references (not full serialized content)
- **MCP tools** — `get_property_history`, `get_snapshot`, `get_snapshots` for AI agent access
- **Self-recording prevention** — sink subjects filtered from recording to avoid feedback loops

### Documents
- `docs/plans/2026-04-01-history-design.md` — detailed design decisions (implementation reference)
- `docs/plans/2026-04-01-history-implementation.md` — step-by-step implementation plan (12 tasks)
- `src/HomeBlaze/HomeBlaze/Data/Docs/architecture/design/history.md` — permanent architecture document with diagrams and data flows

PR #336, opened 2026-06-06
feat: Add time-series history to HomeBlaze

RicoSuter:
Adds time-series history to HomeBlaze: eligible `[State]` properties are recorded by independent history stores, queried through a stateless cross-store merger, and surfaced in a property history dialog and a `get_property_history` MCP tool. Two stores ship, In-Memory (priority 100, full-resolution live edge) and SQLite (priority 50, durable local history); a TimescaleDB tier is designed but not built.

Two supporting changes ride along because history depends on them. Every user-facing timestamp now renders in a viewer-selected display timezone rather than the server's, which is what makes a history chart readable at all on a remote server. And the Philips Hue integration was reworked so its derived state is observable through interception, since state that never raises a change cannot be recorded.

It also fixes the E2E fixture, which opened a browser context per test and closed none of them until the suite ended. Each open page holds a live Blazor circuit that the host re-renders on every state change, and the resulting pile of idle circuits was making the last E2E test miss its clicks on CI.

Supersedes #317 and #249, which were the two planning branches for this work.

## Why

HomeBlaze could show the current value of a property but nothing about how it got there. Charts, historical queries, and AI questions about the past all need recorded samples, and a home installation and an industrial one want different storage.

The design avoids a central history coordinator. Each store is an `[InterceptorSubject]` `BackgroundService` that records and serves its own data and publishes its own coverage, matching how connectors are already configured, and a stateless merger stitches results by priority and coverage. The store engines are graph-free and operate on canonical paths and typed values, so storage logic stays testable without a HomeBlaze graph and can be extracted into a generic library later.

Recording correctness rests on coverage rather than on the presence of samples. A coverage range asserts that the store was collecting its configured stream throughout that interval and detected no loss, so absence of a sample means the property did not change. That distinction is what lets the chart draw "no data" differently from "no value", and it is why restarts, bounded-queue overflows, and retention all have explicit effects on coverage.

## Contract

For anyone implementing an `IHistoryStore`:

- `CoverageRanges` is an immutable, ordered, non-overlapping snapshot of half-open `[From, To)` intervals. A range asserts the store was collecting throughout it and detected no loss; it is store-wide and all-or-nothing, never per property.
- A store must install its change subscription before beginning a coverage session, so no change can fall inside claimed coverage without reaching the engine.
- A store that drops a sample must end its current range at the first dropped change and start a new one only after it is healthy again. Coverage must never extend past data that is not durable.
- Queries require a non-empty path and aggregation, `From < To`, a positive bucket when given, and a positive `MaxPoints`. Stores return the newest `MaxPoints` points in ascending timestamp order and set `Truncated` when older results were omitted.
- Query errors propagate. A failed store must not be reported as an empty store.

For callers of the merger:

- A bucket is served by exactly one store, never split, so aggregates are never combined across stores.
- Uncovered buckets are explicit null points rather than omitted ones, and carried state (`Last`, `TimeWeightedAverage`) resets across an uncovered interval instead of interpolating through it.
- Numeric aggregations on JSON-backed properties throw `HistoryAggregationNotSupportedException`.

## Breaking changes

All of these are confined to `src/HomeBlaze`, which is not packaged. They matter to plugin authors compiling against these assemblies, not to NuGet consumers of the Namotion.Interceptor libraries, which this branch does not touch.

- API: `SubjectPathResolverExtensions.WithPathResolver()` is removed, and `SubjectPathResolver` now takes `Func<IInterceptorSubject?>` instead of `RootManager`. `RootManager` needs the resolver registered in the context before it loads the graph, so taking `RootManager` in the constructor was a cycle. The resolver is registered by the standard HomeBlaze service registration; callers that invoked `WithPathResolver()` should drop the call.
- API: `McpBuilderExtensions.WithHomeBlazeMcpTools` gained a trailing `params Func<McpToolProviderContext, IMcpToolProvider>[]`, which is how the history package contributes its tool without `HomeBlaze.AI` referencing the history stack. Source compatible; a precompiled caller must be recompiled.
- Behavior: Hue operations now throw on an API error instead of returning silently, and no longer write the expected value into the local resource before the bridge confirms it. State arrives on the next poll. A caller that relied on the operation being silent on failure, or on the value being readable immediately after the call returns, has to change.
- Behavior: the shipped sample Hue bridge configuration polls every 60 seconds instead of every 500 ms. An existing `Hue.json` is not modified; only the shipped default changed.

`StateUnit.Byte` was added to the enum, which is additive.

The five new `VerifyChecksTests.PublicApi.verified.txt` snapshots are all new files for the new history packages. No existing snapshot changed.

## Diff composition

|---|---:|---:|---:|---:|
| Production code | 70 | 8'554 | 134 | +8'420 |
| Tests and benchmarks | 74 | 9'102 | 46 | +9'056 |
| Documentation | 6 | 440 | 78 | +362 |
| Project and CI files | 12 | 193 | 1 | +192 |
| **Total** | **162** | **18'289** | **259** | **+18'030** |

## Verification

- [x] Documentation updated
- [x] Unit tests
- [x] Integration tests
- [ ] ~~Benchmarks~~ nothing in the benchmark project reaches this change; no core library code is touched. Outside `src/HomeBlaze` the branch only adds the new projects to `Namotion.Interceptor.slnx` and a `.codex/config.toml` pointing Codex at the local HomeBlaze MCP endpoint through an exported dev certificate instead of disabling TLS verification
- [ ] ~~Load tests~~ the Connector Tester exercises the connector stack, which this change does not modify
- [ ] ~~Chaos tests~~ same reason

Documentation: `history.md` rewritten for the implemented v1 design, `localization.md` added for the display timezone, and `ai.md`, `overview.md`, `state.md` and `versioning.md` updated where they referenced the changed behavior.

Unit tests: `dotnet test src/Namotion.Interceptor.slnx --filter "Category!=Integration"`, 3'770 passed, 0 failed, across 34 test assemblies.

Integration tests: `dotnet test src/HomeBlaze/HomeBlaze.E2E.Tests`, 26 passed, 0 failed, including the Playwright coverage of the property history dialog. `Validation_NameErrors_PreventAdvancing` had been failing on roughly a third of CI runs while never reproducing locally; that is the fixture leak described above, and it is fixed here rather than left for a follow-up.

Not run: the Connector Tester load and chaos profiles, for the reason above.


src/HomeBlaze/HomeBlaze.History.Sqlite/SqliteHistoryStoreCore.cs
```diff
@@ -0,0 +1,1198 @@
...
+namespace HomeBlaze.History.Sqlite;
+
+/// <summary>
+/// The graph-free, SQL-backed history engine. Operates only on canonical path strings and typed
+/// values: partition-file management, schema, batched write plus periodic flush, raw queries,
+/// look-back, coverage, and metrics. Mirrors the value routing and point mapping of
+/// <c>InMemoryHistoryStoreCore</c> so query results are identical, but persists rows into
+/// partitioned SQLite database files with <c>value_json</c> stored as TEXT.
+/// </summary>
+internal sealed class SqliteHistoryStoreCore : IDisposable
```
RicoSuter: shouldnt we use async apis for sqlite queries etc.?

src/HomeBlaze/HomeBlaze.History.Sqlite/SqliteHistoryStoreCore.cs
```diff
@@ -0,0 +1,1198 @@
...
+{
```
RicoSuter: can we split this into multiple classes (eg 300 locs each)? is there a good seam for that?

src/HomeBlaze/HomeBlaze.History.Sqlite/SqliteHistoryStoreCore.cs
RicoSuter: shouldnt this also implement IHistoryStore so that we can later move it to Namotion.Interceptor.History.*? Not sure the postfix *Core is good, maybe the subject should have the *Subject postfix and core removed? or better idea?

src/HomeBlaze/HomeBlaze.History.Abstractions/HistoryAggregations.cs
```diff
@@ -0,0 +1,43 @@
...
+    public const string Maximum = "Maximum";
+
+    /// <summary>Sum of values in the bucket.</summary>
+    public const string Sum = "Sum";
+
+    /// <summary>Number of samples in the bucket.</summary>
+    public const string Count = "Count";
+
+    /// <summary>Sample standard deviation.</summary>
+    public const string StdDev = "StdDev";
```
RicoSuter: maybe we should not use abbreviations?

src/HomeBlaze/HomeBlaze.History/HistoryStoreMerger.cs
```diff
@@ -0,0 +1,507 @@
...
+namespace HomeBlaze.History.Abstractions;
+
+/// <summary>
+/// Stateless cross-store query merger. Resolves a single <see cref="HistoryQuery"/> against a set
+/// of <see cref="IHistoryStore"/> instances by planning non-overlapping sub-queries (raw coverage
+/// subtraction or per-bucket single-owner dispatch) and executing them under a shared point budget.
+/// The merger is a pure function over the store set, which is the extraction-ready seam: the host
+/// decides where the set comes from (HomeBlaze: the registry's known subjects).
+/// </summary>
+public static class HistoryStoreMerger
```
RicoSuter, line 14: not sure this belongs to history.abstractions, maybe should be in homeblaze.history? abstractions should only be about interfaces, or is this needed for some ref reasons or because it's good design?

src/HomeBlaze/HomeBlaze.History.Abstractions/IHistoryStore.cs
```diff
@@ -0,0 +1,84 @@
...
+    /// </summary>
+    int Priority { get; }
+
+    /// <summary>
+    /// Gets the time window this store can answer queries for.
+    /// </summary>
+    HistoryCoverage Coverage { get; }
+
+    /// <summary>
+    /// Gets the aggregation identifiers (see <see cref="HistoryAggregations"/>) this store supports.
```
RicoSuter, line 29: maybe we change it to ... current... so it's clear that it is a snapshot and might change for every read? or is that clear?

src/HomeBlaze/HomeBlaze.History.Abstractions/HistoryAggregations.cs
```diff
@@ -0,0 +1,43 @@
...
+
+    /// <summary>
+    /// Aggregations every store supports regardless of column type. The global eligibility
+    /// check skips these (see the cross-store merger).
+    /// </summary>
+    public static readonly IReadOnlySet<string> Universal =
```
RicoSuter: Find better name for Universal, isnt clear from naming i think.

src/HomeBlaze/HomeBlaze.History.Abstractions/HistoryColumns.cs
```diff
@@ -0,0 +1,50 @@
...
+/// <summary>
+/// Single source of truth for routing a value into a column (write) and building
+/// column-targeted SQL (read).
+/// </summary>
+public static class HistoryColumns
+{
+    /// <summary>
+    /// Returns the column a value of <paramref name="propertyType"/> is stored in.
+    /// </summary>
+    public static ValueColumn ValueColumnFor(Type propertyType)
```
RicoSuter:
```suggestion
    public static ValueColumn GetValueColumnFor(Type propertyType)
```

src/HomeBlaze/HomeBlaze.History.Abstractions/HistoryStoreMerger.cs
```diff
@@ -0,0 +1,507 @@
...
+    /// For bucketed queries the check runs against the bucket-aligned span the planner actually
+    /// enumerates (from the start of the bucket containing <c>From</c> through the end of the bucket
+    /// containing <c>To</c>), not the raw <c>[From, To)</c>. The bucketed planner assigns each
+    /// epoch-aligned bucket only to a store whose coverage fully contains that bucket, so when
+    /// <c>From</c>/<c>To</c> are not bucket-aligned the edge buckets extend outside <c>[From, To)</c>.
+    /// Checking the aligned span keeps eligibility and per-bucket ownership in agreement: eligibility
+    /// throws iff some bucket the planner would enumerate has no supporting store that fully contains
+    /// it (otherwise that bucket would be a silent gap the eligibility check had promised away).
+    /// </summary>
+    internal static void CheckEligibility(IReadOnlyList<IHistoryStore> ordered, HistoryQuery query)
```
RicoSuter:
```suggestion
    internal static void EnsureEligibility(IReadOnlyList<IHistoryStore> ordered, HistoryQuery query)
```

src/HomeBlaze/HomeBlaze.History.Sqlite/SqliteHistoryStoreCore.cs
RicoSuter: Decided to keep SQLite synchronous. Microsoft.Data.Sqlite's *Async methods run synchronously (SQLite is a synchronous in-process library), and the store serializes all connection access under a single re-entrant lock, so going async would mean reworking that lock into a SemaphoreSlim for no real I/O gain. The async IHistoryStore contract is satisfied via Task.FromResult. The upcoming TimescaleDB store will be genuinely async via Npgsql.

src/HomeBlaze/HomeBlaze.History.Sqlite/SqliteHistoryStoreCore.cs
RicoSuter: Done in 57faf340: split into SqliteHistoryStore (engine, lock + connections + orchestration), SqliteValueRouting, SqliteHistoryWriter, SqliteHistoryReader (raw + look-back + chain), and SqliteBucketReader (bucketed + TWA).

src/HomeBlaze/HomeBlaze.History.Sqlite/SqliteHistoryStoreCore.cs
RicoSuter: Done in 17ca6f1e: the graph-free engine is now the public SqliteHistoryStore : IHistoryStore (the extraction unit ready to lift into a generic library), and the [InterceptorSubject] adapter is renamed SqliteHistoryStoreSubject, which still implements IHistoryStore by delegating so KnownSubjects discovery is unchanged. Same restructure for the InMemory store.

src/HomeBlaze/HomeBlaze.History.Abstractions/HistoryAggregations.cs
RicoSuter: Done in b369b5a4: StdDev renamed to StandardDeviation (constant and value).

src/HomeBlaze/HomeBlaze.History/HistoryStoreMerger.cs
RicoSuter: Done in 4c34ffc8: HistoryStoreMerger moved to a new HomeBlaze.History package. HomeBlaze.History.Abstractions now keeps only the interface, the DTOs, and the pure helpers (HistoryColumns, BucketAlignment, HistoryEligibility, HistoryAggregations).

src/HomeBlaze/HomeBlaze.History.Abstractions/IHistoryStore.cs
RicoSuter: Done in b369b5a4: IHistoryStore.Coverage renamed to CurrentCoverage, with the doc comment clarifying it is a per-read snapshot that can change between reads.

src/HomeBlaze/HomeBlaze.History.Abstractions/HistoryAggregations.cs
RicoSuter: Done in b369b5a4: Universal renamed to AlwaysAvailable.

src/HomeBlaze/HomeBlaze.History.Abstractions/HistoryColumns.cs
RicoSuter: Done in b369b5a4: applied (ValueColumnFor renamed to GetValueColumnFor).

src/HomeBlaze/HomeBlaze.History.Abstractions/HistoryStoreMerger.cs
RicoSuter: Done in b369b5a4: applied (CheckEligibility renamed to EnsureEligibility).

src/HomeBlaze/HomeBlaze.AI/Mcp/HistoryToolParsing.cs
```diff
@@ -0,0 +1,132 @@
...
+namespace HomeBlaze.AI.Mcp;
+
+/// <summary>
+/// Pure parsing and normalization for the <c>get_property_history</c> MCP tool. Kept separate so
+/// the boundary parsing is unit-testable without a live MCP server or registry.
+/// </summary>
+public static class HistoryToolParsing
+{
+    /// <summary>All recognized aggregation identifiers (canonical PascalCase).</summary>
+    public static readonly IReadOnlySet<string> AllAggregations = new HashSet<string>(StringComparer.Ordinal)
```
RicoSuter: Would it be better to dynamically resolve these from the available stores (union)?

docs/superpowers/specs/2026-06-07-history-design.md
```diff
@@ -0,0 +1,573 @@
+# Time-Series History Design (Planning Spec)
```
RicoSuter: delete this file

src/HomeBlaze/Directory.Build.props
```diff
@@ -1,4 +1,8 @@
 <Project>
+  <!-- MSBuild stops at the nearest Directory.Build.props, so without this import the HomeBlaze tree
```
RicoSuter: remove comment, its clear

src/HomeBlaze/HomeBlaze/Data/Docs/architecture/design/ai.md
```diff
@@ -34,7 +34,7 @@ MCP tools are split across two packages, configured via a single `McpServerConfi
 | Package | Tools | Scope |
 |---------|-------|-------|
 | `Namotion.Interceptor.Mcp` | `query`, `get_property`, `set_property`, `list_types` | Any Namotion.Interceptor application |
-| `HomeBlaze.AI` | `list_methods`, `invoke_method` (via `IMcpToolProvider`) + subject enrichment (`$type`, `$icon`, `$title` via `IMcpSubjectEnricher`) + concrete type discovery (via `IMcpTypeProvider`). History tools (`get_property_history`, `get_event_history`, `get_command_history`) planned for later | HomeBlaze-specific features |
+| `HomeBlaze.AI` | `list_methods`, `invoke_method` (via `IMcpToolProvider`) + subject enrichment (`$type`, `$icon`, `$title` via `IMcpSubjectEnricher`) + concrete type discovery (via `IMcpTypeProvider`) + `get_property_history`. Event and command history tools remain planned | HomeBlaze-specific features |
```
RicoSuter: keep method/tool names in docs (events and commands)?

src/HomeBlaze/HomeBlaze.Services/ServiceCollectionExtensions.cs
```diff
@@ -26,7 +26,17 @@ public static IServiceCollection AddHomeBlazeServices(this IServiceCollection se
...
-        services.AddHostedService(sp => sp.GetRequiredService<RootManager>());
+        services.AddScoped<ITimeZoneDisplay, TimeZoneDisplayService>();
+        services.AddHostedService(sp =>
+        {
+            // SubjectPathResolver registers itself as a context service in its constructor, and
+            // subjects attached by RootManager (history stores) need it the moment their hosted
+            // service starts. Force the singleton before RootManager loads the graph; otherwise the
+            // resolver would first be constructed by whichever Blazor component injects it, which
+            // happens after startup.
+            sp.GetRequiredService<SubjectPathResolver>();
```
RicoSuter: is there a cleaner approach here? looks a bit hacky.

src/HomeBlaze/HomeBlaze.Host.Services/Display/StateUnitExtensions.cs
```diff
@@ -17,7 +18,7 @@ private static readonly (StateUnit Unit, string Suffix, decimal Factor)[][] Unit
         [(StateUnit.WattHour, "Wh", 1m), (StateUnit.KilowattHour, "kWh", 1000m)],
         [(StateUnit.Millimeter, "mm", 1m), (StateUnit.Meter, "m", 1000m), (StateUnit.Kilometer, "km", 1_000_000m)],
         [(StateUnit.Milliampere, "mA", 1m), (StateUnit.Ampere, "A", 1000m)],
-        [(StateUnit.Kilobyte, "kB", 1m), (default, "MB", 1000m), (default, "GB", 1_000_000m)],
+        [(StateUnit.Byte, "B", 1m), (StateUnit.Kilobyte, "kB", 1000m), (default, "MB", 1_000_000m), (default, "GB", 1_000_000_000m)],
```
RicoSuter, line 21: should this be 1024, 1024x1024, etc?

src/HomeBlaze/Namotion.Devices.Philips.Hue/HueResourceEcho.cs
```diff
@@ -0,0 +1,28 @@
...
+/// Applies the optimistic local echo after a successful bridge command.
+///
+/// The Hue SDK resources are plain mutable DTOs held behind an intercepted property. Mutating one in
+/// place (<c>LightResource.On.IsOn = true</c>) changes no intercepted property, so nothing recomputed
+/// the derived state, nothing notified the UI and nothing recorded a history sample: the change only
+/// appeared after the next bridge poll, up to a minute later. Re-assigning the same instance does not
+/// help either, because the equality check vetoes a write of the value already there. So the change is
+/// applied to a copy and the copy is assigned, which is exactly what the poll path does.
+/// </summary>
+internal static class HueResourceEcho
```
RicoSuter: echo seems of, maybe just HueResource or HueResourceMutation or something like that?

Issue #347, opened 2026-06-10
Epic: Source transaction correctness and consistency roadmap


RicoSuter:
Tracking issue for the source transaction correctness and consistency work. The goal across all items: a committed value reaches its source exactly once, every failure flow either reverts or converges, divergence is never silent, and the change notification contract (`Source`, later `ChangeOrigin`) is truthful enough to build validation and reconciliation on top of.

The table is the intended working order; dependencies refer to row numbers.

| # | Issue | PR | After | Note |
|---|---|---|---|---|
| 1 | #343 | #344 (merged) | - | **Transaction commit local apply is re-pushed to sources by the change queue (double write, revert race).** Establishes the contract that `Source` means "the system that confirmed this value" and makes the transaction writer the only thing writing to sources during a commit. |
| 2 | #345 | #348 (implementation complete, in review) | 1 | **Publish hook cascade writes with a null source, like derived recalculations.** One rule for framework-invoked consequence writes: everything the local model computes is local-origin and flows to bound sources. Hooks become fully origin-blind (reads too); restoring read-only trigger visibility is row 8. Removes a divergence row from the row 4 design table. |
| 3 | ~~#346~~ | - | - | **Rejected, superseded by #342.** The flush-time echo filter was prototyped and found unsound: moving the skip past the arrival-ordered dedup lets a delayed subscription echo of an older value win the per-property merge and then be skipped, suppressing a newer local write (confirmed by an OPC UA client/server value-loss regression). The confirmed value and a stale echo are indistinguishable at the queue (both `Source == _source`), so no filter-level rule fixes it; it needs versioned provenance or state-based reconciliation. The narrow original race is transient and self-healing, folded into row 4 (#342). See #346. |
[...]

PR #348, opened 2026-06-10
Publish hook cascades, INPC raises and transformed trigger values with a null source [BLOCKED]

### LINKED ISSUE #345: Publish hook cascade writes with a null source, like derived recalculations
## Summary

Cascade writes made from generated property hooks (`OnXChanging`/`OnXChanged`) currently inherit the ambient source scope of the write that triggered them. Derived property recalculations deliberately do the opposite: `DerivedPropertyChangeHandler.NotifyDerivedPropertyChanged` publishes under an explicit `SubjectChangeContext.WithSource(null)` scope, because a derived value is computed by the local model and no source confirmed it. This issue proposes applying the same policy to hook cascade writes, so the system has one strictly truthful rule: `Source` on a change notification marks exactly the values a source confirmed (transaction stage 1) or sent (inbound). Everything the local model computes is local-origin and flows to bound sources like any local write.

Follow-up to #343 (which documented and pinned the current inheritance semantics) and a concrete proposed answer to part of question 8 in #342. Should land after the #343 PR.

## Current behavior

The ambient `SubjectChangeContext` is thread static. When a value is applied under a source scope (inbound `SetValueFromSource`, or a transaction commit apply after #343), any property write made synchronously inside that apply inherits the scope:

| Consequence write | Notification `Source` | Pushed to the owning source? |
|---|---|---|
| `OnXChanging`/`OnXChanged` cascade write | inherited (the source) | no, suppressed as an echo |
| Derived recalculation | always `null` (explicit reset) | yes, always |

The cascade row means a hook-computed value never reaches a source it is bound to when the trigger came from that source (inbound, since forever) or through a source transaction (after #343). The local model and the source silently diverge on that property (divergence row 6 in #342). The derived row shows the codebase already contains the opposite, deliberate policy for the equivalent concept.

## What the hooks are designed for, and how they are actually used

The generator emits `OnXChanging(ref TValue newValue, ref bool cancel)` (pre-write coercion, cancellation, side effects that may throw) and `OnXChanged(TValue newValue)` (post-write reaction). The documented archetype is the hardware bridge (`OnLedAChanged => _gpio.Write(...)`).

Actual in-tree usage (GpioPin, the only production hooks) confirms the pattern: hardware I/O plus cascade writes of locally computed facts, for example `Value = Controller.Read(...)` after a mode switch and `Status = ServiceStatus.Error` / `StatusMessage = ...` on failure. These are the local model computing state about itself, which is exactly the category the derived handler publishes as locally sourced. No in-tree hook writes a source-bound property today and no hook reads `SubjectChangeContext`, so the change breaks nothing in this repository.

## Proposal

Make the generator wrap hook invocations in a null-source scope:

```csharp
using (SubjectChangeContext.WithSource(null))
{
    OnXChanged(value);
}
```

- `WithSource(null)` resets only the source and preserves the ambient timestamps, so cascade timestamp sharing keeps working.
- The generator knows statically which properties declare hooks, so properties without hooks pay nothing. No hot-path checks anywhere.
- Both hooks (`Changing` and `Changed`) get the same treatment.

## Resulting behavior changes

- Transaction commit: a cascade written by a hook during the apply now publishes with `Source = null` and is delivered to its bound source by the background queue, like a non-transactional write (this matches how derived values already behave). The "write the cascade value explicitly into the transaction" guidance from #343 becomes optional instead of required for delivery, though it remains the way to get confirmed, atomic delivery.
- Inbound: a cascade triggered by an inbound source value is now pushed back to the bound source instead of silently diverging. Echo loops are bounded by the value-equality short-circuit; persistent loops require value-changing oscillation (for example lossy conversions), the same accepted risk class as derived recalculation ping-pong.
- The asymmetry table above collapses to one row: all locally computed consequence writes publish with `Source = null`.

## INotifyPropertyChanged handlers

`RaisePropertyChanged` subscribers that write back into the model would still inherit the ambient scope (the raise happens inside the write pipeline; note the derived handler currently raises INPC outside its null-source scope too). Two options:

1. Also wrap the INPC raise in `WithSource(null)` for uniformity, making every framework-invoked callback local-origin.
2. Leave INPC as is and document that INPC handlers must not mutate the model: they exist to react to the event (UI refresh, logging, forwarding), and model mutation from change handlers risks re-entrancy and stack overflows regardless of source semantics.

Recommendation: document the prohibition in either case; option 1 is cheap defensive consistency and fixes the existing inconsistency where derived INPC raises escape the null scope.

## Scope of work

- Generator: wrap hook invocations (and INPC raise if option 1) in `WithSource(null)`.
- Tests: flip the two cascade scope tests in `SubjectTransactionEchoSuppressionTests` (commit and inbound trigger now expect `Source = null` and delivery), add an inbound cascade delivery test at the pump level, keep the derived test unchanged.
- Docs: update the "Change notification source semantics" section in connectors.md and the stage 2 paragraph in tracking-transactions.md (replace the cascade-inheritance rule), note the callback-purity guidance, touch generator.md where hooks are described.
- Out of scope: arbitrary user code running inside a `WithState` scope (for example code called from a custom source implementation) keeps ambient inheritance; the rule is scoped to framework-invoked consequence callbacks. Lifecycle handlers stay as they are.

## Relationship to other issues

- #343 defined and pinned the current semantics; this issue revises the cascade half after the derived half turned out to already implement the proposed policy.
- #342 question 8 (consequence-write policy): this is the concrete proposed answer for hook cascades and INPC; reconciliation concerns in #342 shrink accordingly (divergence row 6 disappears if this lands).

RicoSuter:
Closes #345.

Publishes framework-invoked consequence writes with a null source so the system has one truthful rule: a change carries a `Source` only when its stored value is exactly the value that source confirmed (transaction stage 1) or sent (inbound); everything the local model computes is local-origin and flows to bound sources like any local write. Hook cascade writes (`OnXChanging`/`OnXChanged`) and INPC handler write-backs align with the policy derived recalculations already implement, so a hook-computed value bound to a source no longer silently diverges when the trigger came from that source.

The rule also covers transformed trigger values: when an `OnXChanging` hook changes the incoming value itself (a clamp or normalization via `ref newValue`), the write publishes with a null source and the corrected value flows back to the bound source, converging the source to the model instead of silently diverging. Such transforms must be projections (idempotent) so the correction terminates; see the notes in docs/connectors.md. The correction is published only when the transform actually changes the stored value: an inbound value whose projection equals the current stored value is dropped by the value equality check, so no correction is sent and the source keeps its out-of-range value until the next model change. Reasserting the model value to a diverged source is a source-binding concern left to the typed change-origin work (#365). Transaction commits are naturally consistent: the hook transforms at capture, so the source confirms the already-transformed value and the commit apply keeps its source stamp truthfully.

Inside the local-origin scope hooks are intentionally origin-blind: they cannot observe which source triggered the cascade (`SubjectChangeContext.Current.Source` reads null), only the ambient timestamps remain visible. Restoring read-only trigger visibility as causal metadata is a separate feature, the typed `ChangeOrigin` discriminator with `TriggeredBy` tracked in #365.

The runtime surface is one new scope, `SubjectChangeContext.WithLocalOrigin()`, with a fast path when the ambient source is already null; the remaining changes are in the generator: hook calls are wrapped in a local-origin scope only for properties whose hooks are actually implemented, the transform comparison runs only under an ambient source, and the generated `RaisePropertyChanged` checks subscribers first and enters the scope only when a source is set. The local-origin behavior of the raise is a contract on `IRaisePropertyChanged` itself: generated implementations fulfill it, and hand-written implementations get it in one call by forwarding to the new `SubjectChangeContext.RaisePropertyChanged(handler, sender, propertyName)` helper (documented on the interface and in docs/subject-guidelines.md). When a base exposes the raise only through an explicit interface implementation, the generator emits a protected forwarder so setters and descendants keep a direct call path.

## Breaking behavior

- Hook cascade writes triggered by an inbound update or a commit apply were previously suppressed toward that source as echoes; they are now delivered to bound sources like any local write. Deployments where hooks compute values bound to writable device nodes will start writing those values to the device.
- A value transformed by an `OnXChanging` hook during a source apply (a clamp or normalization) is now pushed back to the bound source as a correction instead of silently diverging. Transforms must be idempotent projections.
- Hooks and INPC handlers always observe a null `SubjectChangeContext.Current.Source`; code that branched on the ambient source inside these callbacks must move that logic to an interceptor or the change observable (see #365 for planned read-only trigger visibility).
- Hand-written `IRaisePropertyChanged` implementations must now invoke their handlers under `SubjectChangeContext.WithLocalOrigin()`; an implementation that does not is out of contract and its handlers observe the ambient source. Forward to `SubjectChangeContext.RaisePropertyChanged(PropertyChanged, this, propertyName)` for the compliant behavior in one line; see the manual INPC base section in docs/subject-guidelines.md.

The first commit is the agreed design spec under docs/superpowers/specs/; the implementation follows on this branch. Part of the #347 roadmap, lands after #344.


RicoSuter:
Superseded by #366 (typed ChangeOrigin with one-shot per-write source stamping, design in docs/superpowers/specs/2026-07-12-change-origin-design.md). The mechanism moved from scope-based counter-scopes at every callback site to a per-write origin stamp that makes local-by-default structural: the generator stays untouched, the manual IRaisePropertyChanged contract is unnecessary, and the four documented holes of this approach are closed. The behavioral tests from this branch were ported to #366 as its semantic safety net and all pass under the new mechanism.

PR #350, opened 2026-06-21
Promote ThroughputCounter and add opt-in ChangeQueueProcessor backpressure

RicoSuter:
## Summary

Two independent shared-infrastructure changes that the upcoming HomeBlaze time-series history feature depends on (Phase 0 of the history design). Both are independently useful and leave existing connectors unchanged.

- Promote `ThroughputCounter` from `internal sealed` in `Namotion.Interceptor.OpcUa` to `public sealed` in `Namotion.Interceptor.Connectors`, so OPC UA connectors and the history stores can share one lock-free 60-second sliding-window rate counter. Pure move, behavior unchanged; its test moves to `Namotion.Interceptor.Connectors.Tests`. No new project reference (OPC UA already references Connectors).
- Add opt-in bounded-queue backpressure to `ChangeQueueProcessor`: a new optional `maxQueueDepth` constructor parameter (default `null` = unbounded, preserving current behavior) plus a `DropCount` metric. When a bound is set and exceeded, the oldest unprocessed buffered change is dropped (best-effort, never blocks the hot path) and `DropCount` increments. The bound applies only to the buffered queue; the immediate path is untouched.

## Test plan

- `dotnet build src/Namotion.Interceptor.slnx` clean (warnings as errors).
- `dotnet test src/Namotion.Interceptor.slnx --filter "Category!=Integration"` passes.
- New tests: overflow drops the oldest entry and increments `DropCount`; the default `null` path retains every change (existing connectors unaffected).
- Public API snapshot accepted for `Namotion.Interceptor.Connectors` (adds `ThroughputCounter`, the `maxQueueDepth` ctor parameter, and `DropCount`).

### DIFF
--- a/src/Namotion.Interceptor.Connectors.Tests/ChangeQueueProcessorTests.cs
+++ b/src/Namotion.Interceptor.Connectors.Tests/ChangeQueueProcessorTests.cs
@@ -231,6 +231,74 @@ public class ChangeQueueProcessorTests
         Assert.Equal(1, flushCount);
     }
 
+    [Fact]
+    public void WhenBoundedQueueOverflows_ThenOldestDroppedAndDropCountIncrements()
+    {
+        // Arrange
+        var context = new InterceptorSubjectContext();
+        context.WithRegistry();
+        context.WithPropertyChangeQueue();
+
+        var subject = new Person(context);
+        var property = new PropertyReference(subject, nameof(Person.FirstName));
+
+        var processor = new ChangeQueueProcessor(
+            source: null,
+            context: context,
+            propertyFilter: _ => true,
+            writeHandler: (_, _) => ValueTask.CompletedTask,
+            bufferTime: TimeSpan.FromMilliseconds(50),
+            logger: NullLogger.Instance,
+            maxQueueDepth: 2);
+
+        // Act - enqueue five changes into a buffer bounded to two
+        EnqueueBuffered(processor, property, null, "v1");
+        EnqueueBuffered(processor, property, "v1", "v2");
+        EnqueueBuffered(processor, property, "v2", "v3");
+        EnqueueBuffered(processor, property, "v3", "v4");
+        EnqueueBuffered(processor, property, "v4", "v5");
+
+        processor.Dispose();
+
+        // Assert - three oldest dropped, the two newest (v4, v5) retained in order
+        Assert.Equal(3, processor.DropCount);
+        Assert.Equal(new[] { "v4", "v5" }, RemainingNewValues(processor));
+    }
+
+    [Fact]
+    public void WhenUnbounded_ThenNothingIsDroppedAndDefaultPathIsUnchanged()
+    {
+        // Arrange
+        var context = new InterceptorSubjectContext();
+        context.WithRegistry();
+        context.WithPropertyChangeQueue();
+
+        var subject = new Person(context);
+        var property = new PropertyReference(subject, nameof(Person.FirstName));
+
+        // No maxQueueDepth argument: default null == unbounded (existing connector behavior)
+        var processor = new ChangeQueueProcessor(
+            source: null,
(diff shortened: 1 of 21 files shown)


src/Namotion.Interceptor.Connectors/ThroughputCounter.cs
```diff
@@ -1,6 +1,10 @@
-namespace Namotion.Interceptor.OpcUa;
+namespace Namotion.Interceptor.Connectors;
 
-internal sealed class ThroughputCounter
+/// <summary>
+/// Lock-free 60-second sliding-window rate counter (changes per second).
+/// Shared by OPC UA connectors and the history stores for incoming/recorded rate metrics.
```
RicoSuter: remove this line, not relevant as docs (users is not necessary to document).

src/Namotion.Interceptor.Connectors/ChangeQueueProcessor.cs
```diff
@@ -56,19 +64,24 @@ public class ChangeQueueProcessor : IDisposable
...
+    /// the oldest unprocessed change and increments <see cref="DropCount"/>, so the newest change is retained.</param>
     public ChangeQueueProcessor(
         object? source,
         IInterceptorSubjectContext context,
         Func<PropertyReference, bool> propertyFilter,
         Func<ReadOnlyMemory<SubjectPropertyChange>, CancellationToken, ValueTask> writeHandler,
         TimeSpan? bufferTime,
-        ILogger logger)
+        ILogger logger,
+        int? maxQueueDepth = null)
```
RicoSuter: can we keep logger as last parameter and maxQueueDepth without default, maybe add config to all places where CQP is created with default null?

src/Namotion.Interceptor.Connectors.Tests/ChangeQueueProcessorTests.cs
```diff
@@ -264,4 +332,28 @@ private static async Task TriggerFlushAsync(ChangeQueueProcessor processor)
...
+
+        var change = SubjectPropertyChange.Create(
+            property, null, DateTimeOffset.UtcNow, null, oldValue, newValue);
+
+        method!.Invoke(processor, [change]);
+    }
+
+    private static string?[] RemainingNewValues(ChangeQueueProcessor processor)
+    {
+        var changesField = typeof(ChangeQueueProcessor)
```
RicoSuter: can we avoid using reflection in tests?

src/Namotion.Interceptor.Connectors/ThroughputCounter.cs
RicoSuter: Removed that line. The summary now only describes what the counter does.

src/Namotion.Interceptor.Connectors/ChangeQueueProcessor.cs
RicoSuter: Done in a05741c8. maxQueueDepth is now required and placed before logger. It is threaded through every CQP creation site: a MaxQueueDepth property (default null, meaning unbounded) was added to the OPC UA, MQTT and WebSocket server and client configurations, and the client sources receive it through a new optional SubjectSourceBase parameter.

src/Namotion.Interceptor.Connectors.Tests/ChangeQueueProcessorTests.cs
RicoSuter: Rewrote both backpressure tests to drive ProcessAsync with real property changes (a large bufferTime so the periodic flush never drains the buffer, and a sentinel source so the direct changes are not filtered out), and removed the reflection helpers. The bounded test asserts DropCount; verifying the exact surviving values without reflection would need a blocked-flush gate, which I can add if you want it.

PR #354, opened 2026-07-06
Add source monitoring (synchronization state, source event stream)

RicoSuter:
## Summary

Consumers of a subject tree cannot tell whether a property value is actually in sync with its external source. At startup the tree shows unconfirmed default values while sources connect asynchronously, and after a connection loss it silently shows stale values. This PR adds a source-level synchronization state, a typed event stream for source metadata, a branch-scoped awaitable primitive, and a derived per-property state.

Documentation is in `docs/connectors-monitoring.md`.

| | added | removed |
|---|---:|---:|
| Production code | 1567 | 56 |
| Tests | 3879 | 55 |
| Documentation | 321 | 3 |
| API snapshots | 98 | 1 |
| **Total** | **5865** | **115** |

Tests are roughly two thirds of the change, which is deliberate: every behaviour the design relies on is pinned by a test verified to fail when the production line it covers is reverted, including the concurrency ones. One exception is called out honestly under Review history.

**The API is the deliverable.** Nothing in this repository consumes `WithSourceMonitoring` or `DeferWaitCompletion` outside tests and the documentation. That would normally breach the rule that a PR should not add public API whose callers land later, but that rule exists to stop dead code accumulating behind a promise, and here the API is the product: the consumer is an external application built on this library, outside this repository. Its ergonomics therefore rest on the worked examples in `docs/connectors-monitoring.md` rather than on an in-repo caller.

## The API

```csharp
// Wait until every source that can claim into this branch has finished its initial load.
await root.Plants["LineA"].WaitForSynchronizationAsync(cancellationToken);

// Now well defined per property.
var state = property.GetSourceState();   // Unclaimed | Connecting | Synchronized | Stopped
```

Waiting on the tree root means the whole tree, so one method covers both cases. A source is in scope when its `RootSubject` and the anchor lie on the same root-to-leaf path, in either direction, so a failing source on a sibling branch cannot block a consumer that does not depend on it.

There are two ways to declare registration complete, and no third.

```csharp
.WithSourceMonitoring(builder.Services)   // completes when host startup finishes
```

covers every source that exists by the end of host startup, whether registered directly in DI or attached to the subject graph. Applications that create sources *after* startup use the parameterless overload and signal for themselves:

```csharp
await LoadAsync(stoppingToken);
_context.CompleteSourceRegistration();
```

## What it adds

- `SourceState` (`Unclaimed`, `Connecting`, `Synchronized`, `Stopped`), with `Stopped` terminal and enforced by a guard in `StartAsync`
- `ISubjectSource` gains `State`, `LastSynchronizedAt`, `PendingWriteCount` and a `StateChanged` event. Breaking for direct implementers; free for anything deriving from `SubjectSourceBase`
- A typed `SourceEvent` stream covering registration, state changes and ownership (`PropertyClaimed`, `PropertyReleased`), delivered on per-subscriber queues so a slow handler delays only itself
- `SourceEvent.CurrentState`, an authoritative read, plus `SubjectPropertyChange.GetCurrentValue<TValue>()` giving the same idiom to property changes
- A registration count that is fail-safe by default: forgetting to signal hangs a wait, which is loud, rather than completing it early on a partially registered tree, which is silent
- `IStartupCompletionDeferrer`, which lets the hosting layer hold registration open from the moment a source is attached until its queued start has actually run. Both attach paths take holds: awaiting the start blocks the caller, but not whatever else decides startup is finished
- `PropertyReference.TryAddPropertyData`, the atomic add-if-absent primitive backing exact claim emission

## Three decisions worth knowing about

**State transitions are driven from `SubjectPropertyWriter`, not from the pump.** The obvious placement is wrong: the built-in clients handle ordinary reconnects themselves and call `StartBuffering` and `LoadInitialStateAndResumeAsync` from five sites outside `ExecuteAsync` while the pump sits in `ProcessAsync`. A pump-placed transition reports `Synchronized` straight through every real outage. Driving from the writer, which all five paths already use, means no connector needed changes. OPC UA additionally needs `ReportConnectionLost()` at keep-alive failure, because it detects loss before it buffers.

**`CurrentState` is what consumers apply, not `NewState`.** An ownership event is enqueued after its compare-and-set, so same-property events can arrive inverted. `OldState`/`NewState` describe one transition and are for logging; applying them blindly can leave a view permanently diverged.

**Source monitoring reports ownership, not graph membership.** It answers which source owns a property and what state that source is in. Whether a subject is currently in the object graph is a registry question, so ask `ISubjectRegistry.TryGetRegisteredSubject`. In practice the two rarely need distinguishing, because every built-in connector releases its claims on detach, which publishes `PropertyReleased`.

## Known limits, stated rather than implied

**MQTT's `Synchronized` is weaker.** It means subscriptions are established, not that retained values have arrived, because `LoadInitialStateAsync` returns null and MQTT has no end-of-retained signal in 3.1.1 or 5.0. Raising QoS does not change this. Documented rather than hidden by excluding MQTT from waits, which would make a tree-wide wait lie by omission. Opt-in barrier tracked in #418.

**No multi-tree coverage.** `ContextInheritanceHandler` records only the first attach, so a subject in two trees resolves the first tree's monitor alone. A characterization test pins this so it fails loudly if that ever changes.

**The stream is a notification, not a ledger.** Same-property events can be enqueued out of order, so it cannot reconstruct history. Versioned ownership entries are recorded as deferred, with the working lock-free mechanism written down.

**An empty wait scope completes rather than blocks.** Correct for a genuinely local-only branch, and indistinguishable from a branch whose source was never created. Once the application has called `CompleteSourceRegistration` it has asserted its sources are set up, so a branch with nothing to wait for is treated as already synchronized. A completed wait means "nothing here is still loading", not "this branch is live".

## Testing

2898 unit tests pass. OPC UA (291), MQTT (74) and WebSocket (130) integration suites pass locally, including outage tests for all three connectors that assert the source reports `Connecting` for the duration of a real fault-injected outage and recovers afterwards.

`SourceRegistrationHostingTests` is the first test in the repository that references both `Namotion.Interceptor.Hosting` and `Namotion.Interceptor.Connectors`. The two are siblings that do not reference each other, which is why their interaction had no coverage at all, and is how the registration-timing defect below survived every earlier review.

Both connector benchmarks were compared against master with `scripts/benchmark.ps1`. `SubjectSourceBenchmark` and `SubjectTransactionBenchmark` are flat: every mean sits inside its error bar, and allocated bytes are identical on both sides (12289 B, and 4.96 / 12.41 / 42.07 KB across the transaction modes). Allocation is the number worth reading there, since the timings on a laptop are noisier than the deltas.

One gap worth stating rather than implying: `SetSource` appears only inside `[GlobalSetup]` in both benchmarks, so the per-claim cost this feature adds is not covered by either. That cost is a cached `GetServices<SourceMonitor>()` lookup, argued from the code rather than measured.

## Review history

The design went through four review rounds before implementation and two after. Six parallel reviews (correctness, thread safety, simplification, performance, API surface, test honesty) ran against the finished branch, then five more against the merged result, each scoped to one high-risk area.

Defects caught and fixed along the way: transitions had to move off the pump; `BackgroundService` turned out to be restartable, so terminal `Stopped` needed a real guard rather than an assumption; a cycle in the parent graph could livelock the ancestor walk while holding the monitor's lock; `Register`/`Unregister` published outside the lock that makes the subscription snapshot exactly-once; and the wait engine's diagnostic warnings were silently dead in the configuration the documentation leads with.

The last round found four more, all fixed and each pinned by a test verified to fail when the fix is reverted:

- **Registration could complete before attached sources had registered.** The gate fires on `ApplicationStarted`, but attaching a source queues its `StartAsync` rather than running it inline, so a wait saw an empty scope and completed vacuously against an unsynchronized tree. Hosting now holds registration open across that gap.
- **A disposed source could stay registered forever.** `StartAsync`'s unwind re-read `_registeredMonitors` instead of the array it registered with, so a `Dispose` landing between the assignment and the loop stranded the registration, leaking the source, its writer, its retry queue and its root subject.
- **`StateChanged` could be delivered before `SourceRegistered`,** leaving a `NewState`-based consumer permanently one state stale.
- **The empty-scope warning logged once per call** rather than once per branch. Later removed entirely on review: it tried to distinguish a legitimately source-less branch from a forgotten source, which it cannot do, and it had managed to be silently dead in one configuration or another three separate times.

Two tests were hollow and have been rebuilt: the reparent test completed vacuously before the reparent it was named after, and the throw-isolation test drove its throw from a diagnostic that had since become one-shot. Both now fail when the code they pin is broken.

The same round removed `PropertyEnteredView` and `PropertyLeftView`, which never fired for any shipped connector: `SourceOwnershipManager` releases claims on `SubjectDetaching`, which runs before `ILifecycleHandler.HandleLifecycleChange`, so the monitor's scan always found no owner. Removing them also removed the tree-membership tracking that existed only to serve them, and with it a lock-order inversion where seeding walked the lifecycle interceptor's subject set from inside a `TryAddService` factory, a nesting `InterceptorSubjectContext` documents as forbidden.

A final round ran four reviewers over the whole branch: correctness and concurrency, test honesty, public API and documentation accuracy, plus a targeted regression check on a cleanup pass. It found that the cleanup had introduced two regressions of its own, both fixed: a lazily resolved logger that latched a null result and killed every wait-engine warning, and a deduplication that made the hot wait-re-evaluation path allocate. It also found three more defects, all fixed: the awaiting attach overload took no startup hold, `HostedServiceHandler` relied on registration order for an ordering it needed by construction, and a subscription captured its logger at subscribe time and silently swallowed handler exceptions for its lifetime.

The test-honesty audit ran 26 mutations against the suite. Twenty-three were killed. The three survivors are now covered: wait-list cleanup on both the completion and cancellation paths, the drain handoff (extracted so it can be driven directly, with its call site pinned separately, because testing the method does not prove the loop calls it), and `Register`'s own re-evaluation, which turned out to be unreachable rather than untested and is now documented as defensive.

A last pass took the three remaining edge cases, each of which did damage out of proportion to its cause. A wait re-evaluation could throw out of `LifecycleInterceptor`'s attach lock and leave the graph half-attached, with later handlers skipped and child properties never attached. A transition already in flight when `Unregister` ran published `StateChanged` after `SourceUnregistered`, handing consumers a state for a source they had already removed. And a source both registered in DI and attached to the graph ran two pumps against one connection, where the first to exit latched `Stopped` terminally while the second was still applying live values. All three are fixed and pinned.

One caveat worth stating rather than burying. The store-load reordering fix in the subscription drain rests on the memory-model argument and a reviewer's harness, not on a test in the suite: the race reproduces roughly once in 500 million aligned attempts, which is beyond unit-test reach. A static source-scan test pins the `Interlocked.Exchange` at that call site so the intent cannot be silently reverted.

## After review

The branch was reviewed by five independent agents in a final round (correctness, thread-safety, simplification, public API and naming, documentation). Correctness and thread-safety returned no Critical or Major finding, having re-derived the lock graph, the satisfaction matrix, hold arithmetic and the wakeup ordering by hand.

Acting on that round and on PR feedback, three things were then deleted rather than fixed: the empty-scope and all-`Stopped` warnings with their latches and the `ILoggerFactory` requirement, the tree-membership tracking behind two event kinds that never fired, and the hosted-service barrier that the startup holds made unnecessary. `WithSourceMonitoring` moved to its package root to match the other sixteen `With*` builders, and `SubjectPropertyWriter` now takes `SubjectSourceBase` rather than the interface it could only downcast.


RicoSuter:
> the anchor's scope matches no source at all (a mistyped root, a device that was never configured), the wait never completes

can we change that wait completes immediately as there is no source to wait for? why block forever?

RicoSuter:
## Known caveats

Rewritten after the final review rounds. Three earlier caveats no longer exist, because the machinery they described was removed rather than fixed, and everything the last rounds found was fixed rather than accepted. What remains below is genuinely accepted: protocol facts, documented contracts, and one coverage note. Everything a consumer needs to know is in `docs/connectors-monitoring.md`; the "Documented" column says where.

| # | Caveat | Status | Documented |
|---|---|---|---|
| 1 | Store-load reordering in the subscription drain | **Accepted, guarded.** The hardware race reproduces roughly once in 500 million aligned attempts and is not reachable by a unit test. A static source-scan test pins `Interlocked.Exchange` at that call site and fails if it is reverted to `Volatile.Write`, so the intent is guarded even though the race is not | Code comment at the call site |
| 2 | ABBA hazard for custom `ISubjectSource` implementers | Accepted, and narrower than it reads. `SubjectSourceBase` supplies `State`, `LastSynchronizedAt` and `StateChanged` lock-free, and every built-in connector derives from it, so this needs someone to implement the interface from scratch and to wrap a trivial getter in the lock they raise events under. The alternative, caching source state in the monitor, trades a documented contract for a coherence problem and a staleness window in the wait | Yes, plus XML docs on the members |
| 3 | An empty wait scope cannot be distinguished from a missing source | Accepted, and inherent. A branch with no source is legitimate and common; a forgotten source looks identical from inside. No warning is emitted: once the application has called `CompleteSourceRegistration` it has asserted its sources are set up, so a branch with nothing to wait for is treated as already synchronized | Yes |
| 4 | MQTT's `Synchronized` means subscriptions established, not retained values received | Accepted. A protocol limitation: neither 3.1.1 nor 5.0 defines an end-of-retained signal, and QoS does not change it | Yes, with #418 |
| 5 | A subject shared across two trees participates fully in only the first | Accepted. Context inheritance adds the parent fallback on first attach only | Yes |
| 6 | The stream is a notification, not a ledger | Accepted. Same-property events are enqueued after their compare-and-set and can arrive inverted, so the stream cannot reconstruct transition history. Consumers maintain views, not logs | Yes |
| 7 | Per-protocol behaviour is pinned only by `Category=Integration` tests | Accepted. They need exclusive ports; all three connector suites pass | n/a |

### Removed rather than resolved

Three entries from the previous version are gone because the code behind them is gone:

**Tree-membership tracking in `CurrentState`.** `PropertyEnteredView` and `PropertyLeftView` never fired for any shipped connector, so the membership table, `IsMember`, `SourceEvent.Monitor` and the seeding of pre-existing subjects all served nothing. Source monitoring now reports ownership only; graph membership is a registry question. This also deleted a lock-order inversion, since seeding walked the lifecycle interceptor's subject set from inside a `TryAddService` factory.

**The hosted-service barrier.** `WaitForPendingHostedServiceActionsAsync` existed so an application could avoid declaring registration complete while an attached source was still on its way in. Hosting now holds registration open itself, from the attach until the queued start has run, so the barrier and the "did you remember to call it?" footgun are both gone.

**The empty-scope warning, and the warning-once test behind it.** The warning was removed entirely rather than fixed: it existed to hint that an empty scope might be a forgotten source, which it cannot actually determine, and it had managed to be silently dead in one configuration or another three separate times. Removing it took both warn-once latches, the all-`Stopped` warning and the requirement that an `ILoggerFactory` be reachable from the context with it, and reduced `IsSatisfied` to "registration complete, and every in-scope source is `Stopped` or `Synchronized`".

### Closed by the final review round

Four reviewers ran over the whole branch (correctness and concurrency, test honesty, API and documentation accuracy, and a regression check on a cleanup pass). Everything they found is fixed rather than accepted, so none of it appears above:

- A cleanup pass had introduced two regressions of its own: a lazily resolved logger that latched a null result and killed every wait-engine warning, and a deduplication that made the hot wait-re-evaluation path allocate.
- A subscription captured its logger at subscribe time, silently swallowing every handler exception for its lifetime.
- The awaiting attach overload took no startup hold; blocking its caller does not block the startup gate.
- `HostedServiceHandler` relied on registration order for an ordering it needs by construction.
- The empty-scope warning could not fire at all under the parameterless overload, which is the one the docs recommend for runtime sources.
- Three mutations survived the test audit: wait-list cleanup, the drain handoff, and `Register`'s own re-evaluation. The first two are now covered; the third turned out to be unreachable rather than untested and is documented as defensive.

A final pass closed the last three, each of which did damage out of proportion to its cause:

- A wait re-evaluation could throw out of `LifecycleInterceptor`'s attach lock, leaving the graph half-attached: later handlers skipped, `SubjectAttached` never raised, child properties never attached. The interceptor documents its handlers as exception-free, so this now conforms rather than relying on nothing ever throwing.
- A transition already in flight when `Unregister` ran had captured the handler list before the `-=`, so it published `StateChanged` after `SourceUnregistered`. It is now dropped rather than reordered, which is the right call for a stream that is a notification rather than a ledger.
- A source both registered in DI and attached to the subject graph ran two pumps against one connection; the first to exit latched `Stopped` terminally while the second was still applying live values. A second start is now a logged no-op.

Six test-quality items were closed alongside them, including a genuinely hollow test, a race harness that caught its own timeout as the pass path (so under load it silently tested nothing), and new public core API that had no direct test.

### Not caveats, but worth knowing

The feature now holds zero weak references. The two that existed, membership tracking and warn-once bookkeeping, are gone and replaced respectively by nothing and by subject-scoped data with the subject's own lifetime.

`SubjectSourceExtensions.WriteLocks` is a pre-existing `ConditionalWeakTable` on master, unrelated to this PR, carrying its own TODO about the `SemaphoreSlim` it never disposes. Flagged rather than pulled into scope.

### Removed after the final review round

Three more pieces went rather than being carried: the empty-scope warning above, `WithSourceMonitoring`'s placement outside its package root (it was the only one of seventeen `With*` builders not in `InterceptorSubjectContextExtensions`), and `SubjectPropertyWriter`'s `ISubjectSource` parameter, which it could only use via a downcast that silently did nothing for a direct implementer.

The feature page and the code comments were also cut back to what is not recoverable from the code; comment density in the monitoring files went from 32-54% down to 26-37%.

docs/connectors-monitoring.md
```diff
@@ -0,0 +1,307 @@
...
+
+Two edge cases in that scoping matter before you rely on a wait.
+
+No wait can complete before source registration is complete, whatever its scope: that is the first condition checked, ahead of any scope evaluation. So until `CompleteSourceRegistration()` has run (and any `DeferWaitCompletion()` holds are released), every wait blocks.
+
+What registration completion changes is the meaning of an empty scope. Before it, a scope that matches nothing is ambiguous: it could mean "no source for this branch, ever" or merely "not registered yet". After it, the application has declared its source set complete, so a scope that still matches no source means the branch is local-only, and the wait completes immediately, the same way an all-`Stopped` scope does below.
```
RicoSuter: this section and the next seems strangely written or is unclear to me. so before complete the wait will block until complete is done, right? after that it should only wait for the sources it can find at that time, is that correct?

docs/connectors-monitoring.md
```diff
@@ -0,0 +1,307 @@
...
+public enum SourceState
+{
+    Unclaimed,
+    Connecting,
+    Synchronized,
+    Stopped
+}
+```
+
+| State | Meaning |
```
RicoSuter: should we avoid repeating the enum members in this doc maybe?

docs/connectors-monitoring.md
```diff
@@ -0,0 +1,307 @@
...
+| `SourceUnregistered` | `null` | A source unregistered, which happens when it is disposed. |
+| `StateChanged` | `null` | A source's own state changed. |
+| `PropertyClaimed` | set | A source took ownership of a property. |
+| `PropertyReleased` | set | A source gave up ownership of a property. |
+| `PropertyEnteredView` | set | An already-claimed property joined the tree when its subject attached; ownership did not change. |
+| `PropertyLeftView` | set | A still-claimed property left the tree when its subject detached; ownership did not change. |
+
+`SourceMonitor.Publish` enqueues every event onto each subscriber's own queue; each subscription drains its own queue on a single worker at a time, so a slow handler delays only that subscription. There is no ordering guarantee across subscriptions: two subscribers can observe events in different relative orders under concurrent activity.
+
+## Breaking Change for Custom ISubjectSource Implementers
```
RicoSuter: the document here should only document the latest state not migration notes, also there is currently no implementor of this interface outside this repo and in this repo we already migrated, we can drop that chapter - maybe we should just describe how to implement it (not sure it's needed as the xml docs on the interface should be clear enough, no?)

docs/connectors-monitoring.md
```diff
@@ -0,0 +1,307 @@
...
+
+Deriving from `SubjectSourceBase` is recommended instead of implementing `ISubjectSource` directly: it implements all four members, drives the connection-phase transitions automatically through `SubjectPropertyWriter`, registers and unregisters with every reachable monitor around its pump lifecycle, and is what every built-in connector (OPC UA, MQTT, WebSocket) already does. See [Implementing a Source](connectors.md#implementing-a-source) for the base class's hooks.
+
+## Worked Sample: Availability Attributes
```
RicoSuter: not sure this is a good sample, maybe we should make IsAvailable dynamic attribute of an actually bound property?

docs/connectors-monitoring.md
RicoSuter: Rewritten and shortened in 741d7795.

Your first half was right: before `CompleteSourceRegistration()` (and while any `DeferWaitCompletion()` hold is outstanding) every wait blocks, whatever its scope, because registration is checked ahead of any scope evaluation.

The second half needed correcting, and the doc now says so explicitly: a pending wait is not limited to the sources present when registration completed. Satisfaction is re-evaluated live, so a source registering later is picked up and can re-block a wait that has not finished yet. Only an already-completed wait is frozen, since a completed task cannot be un-completed.

docs/connectors-monitoring.md
RicoSuter: Done in 741d7795. Both `SourceState` member tables are gone; the page now describes what the states mean for a consumer deciding what to do, and the enum's XML docs carry the authoritative member list, so the two cannot drift.

docs/connectors-monitoring.md
RicoSuter: Agreed and deleted in 741d7795.

This was the right call for a reason beyond tidiness: that chapter produced three separate defects across three review rounds (a wrong accessibility claim, a reference to an internal method, and guidance inverted by the empty-scope change), precisely because nothing in the repo exercises it. Every connector derives from `SubjectSourceBase`, so no compiler or test ever checked that prose.

Three things in it were load-bearing, so they moved to XML docs on `ISubjectSource` rather than being lost: the CS0067 stub note onto `StateChanged`, and the register-with-every-reachable-monitor requirement (via `GetServices<SourceMonitor>()`, not the throwing `GetSourceMonitor()`) onto the type. The lock-free getter contract was already there. The dangling link in connectors.md was repointed.

docs/connectors-monitoring.md
RicoSuter: Reworked in 741d7795 along the lines you suggested. `IsAvailable` is now a derived attribute on the bound property via `AddDerivedAttribute`, read with `TryGetAttribute("IsAvailable")`, instead of a hand-rolled `ConnectionState` shadow on a `Device` class.

The sample was verified by compiling the exact code against the test project before committing, since several earlier doc samples here did not compile. It keeps the one correctness point it exists for: apply `event.CurrentState` for the four property event kinds, but call `property.GetSourceState()` per property on `StateChanged`, because `CurrentState` on a source-level event is the source's state and says nothing about any individual property.

src/Namotion.Interceptor.Tracking/IStartupCompletionDeferrer.cs
```diff
@@ -0,0 +1,28 @@
...
+/// start is still on its way in, completing a wait against a tree that is not synchronized. Those
+/// two packages are siblings and neither references the other, so the handshake lives here, in the
+/// package they both build on.
+/// <para>
+/// A holder takes a hold before queueing the work and disposes it once the work has run, including
+/// when the work throws. Holds are counted, so concurrent holders compose. Taking a hold never
+/// un-completes a signal that has already fired; it only blocks one that has not.
+/// </para>
+/// </remarks>
+public interface IStartupCompletionDeferrer
```
RicoSuter, line 17: why in tracking and not in connectors/monitoring?

src/Namotion.Interceptor.Connectors/SubjectPropertyWriter.cs
```diff
@@ -15,11 +16,18 @@ namespace Namotion.Interceptor.Connectors;
...
     private readonly ILogger _logger;
     private readonly Lock _lock = new();
 
     private List<Action>? _updates = [];
 
+    // Bumped by every StartBuffering call. LoadInitialStateAndResumeAsync captures the generation
+    // in effect when it starts and compares it again after its (possibly long) await: if a later
+    // StartBuffering happened in between, this call's snapshot is stale and must not be applied,
+    // replayed, or certified as Synchronized - see LoadInitialStateAndResumeAsync.
+    private int _generation;
```
RicoSuter, line 28: use long to avoid overflow? or overkill?

src/Namotion.Interceptor.Connectors/SubjectPropertyWriter.cs
```diff
@@ -28,6 +36,7 @@ public sealed class SubjectPropertyWriter
     public SubjectPropertyWriter(ISubjectSource source, ILogger logger)
     {
         _source = source;
+        _sourceBase = source as SubjectSourceBase;
```
RicoSuter: seems a bit hacky, will it also work for ISubjectSources? or is this the point that manual impls must handle all itself? maybe we should just change the ctor here to SubjectPropertyWriter as ISubjectSource is mostly meant as marker interface and eg own impls should impl their own writer? or too limiting?

src/Namotion.Interceptor.Connectors/Namotion.Interceptor.Connectors.csproj
```diff
@@ -11,6 +11,7 @@
 	</ItemGroup>
 
 	<ItemGroup>
+		<PackageReference Include="Microsoft.Extensions.DependencyInjection.Abstractions" Version="9.0.10" />
```
RicoSuter, line 14: why do we need that?

docs/superpowers/plans/2026-08-05-source-sync-state.md
```diff
@@ -0,0 +1,3571 @@
+# Source Synchronization State Implementation Plan
```
RicoSuter: remove

docs/connectors-source-monitoring.md
```diff
@@ -0,0 +1,275 @@
...
+
+```csharp
+await root.Kitchen.WaitForSynchronizationAsync(stoppingToken);
+```
+
+A source is in scope for an anchor when its root subject and the anchor lie on the same root-to-leaf path, in either direction: the source's root is an ancestor of (or is) the anchor, so it may claim into the awaited branch, or the source's root sits inside the anchor's own subtree. A source on a sibling branch is in neither set, so a source that never connects, or fails, on an unrelated branch never blocks a wait scoped to a branch it cannot claim into.
+
+Two edge cases in that scoping matter before you rely on a wait.
+
+If, once source registration is complete, the anchor's scope matches no source at all (a mistyped root, a device that was never configured), the wait never completes. It blocks until cancelled, and the only sign anything is wrong is a one-time warning logged the first time that empty scope is detected for that wait. If a wait you expected to complete instead sits forever, check the log for that warning before looking anywhere else.
```
RicoSuter: do we really need a warning? i'd drop that and the need for a logger factory etc. i mean after complete start we can expect that sources are correctly setup or the tree has no sources, no need to log a warning as we can assume all is correct.

docs/connectors-source-monitoring.md
```diff
@@ -0,0 +1,275 @@
...
+| `Unclaimed` | No source owns this property. |
+| `Connecting` | A source owns it, but that source hasn't completed its initial load (or has reconnected since it last did). |
+| `Synchronized` | The owning source completed its initial load. What that guarantees differs per protocol; see [What Synchronized Means per Protocol](#what-synchronized-means-per-protocol). |
+| `Stopped` | The owning source shut down; it will not restart. |
+
+`GetSourceState()` is only fully meaningful once the branch containing the property has been awaited through `WaitForSynchronizationAsync`. Before any claiming has happened, `Unclaimed` cannot be distinguished from "not yet claimed, but will be." After a claim it reports `Connecting`, so "will synchronize, still loading" is already distinguishable from "no source" even before the wait completes.
+
+## What Synchronized Means per Protocol
+
+OPC UA's `LoadInitialStateAsync` batch-reads every owned property from the session before returning, and WebSocket's applies the full-state message the server sends on connect. For both, reaching `Synchronized` means real values were confirmed from the external system.
```
RicoSuter: add link to opcua-client.md?

docs/connectors-source-monitoring.md
```diff
@@ -0,0 +1,275 @@
...
+
+MQTT's `LoadInitialStateAsync` always returns `null`. Retained messages arrive indistinguishably from any other message, through the normal subscription handler, and neither MQTT 3.1.1 nor 5.0 defines a signal that says "no more retained messages are coming" for a subscription. `SubjectSourceBase` therefore reaches `Synchronized` once the client's subscriptions are established, not once retained values have actually arrived. This is a protocol limitation, not something left unfinished: raising `DefaultQualityOfService` does not change it, since QoS governs delivery guarantees for messages that are sent, not whether the broker tells the client when a topic's retained backlog is exhausted. See [#418](https://github.com/RicoSuter/Namotion.Interceptor/issues/418) for an opt-in barrier that would wait for the first message (retained or live) per subscribed topic before declaring `Synchronized`.
```
RicoSuter: add link to mqtt md file?

docs/connectors-source-monitoring.md
```diff
@@ -0,0 +1,275 @@
...
+
+## Applications That Create Sources at Runtime
+
+`WithSourceMonitoring(builder.Services)` registers a `SourceRegistrationGate` hosted service that calls `CompleteSourceRegistration()` once `IHostApplicationLifetime.ApplicationStarted` fires. That fits an application where every source is a DI-registered hosted service, started as part of ordinary host startup.
+
+An application that creates sources dynamically, for example devices discovered after the host has already started, uses the parameterless `WithSourceMonitoring()` overload instead, and declares registration complete itself once it has started everything it intends to for that batch:
```
RicoSuter: maybe say that it's about dynamically directly during startup process but after DI complete. if it's for plug and play (way after start) this is not useful.

docs/connectors-source-monitoring.md
```diff
@@ -0,0 +1,275 @@
...
+| `SourceUnregistered` | `null` | A source unregistered, which happens when it is disposed. |
+| `StateChanged` | `null` | A source's own state changed. |
+| `PropertyClaimed` | set | A source took ownership of a property. |
+| `PropertyReleased` | set | A source gave up ownership of a property. |
+| `PropertyEnteredView` | set | An already-claimed property joined the tree when its subject attached; ownership did not change. |
+| `PropertyLeftView` | set | A still-claimed property left the tree when its subject detached; ownership did not change. |
+
+`SourceMonitor.Publish` enqueues every event onto each subscriber's own queue; each subscription drains its own queue on a single worker at a time, so a slow handler delays only that subscription. There is no ordering guarantee across subscriptions: two subscribers can observe events in different relative orders under concurrent activity.
+
+## Breaking Change for Custom ISubjectSource Implementers
```
RicoSuter: remove this, we dont need migration guide here

src/Namotion.Interceptor.Connectors/Monitoring/SourceMonitoringExtensions.cs
```diff
@@ -0,0 +1,107 @@
...
+using Namotion.Interceptor.Tracking;
+using Namotion.Interceptor.Tracking.Lifecycle;
+
+namespace Namotion.Interceptor.Connectors.Monitoring;
+
+/// <summary>Consumer-facing entry points for source monitoring.</summary>
+public static class SourceMonitoringExtensions
+{
+    private const string NoMonitorMessage =
+        "No SourceMonitor is reachable from this context. Call WithSourceMonitoring() on the tree root context.";
```
RicoSuter, line 14: add empty line?

src/Namotion.Interceptor.Connectors/Monitoring/SourceMonitoringExtensions.cs
```diff
@@ -0,0 +1,107 @@
...
+    public static SourceState GetSourceState(this PropertyReference property)
+    {
+        return property.TryGetSource(out var source) ? source.State : SourceState.Unclaimed;
+    }
+
+    /// <summary>
+    /// Resolves the single reachable monitor.
+    /// </summary>
+    /// <exception cref="InvalidOperationException">No monitor is reachable, or more than one is.</exception>
+    public static SourceMonitor GetSourceMonitor(this IInterceptorSubjectContext context)
```
RicoSuter, line 34: is this called somewhere inside the library or just public api?

src/Namotion.Interceptor.Tracking/IStartupCompletionDeferrer.cs
RicoSuter: Because `Hosting` and `Connectors` are siblings that do not reference each other, and `Tracking` is their common ancestor. Hosting needs to take the hold (it owns the queued start) and Connectors needs to grant it (it owns the wait), so the seam has to live somewhere both already depend on. Putting it in `Connectors.Monitoring` would mean Hosting referencing Connectors.

src/Namotion.Interceptor.Connectors/SubjectPropertyWriter.cs
RicoSuter: Overkill, keeping `int`. It increments once per connect cycle, and the check is equality rather than ordering, so even a wrap after ~2 billion reconnects would be harmless.

src/Namotion.Interceptor.Connectors/SubjectPropertyWriter.cs
RicoSuter: Agreed, and done in bdd0f95e. `SubjectPropertyWriter` now takes `SubjectSourceBase` rather than `ISubjectSource`, so the `as` cast is gone. It drives the source state transitions, which only the base class defines, so the interface parameter silently did nothing for a direct implementer. Nothing outside this repo constructs a writer, so widening it again stays open if a real case turns up. Eight tests moved off `Mock<ISubjectSource>` onto the real base as a result, which is closer to what they actually exercise.

src/Namotion.Interceptor.Connectors/Namotion.Interceptor.Connectors.csproj
RicoSuter: For the `IServiceCollection` parameter on the `WithSourceMonitoring(context, services)` overload, which registers the hosted service that completes registration at `ApplicationStarted`. `Hosting.Abstractions` was already referenced; this adds only the DI one.

docs/superpowers/plans/2026-08-05-source-sync-state.md
RicoSuter: Already done - the plan file is untracked and not part of the PR (`git ls-files docs/superpowers/` is empty). This thread is against an older revision.

docs/connectors-source-monitoring.md
RicoSuter: Agreed, dropped in bdd0f95e. It took a surprising amount with it: both warn-once latches, the all-`Stopped` warning, and the requirement that an `ILoggerFactory` be reachable from the context. `IsSatisfied` collapses to "registration complete, and every in-scope source is `Stopped` or `Synchronized`".

Your reasoning also removes a recurring failure mode rather than just some code: this warning managed to be silently dead in one configuration or another three separate times during review.

docs/connectors-source-monitoring.md
RicoSuter: Done in bdd0f95e - links to `connectors-opcua-client.md`.

docs/connectors-source-monitoring.md
RicoSuter: Done in bdd0f95e - links to `connectors-mqtt.md`.

docs/connectors-source-monitoring.md
RicoSuter: Done in bdd0f95e. It now reads "during startup but after the DI container is built", and the section states the limit explicitly: waits describe startup, not steady state, so a source created long after startup cannot re-block a wait that already completed. For those, follow the event stream instead.

docs/connectors-source-monitoring.md
RicoSuter: Already done - the migration chapter was deleted in 741d7795 following your earlier comment. This thread is against an older revision.

src/Namotion.Interceptor.Connectors/Monitoring/SourceMonitoringExtensions.cs
RicoSuter: Added.

src/Namotion.Interceptor.Connectors/Monitoring/SourceMonitoringExtensions.cs
RicoSuter: Not called inside the library - it is a consumer entry point, and the documented way to reach the stream (`context.GetSourceMonitor().Subscribe(...)`). Internally everything uses the plural `GetSourceMonitors()`, which returns all reachable monitors and does not throw, because a source must register with every one of them.

docs/connectors-monitoring.md
```diff
@@ -52,11 +52,11 @@ Two things about that scoping matter before you rely on a wait.
...
+A pending wait is not frozen to the sources that existed when registration completed. It is re-evaluated on every registration, unregistration and state change, so a source registering later can block it. Only a wait that has already completed is frozen, since a completed task cannot be un-completed. A scope matching no source, or one where every source has `Stopped`, completes immediately rather than blocking.
```
RicoSuter, line 55: question: if a source is added during wait, do we detect that and schedule another wait until it is successful? does this make sense or overkill?

RicoSuter: Yes, detected, and no second wait is needed: it is the same wait, re-evaluated live.

Satisfaction is not a snapshot taken when the wait was created. Every registration, unregistration and state change runs a pass over the pending waits, so a source that registers mid-wait and falls in scope simply becomes another condition that wait now has to satisfy. It stays pending until that source reaches `Synchronized` (or `Stopped`) too. Two tests pin it: `WhenASourceRegistersWhileAWaitIsPending_ThenTheWaitIsReEvaluated` and `WhenASourceIsRegisteredMidWait_ThenItIsIncludedAndReBlocksTheWait`.

The one thing that genuinely cannot happen is re-blocking a wait that has **already completed** — a finished `Task` cannot be un-finished. So the guarantee is "everything in scope at any point between the call and completion", not "everything that will ever exist".

That is the boundary between the two variants rather than a gap. Anything created before the gate opens is covered automatically. For a batch created after that, `DeferWaitCompletion()` is the mechanism: take the hold before creating the sources so waits cannot complete mid-batch, dispose it once they have registered. For sources coming and going long after startup, a wait is the wrong tool entirely and the event stream is the right one — which is why the docs now say plainly that waits describe startup, not steady state.

So: not overkill, and already the behaviour — but worth knowing it holds only while the wait is still pending.

src/Namotion.Interceptor.Connectors.Tests/SourceMonitorTests.cs
```diff
@@ -0,0 +1,681 @@
...
+    {
+        // Arrange
+        var context = CreateContext();
+        var source = new TestStateSource(new Person(context));
+        await source.StartAsync(CancellationToken.None);
+        await source.StopAsync(CancellationToken.None);
+        await AsyncTestHelpers.WaitUntilAsync(() => source.State == SourceState.Stopped);
+
+        // Act
+        await source.StartAsync(CancellationToken.None);
```
RicoSuter, line 375: is this good? shouldnt a source be restartable? or not really realistic as background services are not designed for that?

src/Namotion.Interceptor.Connectors.Tests/SourceMonitorTests.cs
```diff
@@ -0,0 +1,681 @@
...
+        await transition;
+
+        // Assert
+        await AsyncTestHelpers.WaitUntilAsync(() => received.Count >= 2);
+        Assert.Equal(SourceEventKind.SourceRegistered, received.First().Kind);
+        Assert.Equal(SourceEventKind.StateChanged, received.Skip(1).First().Kind);
+    }
+
+    [Fact]
+    public async Task WhenATransitionWasAlreadyInFlightWhenTheSourceUnregistered_ThenItIsNotPublished()
```
RicoSuter, line 505: will the unregister event be published as final even here?

src/Namotion.Interceptor.Connectors.Tests/SourceScopeTests.cs
```diff
@@ -0,0 +1,133 @@
...
+    }
+
+    [Fact(Timeout = 10_000)]
+    public async Task WhenTheParentGraphHasACycle_ThenIsInScopeReturnsPromptlyInsteadOfHanging()
+    {
+        // Arrange
+        // A same-tree reparent gives every node exactly one parent while still forming a cycle:
+        // nothing in ParentTrackingHandler rejects it. An unguarded single-parent walk that starts
+        // from a node in the cycle and never finds its (unrelated) candidate would ping-pong between
+        // the two nodes forever instead of terminating.
```
RicoSuter, line 108: so it uses some sort of visited list to break cycle and still answer correctly? it does a depth first full scan, ie checks all children and all parents in all directions, right?

src/Namotion.Interceptor.Connectors.Tests/SourceStateTests.cs
```diff
@@ -0,0 +1,334 @@
...
+        var person = new Person();
+        var source = new TestStateSource(person);
+        var writer = new SubjectPropertyWriter(source, NullLogger.Instance);
+        writer.StartBuffering();
+
+        // Act
+        await writer.LoadInitialStateAndResumeAsync(CancellationToken.None);
+
+        // Assert
+        Assert.Equal(SourceState.Synchronized, source.State);
```
RicoSuter, line 250: will it report synchronized directly after applying initial state or after replying the collected notifications while loading initial state (correct)?

src/Namotion.Interceptor.Connectors.Tests/SourceMonitorTests.cs
RicoSuter: Your second reading is the right one: it is not realistic, and more importantly it is not needed.

`BackgroundService.StartAsync` really would run `ExecuteAsync` again on a second call, against a fresh uncancelled token — the platform does not enforce anything. That is precisely why the guard is explicit rather than assumed. Without it, a "restarted" stopped source would claim, load and apply live values while `State` stayed `Stopped`, which is the worst of both worlds.

But the deeper reason is that restart is already handled one level down. A source does not need to be restarted to recover from a connection loss: the pump reconnects itself in a loop, calling `StartBuffering` and `LoadInitialStateAndResumeAsync` again and cycling `Connecting → Synchronized` each time. So "restart" as an external operation would only mean recovering from `Stopped`, and `Stopped` is reached exactly once, on shutdown or disposal.

Making it genuinely restartable would mean resetting `_state`, resetting the `_started` latch, re-registering with every monitor and re-claiming properties it never released — which is constructing a new instance, with extra steps and several new race windows. Creating a new instance is the same thing, honestly.

src/Namotion.Interceptor.Connectors.Tests/SourceMonitorTests.cs
RicoSuter: Yes. `SourceUnregistered` is still the last event a subscriber sees for that source.

`Unregister` removes the source, detaches the forwarder and publishes `SourceUnregistered`, all under the monitor lock. The event this test drops is a `StateChanged` that was already in flight — it had captured the handler list before the `-=` — so it would otherwise arrive *after* `SourceUnregistered` and hand a consumer a state for a source it has already removed.

Disposal keeps the useful ordering: `Dispose` transitions to `Stopped` first and unregisters second, so a subscriber sees `StateChanged(Stopped)` and then `SourceUnregistered`. That is pinned separately by `WhenASourceIsDisposedWithoutBeingStopped_ThenStoppedPrecedesUnregistered`.

src/Namotion.Interceptor.Connectors/Monitoring/SourceMonitor.cs
```diff
@@ -0,0 +1,434 @@
...
+/// maintains for the same lifecycle change, so it has to be up to date first.
+/// </remarks>
+[RunsAfter(typeof(ContextInheritanceHandler), typeof(ParentTrackingHandler))]
+public class SourceMonitor : ILifecycleHandler, IStartupCompletionDeferrer
+{
+    private readonly Lock _lock = new();
+    private Func<ILogger?>? _loggerResolver;
+
+    private ILogger? _logger;
+
```
RicoSuter, line 26: remove one line

src/Namotion.Interceptor.Connectors.Tests/SourceScopeTests.cs
RicoSuter: Visited list yes, depth-first yes — but not all directions, and that part is worth correcting.

Each walk goes **upward only**. `SearchGraph` pushes the start node, pops, compares against the target, then pushes `current.GetParents()`. It never descends into children. It is depth-first because `pending` is a `Stack`, and `visited` is a `HashSet` with reference equality, so a cycle terminates: the second visit to a node is skipped rather than re-expanded.

Both directions come from calling it twice with the arguments swapped, not from one omnidirectional scan:

```csharp
IsAncestorOrSelf(sourceRoot, anchor)   // walk up from the anchor, looking for the source's root
|| IsAncestorOrSelf(anchor, sourceRoot) // walk up from the source's root, looking for the anchor
```

The first answers "is the source rooted above me, so it can claim into my branch", the second "is the source rooted inside my branch". A source on a sibling branch satisfies neither, which is what stops an unrelated failing connection from blocking a wait.

The practical consequence of upward-only is that cost scales with ancestor depth rather than subtree size, so anchoring near a leaf of a wide tree is cheap.

src/Namotion.Interceptor.Connectors.Tests/SourceStateTests.cs
RicoSuter: Your parenthetical is right: after the replay, not directly after applying the initial state.

The order inside `LoadInitialStateAndResumeAsync`, all under the writer's lock:

1. re-check the generation, discard if a later `StartBuffering` superseded this cycle
2. `applyAction?.Invoke()` — apply the snapshot the source returned
3. replay every update buffered while the load was in flight
4. `_updates = null`, so subsequent writes apply directly
5. `TransitionTo(Synchronized)`

So `Synchronized` means the snapshot *and* everything that arrived during the load have both landed. Reporting it after step 2 would expose a window where the model is missing live changes that arrived mid-load while claiming to be in sync, which is the failure this buffer-load-replay ordering exists to avoid.

Step 5 sits inside the lock deliberately: it is the only place proven not-superseded. Reporting after releasing the lock would let a `StartBuffering` land between the check and the report, so a stale cycle could certify `Synchronized` for the whole duration of the reconnect that replaced it.

src/Namotion.Interceptor.Connectors/Monitoring/SourceRegistrationGate.cs
```diff
@@ -0,0 +1,24 @@
...
+internal sealed class SourceRegistrationGate(
+    IInterceptorSubjectContext context,
+    IHostApplicationLifetime lifetime) : IHostedService
+{
+    /// <inheritdoc />
+    public Task StartAsync(CancellationToken cancellationToken)
+    {
+        // Awaiting ApplicationStarted here would deadlock the host, because it fires only after
+        // every StartAsync has returned. Register a callback instead.
+        lifetime.ApplicationStarted.Register(context.CompleteSourceRegistration);
```
RicoSuter, line 18: does this callback work in any contexts, asp, wpf, blazor, or would it be better to do it differently eg in executeasny as this is when di is done?

src/Namotion.Interceptor.Connectors/SubjectSourceBase.cs
```diff
@@ -20,13 +22,114 @@ public abstract class SubjectSourceBase : BackgroundService, ISubjectSource
...
+        {
+            var ticks = Interlocked.Read(ref _lastSynchronizedTicks);
+            return ticks == 0 ? null : new DateTimeOffset(ticks, TimeSpan.Zero);
+        }
+    }
+
+    /// <inheritdoc />
+    public event EventHandler<SourceEvent>? StateChanged;
+
+    /// <summary>
```
RicoSuter: i think we should move these monitoring methods down in the class as it's not the interesting part of the source, eg after the main source code

src/Namotion.Interceptor.Connectors/SubjectSourceBase.cs
```diff
@@ -20,13 +22,114 @@ public abstract class SubjectSourceBase : BackgroundService, ISubjectSource
...
+    /// Moves to <paramref name="newState"/> and publishes the change, or does nothing when the
+    /// transition is a no-op or the source has already stopped.
+    /// </summary>
+    /// <remarks>
+    /// The state write, timestamp write and event raise are all inside one lock: a bare
+    /// compare-exchange is not enough, since a writer could set Synchronized, be preempted, let
+    /// disposal set Stopped and unregister, then resume and publish Synchronized after Stopped -
+    /// both compare-exchanges would have succeeded, so no stickiness rule could prevent it.
+    /// </remarks>
+    internal void TransitionTo(SourceState newState)
```
RicoSuter: maybe rename to TransitionStateTo

src/Namotion.Interceptor.Connectors/SubjectSourceBase.cs
```diff
@@ -73,50 +176,132 @@ public abstract ValueTask<WriteResult> WriteChangesAsync(
...
+        // Dispose can race this method: it may run (and find nothing yet in _registeredMonitors to
+        // unregister) between the guard above and the assignment/registration loop just completed.
+        // TransitionTo is monotonic and Stopped is terminal, so if State reads Stopped here, Dispose
+        // has already run - re-check and unwind whatever was just registered so a disposed source
+        // never stays registered forever.
+        //
+        // Unwinds through the LOCAL monitors array, never by re-reading the field: a Dispose that
+        // interleaved between the assignment and the loop above has already taken the field and left
+        // it empty, so re-reading here would strand this call's own registrations forever.
+        // Unregister is a no-op for an unregistered source, so a double unwind is harmless.
```
RicoSuter: compact these new docs?

src/Namotion.Interceptor.Connectors/SubjectSourceBase.cs
```diff
@@ -73,50 +176,132 @@ public abstract ValueTask<WriteResult> WriteChangesAsync(
...
+    {
+        try
+        {
+            while (!stoppingToken.IsCancellationRequested)
             {
-                _logger.LogError(ex, "Failed to listen for changes in source.");
                 try
                 {
-                    await Task.Delay(_retryTime, stoppingToken).ConfigureAwait(false);
+                    _propertyWriter.StartBuffering();
```
RicoSuter, line 161: this block here is same behavior as before, right?

src/Namotion.Interceptor.Connectors/Monitoring/SourceMonitor.cs
RicoSuter: Done in 551df75b.

src/Namotion.Interceptor.Connectors/SubjectSourceBase.cs
RicoSuter: Done in 551df75b. `SubjectSourceBase` now reads pump-first: constructor, the three hooks a connector implements (`StartListeningAsync`, `LoadInitialStateAsync`, `WriteChangesAsync`), `StartAsync`, `ExecuteAsync` and its write helpers, then the monitoring surface, then `Dispose`.

`State`, `LastSynchronizedAt`, `StateChanged`, `ReportConnectionLost` and the transition method moved together, since that is the block a connector author never needs to read.

Fields stayed where they were rather than moving with them: field initialization order is significant in C#, and there was no reason to put that at risk for a readability change.

src/Namotion.Interceptor.Connectors/SubjectSourceBase.cs
RicoSuter: Renamed to `TransitionStateTo` in 551df75b — it says what it transitions.

It is `internal`, so every call site is in-assembly or behind `InternalsVisibleTo` and the compiler verified all of them: `SubjectSourceBase` itself, `SubjectPropertyWriter`, and the `TestStateSource` double.

src/Namotion.Interceptor.Connectors/SubjectSourceBase.cs
RicoSuter: Compacted in 551df75b, down to the two things that are not obvious from the code: that seeing `Stopped` here means `Dispose` already ran, and that the unwind must go through the local array rather than the field, which `Dispose` has already emptied.

src/Namotion.Interceptor.Connectors/Monitoring/SourceRegistrationGate.cs
RicoSuter: It works wherever `IHostApplicationLifetime` exists, which is every Generic Host application: ASP.NET Core, Blazor Server, worker services, and WPF when hosted on the generic host. The one place it does not is **Blazor WebAssembly** — `WebAssemblyHostBuilder` provides no `IHostApplicationLifetime`, so `GetRequiredService<IHostApplicationLifetime>()` would throw there. That gap is currently undocumented; happy to add a line to the page or a clearer exception message if you want it covered.

On moving it into `ExecuteAsync`: that would be worse rather than better. `BackgroundService.StartAsync` invokes `ExecuteAsync` and returns at its first await, so the body runs *concurrently* with the other hosted services starting — there is no guarantee the sources have registered by then, which is exactly the defect the gate exists to prevent.

The property we need is "every `StartAsync` has returned", and `ApplicationStarted` is precisely that signal. Registering a callback rather than awaiting it is also deliberate: awaiting `ApplicationStarted` inside `StartAsync` deadlocks the host, since it only fires once every `StartAsync` has completed.

src/Namotion.Interceptor.Connectors/SubjectSourceBase.cs
RicoSuter: Yes, unchanged from master. The pump loop is the same sequence: `StartBuffering` → `StartListeningAsync` → `LoadInitialStateAndResumeAsync` → build the `ChangeQueueProcessor` → re-apply the retry queue → `ProcessAsync`, with the same catch that logs, transitions back to `Connecting` and waits out the retry delay.

What this branch added sits *inside* `StartBuffering` and `LoadInitialStateAndResumeAsync`, not here: those two now drive the state transitions, which is why no connector needed changes. The only edits to this method itself were removing a `TransitionStateTo(Connecting)` at the top that was provably a no-op (the field already initialises to `Connecting` and `StartAsync` refuses a stopped source), and the `finally` that sets `Stopped` on every exit path.

PR #355, opened 2026-07-06
fix: Capture and reconcile local writes across the full source (re)connect lifecycle [superseded by 420]

RicoSuter:
## Problem

Local writes made against a source-bound property while the source is not fully connected could be silently lost, and the two windows where this happens were reconciled inconsistently.

- **Connect window:** writes made while `StartListeningAsync` / `LoadInitialStateAsync` run.
- **Reconnect-delay window:** writes made during the `retryTime` delay of a hard re-init, when the previous connection's subscription had been disposed and no subscription existed. These were never captured, so on reconnect the initial-state load silently overwrote them. For write-only or non-read-back properties the model kept a value the source never received, a permanent divergence.

The two windows also disagreed: the same write to a source-unchanged property was restored if it landed in the reconnect-delay window (retry queue, old-value reconcile) but dropped if it landed in the connect window (positional supersede, new-value check). Same action, opposite outcome, decided only by timing.

Root cause of the original CI flake (`NullableGuidType_ShouldSyncBothDirections`): the change subscription was created per connection, so a write in these windows fell into a no-subscription gap and never reached the server.

## Fix

Make capture source-lifetime, and reconcile every write the source pump sees while not fully connected through one policy. "The pump" is the scope: the initial connection and its retries. Once a connection has succeeded the connectors handle session loss internally without re-entering the pump, so those reconnects keep flushing queued writes without the reconcile, as they did before. Unifying that path too is a follow-up.

1. `SubjectSourceBase` now owns a single `PropertyChangeQueueSubscription` for the whole source lifetime (created once, outside the retry loop). Capture never stops, so no window has a gap.
2. Each connection attempt synchronously drains owned writes into the bounded `WriteRetryQueue` twice, after the retry delay and after the initial-state load, then runs one reconcile, then a borrowed-subscription `ChangeQueueProcessor` for the connected phase. No background task and no new concurrency: draining, reconcile, and processing run sequentially on the pump task, so the subscription always has a single consumer.
3. The retry-queue reconcile is now 3-way. If the current model value equals the change's new value, send it to the source. If it equals the old value, restore it locally and the connected phase sends it. Otherwise the source diverged from the baseline the write was based on, so drop it (source wins). This subsumes both the connect-window and the reconnect-delay-window reconciliation.

Memory is bounded per connection attempt: each attempt drains the subscription into the bounded (drop-oldest) retry queue before it starts, so repeated failed attempts do not compound. Accumulation *within* one attempt is not bounded, see Known limitations.

## Scope and decisions

- The change lives in `SubjectSourceBase`, `ChangeQueueProcessor`, `WriteRetryQueue`, and `PropertyChangeQueueSubscription`, so the OPC UA, MQTT, and WebSocket source connectors all benefit.
- `ChangeQueueProcessor` gains an `internal` constructor overload that borrows a subscription without disposing it. The existing public constructor is unchanged, so servers are untouched and the public API is unchanged.
- The positional supersede added earlier in this branch is left in place. With the drains running before the connected phase, window writes never reach it, so it is inert on the source path. Keeping it avoids editing `ProcessAsync` and any change to servers. Removing the now-redundant supersede is a deferred cleanup.

## Known limitations

`docs/connectors.md` now scopes its Known Limitations chapter to the cases where the local model and the external system can end up disagreeing, or where the outcome depends on timing. Behavior that always converges moved to the sections that describe it. The list below follows the same split.

**Divergence and races**

- A dropped write leaves a lasting divergence on a property the source never reports back, because nothing later corrects it. Four situations drop one: retry queue overflow during a long outage, retry queue disabled (`writeRetryQueueSize: 0`), a write to a property not yet claimed on the first connection, and a reconcile that classifies the write as diverged because `Equals` is reference equality for arrays, collections and plain reference types.
- A write racing the initial-state load has a timing-dependent winner. Inherent to the local-first model.
- Connector-internal reconnects flush queued writes without the reconcile, so a queued write can overwrite a value the source changed during the outage. Pre-existing, and the reason the fix summary above is scoped to the pump.
- Ordering across different properties can invert: writes reconciled as already-current flush before restored writes travel through the change queue. Same-property ordering is preserved.

**Behavior changes worth release-noting** (these converge, so they are documented in the regular sections rather than as limitations)

- A context without a `PropertyChangeInterceptor` now fails startup with an actionable message instead of churning through connect and retry. A misconfigured deployment that currently runs will stop starting.
- Servers build their processor before the protocol server starts, so startup-window values the model has already moved past are collapsed rather than published in sequence. The settled value is identical; intermediate ones are no longer observed.
- Capture memory while disconnected is bounded per connection attempt, not within one. Between the two drains, which covers the retry delay and all of `StartListeningAsync` plus `LoadInitialStateAsync`, captured changes accumulate without a bound.

## Test plan

- New tests: a write during a failed connect attempt or the retry delay reaches the source on reconnect; the 3-way send branch delivers an already-current write; a borrowed subscription survives processor disposal; the non-blocking subscription drain.
- Existing connect-window capture tests and retry-queue re-apply tests still pass unchanged.
- Full solution builds clean (warnings as errors). Full unit suite green, 0 failures.

### Verification of the root cause (added on revival)

The connect-window loss was confirmed rather than inferred. Injecting a 3 second stall into the window between the readiness signals and the start of the processing loop makes the failure deterministic on the previous pump: the client reports connected at 0.3s, the test's write is raised as a `Local` change, and the value never reaches the server, with the property claimed, the session healthy, nothing pending and nothing logged. A later write to the same property goes through in about 100 ms. With this branch the same harness passes.

Under a loaded harness (2 visible cores plus saturated CPU, emulating a small CI runner) the OPC UA suite reproduced the loss on master and no longer reproduces it here.

The earlier note in this section claiming `NullableGuidType_ShouldSyncBothDirections` was "a local test-host load artifact, not a regression" was wrong and has been removed. It was a real defect, the one this branch fixes.

### Still open, tracked in #416

A second and unrelated defect survives this change: after a client connects, the first server to client subscription update is sometimes never delivered, permanently and with no log (`WriteAndReadArraysOnServer_ShouldUpdateClient`, `ArrayResize_GrowAndShrink_ServerToClient_ShouldSync`). The server side builds its processor before the server starts, so the ordering fixed here does not explain it. It still reproduces on this branch at roughly 1 loaded run in 8.


RicoSuter:
Note for this PR's design, recorded so it survives any rework: after #366 (typed ChangeOrigin) merges, locally computed write-backs produced during the connect and initial-load window become deliverable changes (hook transforms of inbound values and cascades publish as Local origin; a later PR adds Correction changes with old == new == stored). The change-origin design (docs/superpowers/specs/2026-07-12-change-origin-design.md) deliberately stays lifecycle-neutral and relies on this PR to close the load-window gap. The minimal contract this PR must keep, in whatever shape it ends up:

1. Capture across the whole window: the queue subscription exists for the source lifetime, so Local write-backs published during StartListeningAsync / LoadInitialStateAsync are never dropped for lack of a subscription.
2. Delivery after load: the drain plus reconcile sends a captured change whose new value still equals the current model value (this is what delivers transform write-backs and still-valid corrections).
3. Reconcile stays origin-agnostic on values: model equals new value sends, model diverged drops (source wins). This generically handles Correction changes without kind-specific code.
4. Own-source echoes (changes whose origin source is this source) are never re-sent through the drain.

If any of these change, the initial-load scenario in the change-origin spec needs a new owner. A useful regression test once both PRs are in: an inbound initial value transformed by an OnChanging hook reaches the source after load completes.

RicoSuter:
Folded into #420, which is now the single final state to review.

Everything here is carried over unchanged: the source-lifetime subscription covering both the connect and reconnect-delay windows, the drain into the write retry queue, the single reconcile point with source-wins polarity, per-property collapse of parked writes by revision, and the borrowed-subscription ChangeQueueProcessor. An independent review confirmed SubjectSourceBase, WriteRetryQueue, SubjectPropertyWriter and PropertyChangeQueueSubscription are byte-identical there.

The one deliberate change: `IsSuperseded` was removed because `CurrentValueFilter.IsCurrent` is the identical comparison, which is what let the connect-window rule and the steady-state filter become one mechanism. Verified equivalent by invoking both predicates on both branches, with one improvement: this version no longer throws on an unregistered property, which previously killed the processing loop.

The checklist in #420 (comment 5198208576) was worked through item by item there.

RicoSuter:
Reopening: closed prematurely, #420 is not merged yet. Will close once it is.

RicoSuter:
### Audit of the four-point contract from the note above, against #420 as it stands

| # | Contract | Status in #420 |
|---|---|---|
| 1 | Capture across the whole window, source-lifetime subscription | **held**. `SubjectSourceBase` creates it before the retry loop, so it covers `StartListeningAsync`, the load, and the retry delay |
| 2 | Delivery after load: send a captured change whose new value still equals the current model value | **held**. That is the send branch of the single reconcile point |
| 3 | Reconcile stays origin-agnostic on values: model equals new value sends, **model diverged drops (source wins)** | **deliberately reversed** |
| 4 | Own-source echoes are never re-sent through the drain | **held, with one added exception** |

**Item 3 is the one that changed, and the note above says a change here means the initial-load scenario in the change-origin spec needs a new owner.**

Divergence no longer drops. The reconcile now decides in three ways: a later local commit supersedes it, so drop; the model already holds it, so send; otherwise the load moved the model off it, so restore locally and let the connected phase send the re-applied write. Rationale is in #420's description under "Conflict policy: a deliberate reversal": source-wins resolved the ambiguity by discarding a write that had already committed locally, with no error and no way for the caller to know.

The reconcile stays origin-agnostic on values, which is the part of item 3 that Correction changes depend on. What flipped is only which side wins when the values differ, and the supersession check runs first, so a divergence reaching the restore branch means the load moved the value rather than a later commit. That should keep Corrections working, but it is the spec owner's call to confirm rather than mine.

**Item 4's exception:** a transaction confirmation on a property a connector has already written out is re-sent, because our own write can land on the source after the transaction's and leave it holding an older commit with nothing else to repair it.

**The regression test the note asked for exists**: `SubjectSourceBaseTests.WhenInboundInitialValueIsTransformedByHook_ThenTransformedValueReachesSource`, an inbound initial value transformed by an `OnChanging` hook reaching the source after load completes.

Issue #320, opened 2026-05-23
OPC UA loader: defer root-subject value mutations for atomic observer view


RicoSuter:
## Summary

Today the OPC UA `OpcUaSubjectLoader` (`src/Namotion.Interceptor.OpcUa/Client/OpcUaSubjectLoader.cs`) mutates the user's root subject progressively during the load pipeline. Collections appear with empty children that fill in later, dynamic properties are added one-by-one interleaved with IO, subject-reference attachments happen across hundreds of milliseconds of browse/read round-trips.

For UIs bound to the root subject (Blazor, MAUI, anything reactive) this causes visible flicker: properties pop in, collections render empty then populate. For `IObservable` and `ILifecycleHandler` consumers it produces a long stream of small events instead of one cohesive "load committed" signal.

## Goal

Make value-bearing mutations targeting the root subject (collection / dictionary / subject-reference attachments) appear to root observers as a single synchronous burst at the end of `LoadSubjectAsync`, rather than progressively during IO.

**Out of scope:** structural atomicity for `DynamicSubject` roots (where dynamic properties are added during load). That requires a heavier "staging subject" design and should be tracked separately if needed.

## Background

This was explored in detail during the review of #313 (the loader batched-IO rewrite). A draft design document proposed deferring both `AddProperty` and `SetValueFromSource` calls on root via a `RootMutation` queue drained synchronously at the end of `LoadSubjectAsync`.

A `superpowers:code-reviewer` review of that draft surfaced a critical defect: short-circuiting `TryCreateDynamicProperty` to defer `AddProperty` for root would skip the entire downstream classification pipeline (`OpcUaSubjectLoader.cs:251-252` does `if (property is null) continue`), breaking subtree loading for dynamic root children. The `OpcUaDynamicServerClientTests.WhenConnectingToServer_ThenStaticallyMissingPropertiesAreDynamicallyAdded` snapshot would render nearly empty.

The reviewer also corrected a mis-characterization: `AddFallbackContext` actively fires `AttachToContext` events during discovery (verified in `src/Namotion.Interceptor/Interceptors/InterceptorExecutor.cs:58-72`), so `IHostedService.StartAsync` and per-subject `SubjectAttached` events run **during discovery**, not at the apply phase. Atomicity is only achievable for `root.Properties` enumeration changes and value-changed events on root.

## Proposed approach (revised)

[...]

PR #357, opened 2026-07-07
OPC UA four-phase loader: discover, validate, commit

RicoSuter:
## Summary

Rewrites the OPC UA client subject loader from the master base as an explicit four-phase model (discover, validate, commit, subscribe), validated against PR #313's frozen acceptance suite.

This is the third PR in a stack. The shared protocol layer and the subscription-setup fix were split out so this PR contains only the loader rewrite:

1. #397 batched browse/read session primitives and the split status classifier (base: `master`)
2. #398 callback gating and sweep-before-read-after-write (base: #397)
3. **this PR** the four-phase loader (base: #398)

### The four phases

- **Discover** builds an in-memory `OpcUaLoadPlan` and defers ownership claims, root assignments, and values into it. Subjects created during discovery stay off-tree: they are never linked into the parent context until commit, so a planning failure leaks nothing into the registry and there is no rollback path to get wrong. Discovery still needs a `RegisteredSubject` to descend through, so an off-tree subject gets a detached one. One live side effect remains because nested dynamic discovery needs it: dynamic properties are added eagerly to their subjects. On a reused subject that is visible immediately, but it is transient and re-matched by the next load through the attached node-id attribute.
- **Validate** deduplicates and resolves deterministic address-space conflicts (duplicate names, targets, keys, with a smaller-node-id tie-break) as the plan is built.
- **Commit** applies the plan in one ordered phase: ownership is claimed and node-id metadata stamped before subjects are assigned onto root. On mid-commit failure it releases only this-commit claims and does not restore root values, matching PR #313's boundary.
- **Subscribe** is in #398.

### Relative to PR #313

Comparison is in `docs/design/opcua-loader-comparison.md`: comparable loader-core size (1,239 vs 1,298 lines), zero churn outside the OPC UA client internals (PR #313 touched 21 files outside), clearer phase boundaries.

Two config bounds are added (`MaxBrowseContinuations`, `MaxAttributeTraversals`).

## Test Plan

- [x] Frozen PR #313 acceptance suite passes unchanged (one internal-commit test re-pointed to `OpcUaLoadPlan`, assertions preserved)
- [x] OPC UA test project including integration tests: 361 passed
- [x] Solution-wide non-integration tests pass
- [x] Public API snapshot (`VerifyChecksTests.PublicApi`) passes unchanged


RicoSuter:
## Browse round-trip measurement across the loader variants

Measured rather than estimated, because the earlier discussion (including my own numbers) was arithmetic from reading the code and was wrong twice.

**Harness.** Temporary counter around `session.BrowseAsync` / `BrowseNextAsync`, identical tree on every branch, run against the in-process test server (`OpcUaTestServer`): 241 subjects (40 machines x 5 sensors), roughly 1,000 value properties. Server advertises `MaxNodesPerBrowse = 4000` and `MaxBrowseContinuationPoints = 100`.

| variant | browse round-trips | nodes submitted | wall clock |
|---|---|---|---|
| `master` | **1,083** | 1,083 | 523 ms |
| #313 | **9** | 1,084 | 321 ms |
| this PR (#357) | **16** | 1,084 | 321 ms |
| this PR with #397's continuation-point cap disabled | **9** | 1,084 | 333 ms |

Wall clock is not a useful signal here: the test server is on loopback, so round-trip time is close to zero and the numbers are dominated by connection and subscription setup. At a realistic 5 ms RTT, 1,083 round-trips is about 5.4 s of pure browse latency against about 45 ms; at 20 ms WAN it is about 22 s against 180 ms.

### Three conclusions

**1. The IO win belongs to #313, not to this PR.** 1,083 to 9 is a 120x reduction and it is already done there. This PR ports the same phases (see the "mirrors PR 313's LoadSubjectsAsync phases" and "ported from PR 313's OpcUaAttributeLoader" comments) and submits an identical node count.

**2. This PR is a 78% round-trip regression against #313**, 16 against 9, caused by the continuation-point cap introduced in #397. #313 has no such cap and batches at `MaxNodesPerBrowse` (4000); #397 caps at `MaxBrowseContinuationPoints` (100). Disabling the cap returns this PR to exactly 9, which confirms the mechanism. The widest phase is attribute discovery at roughly 1,000 parents, so it splits into 10 calls instead of 1.

The cap is not wrong in principle. `BadNoContinuationPoints` is not in `LoadSkipCodes`, so it throws transient and the retry would use the same batch size, which is the non-convergent loop the cap exists to prevent. But with the shipped default `MaxReferencesPerNode = 0` a stock SDK server never issues a continuation point at all (`CustomNodeManager` only creates one when `MaxResultsToReturn != 0`), so on the default configuration the cap protects a quota that cannot be consumed. It should apply only when `MaxReferencesPerNode != 0`, with `BadNoContinuationPoints` handled adaptively as a fallback for servers that paginate on their own initiative.

**3. The four-phase atomicity layer buys no performance.** It is entirely a correctness-model cost, and the correctness model it buys is not reachable with the current registry APIs.

### Why the model is not reachable

Discovery has to descend through a `RegisteredSubject`. For an off-tree subject there is no real one, so `GetRegisteredView` fabricates `new RegisteredSubject(subject)`. That fabricated view resolves `Properties` correctly, because those derive from the subject's own metadata, but its `Children` is permanently empty: `Children` is only ever written by `SubjectRegistry.HandleLifecycleChange` via `AddChild`, which has exactly one caller.

Every child-reuse path in the planner reads `.Children`, at `OpcUaLoadPlanner.cs:427`, `:462` and `:633`, so each one silently mis-plans against a staged subject. The `:462` path logs "the parent type must instantiate this property in its constructor" and skips, which points at the wrong cause.

Making this correct means teaching the registry to register a subtree detached and splice it in later. That is a core-library feature with a second writer into a structure that has one today, and it is bought to satisfy per-load atomicity, which is a stronger guarantee than the quiescent consistency the project actually targets.

Note also that #320 already scoped this out: "structural atomicity for DynamicSubject roots ... requires a heavier staging subject design and should be tracked separately if needed", and prescribed "defer only value-bearing root mutations and the source-claim burst". #313's `OpcUaLoadContext.QueueOrApplySetValue` already implements exactly that.

### Where this leaves the stack

#313's model is the one to carry forward: subjects attach eagerly during discovery, so `.Children` works, and only root value mutations are deferred to the end. The plan is to merge #398 and #397 (with the cap made conditional), close this PR, and refactor #313 down to that model.

RicoSuter:
Closing in favour of #414.

The measurements in [the comment above](https://github.com/RicoSuter/Namotion.Interceptor/pull/357#issuecomment-5175448321) settled it: the four-phase staged loader added zero IO improvement over #313 and cost 7 extra browse round-trips, while its per-load structural atomicity is not reachable with the current registry APIs. `RegisteredSubject.Children` is written only by `SubjectRegistry.HandleLifecycleChange`, so off-tree discovery silently loses subject reuse.

Nothing was lost by closing this. The test suites of this branch and #313 were roughly 90% byte-identical, and the pieces that were genuinely this branch's contribution are all in #414: the browse-completion contract, the batched type resolver, `OpcUaBrowseName`, and the omit-on-partial-pagination expectations in `OpcUaSubjectLoaderBatchingTests`.

#414 verifies at 9 browse round-trips for the same 241-subject address space this branch took 16 on.

Issue #342, opened 2026-06-09
Define the source consistency contract and built-in divergence reconciliation


RicoSuter:
## Summary

Follow-up to #338 and #340, stepping back from individual fixes. The goal is that the local model never ends up persistently out of sync with its sources: temporary inconsistency windows are unavoidable in a distributed setup, but every divergence should be detected, reported, and repaired without waiting for a reconnect.

This is a design task, not an implementation task. The outcome is a design document under `docs/design/` that defines the consistency contract the library promises and the reconciliation mechanisms needed to honor it, plus follow-up implementation issues.

## Where divergence can enter today

| # | Divergence source | Tracked in | State today |
|---|---|---|---|
| 1 | Transaction commit fails and a revert also fails | #340 (designed, PR #349) | Repair designed: a source-wins resync requested in the commit failure path (on a failed revert, an indeterminate write, or a commit timeout), reported in-band via `SubjectTransactionException`. PR #349 ships only the resync primitive; pending-divergence tracking and events are added by this issue (question 3). Diverged and unrepaired until that implementation lands |
| 2 | Non-transactional write or inbound source apply lands inside the commit window | #338 | Silently overwritten, documented best-effort |
| 3 | Non-transactional outbound write is undeliverable | none | `WriteRetryQueue` retries while connected, but queue overflow silently drops the oldest writes (ring buffer), `writeRetryQueueSize: 0` drops immediately, and a write the source permanently rejects retries forever. Local stays ahead of the source until the next reconnect snap-back |
| 4 | Inbound source value fails to apply locally (validation or setter throws) | none | Logged and dropped by design (docs/connectors.md, Inbound Update Error Handling). Local stays behind the source until the source pushes that property again |
| 5 | No reconciliation while connected | none | The source-wins snapshot plus optimistic re-apply runs only at reconnect (`SubjectSourceBase` buffer-load-replay). Divergence from rows 1 to 4 can persist for the whole connected lifetime of a source |
| 6 | Cascade writes from `OnChanging`/`OnChanged` triggered by a from-source apply inherit the source scope and are never pushed back (derived recalculations are unaffected: they always publish with a null source and still flow) | #345 | A cascade triggered by an inbound source value that writes another source-owned property leaves the source without the cascade value. After #343, transactional commit applies deliberately adopt the same scope semantics, so this applies uniformly to both paths |

[...]

PR #366, opened 2026-07-11
Typed ChangeOrigin with one-shot source stamping

### LINKED ISSUE #345: Publish hook cascade writes with a null source, like derived recalculations

### LINKED ISSUE #365: Typed ChangeOrigin discriminator with causal TriggeredBy for change notifications
## Summary

`SubjectPropertyChange.Source` is an `object?` whose documented meaning is "the system that holds or confirmed this value". After #343 and #345 the rule is truthful but coarse: the different confidence levels (sent inbound versus confirmed by a commit) are indistinguishable, and local-origin writes erase which source triggered them. Replace the raw reference with a typed discriminator, for example `ChangeOrigin { Kind: Local | FromSource | Confirmed, Source, TriggeredBy }` with `Presumed` reserved for later (see the #342 design, question 4 there).

Breaking changes are acceptable at this stage.

## Consumers

1. **Provenance-aware validation.** Validators that strictly validate local user input but treat source values as authoritative. Today a production consumer reads `SubjectChangeContext.Current.Source is ISubjectSource` inside a validator, which reaches into thread-static state and cannot distinguish inbound from commit-confirmed origins. The discriminator should flow into the validator signature. Timing note: validation runs at transaction capture under local context and again at commit replay under the confirmed source scope, so a provenance-aware validator naturally gates user input at capture and skips re-rejection of source-confirmed values at replay.
2. **Causal provenance for framework-invoked callbacks.** After #345 (PR #348), hook cascades and INPC write-backs publish as local origin, and hooks are fully origin-blind, including reads: the ambient context is a single slot that serves both write stamping and trigger observation, so clearing it for write correctness also removes read visibility. `TriggeredBy` restores read-only trigger visibility as data, without restoring authority inheritance for writes.
3. **Reconciliation prioritization.** A `Presumed` kind becomes necessary only if the read-back optimization deferred in the #340/#349 design is adopted. The shape should allow adding kinds without breaking consumers.

## Design notes (from the PR #348 review discussion)

- The generated callback scopes are the designed swap point: `SubjectChangeContext.WithLocalOrigin()` encodes local origin as `Source = null` today and sets `Kind = Local` in the future without changing its signature or any call site.
- A two-slot ambient context (`Source` for write stamping plus a preserved `TriggerSource` for observation) was considered for PR #348 and rejected: two source-like properties on the same struct are easy to confuse, and the design invites `WithSource(Current.TriggerSource)` write-backs that reintroduce inherited authority by hand. `TriggeredBy` as read-only causal metadata on a typed origin avoids both problems.
- Timestamps stay separate from origin: local-origin scopes preserve the ambient changed and received timestamps (time is data describing the logical event) while clearing the source (authority over the value). `ChangeOrigin` models authority and causality, not time.
- Allocation budget: `SubjectPropertyChange` flows on the hottest path, so the origin must be allocation-free per change (readonly struct, cached nodes for common cases). A class allocated per change is unacceptable; benchmark before and after.
- `TriggeredBy` representation: a struct cannot nest a field of its own type, so the options are a flattened trigger pair (trigger kind plus trigger source), a reference to a cached immutable origin node, or restricting the causal chain to one level. Related open questions: how deep does causality go (a cascade of a cascade of a derived recalculation), and does retaining trigger references keep source objects alive longer than intended?
- Echo suppression identity: the outbound queue currently skips changes via `ReferenceEquals(change.Source, targetSource)`. With the discriminator this becomes a rule over `Kind` plus source identity and must be specified explicitly: a `Confirmed` value for source A is an echo for A but a normal outbound write for any other bound source.

## Relationship to other work

- The final set of kinds is informed by the #342 design (tracker resolution kinds; `Presumed` only with read-back). A minimal `Local | FromSource | Confirmed` shape with `TriggeredBy` is implementable independently once #345 (PR #348) is merged; adding kinds later is additive.
- Swap points: `SubjectChangeContext` scopes, `SubjectPropertyChange`, the `ChangeQueueProcessor` echo skip, `SetValueFromSource`, the transaction commit apply, validator signatures, and the docs tables in `docs/connectors.md`.

RicoSuter:
Replaces the ambient, dynamic-extent source scope with a one-shot per-write origin stamp and replaces `SubjectPropertyChange.Source` (`object?`) with a typed `ChangeOrigin` discriminator (`Local | FromSource | Confirmed`). A change carries a source only when its stored value is exactly the value that source sent or confirmed. Every other write, including hook cascades, `INotifyPropertyChanged` handler write-backs, and derived recalculations, is local origin by default, structurally rather than by convention.

Supersedes #348: instead of counter-scopes (`WithLocalOrigin`) at every framework callback plus a viral contract on manual `IRaisePropertyChanged` implementations, the mechanism makes local-by-default fall out of the design. The generator is untouched.

Design doc: `docs/superpowers/specs/2026-07-12-change-origin-design.md`
Implementation plan: `docs/superpowers/plans/2026-07-12-change-origin.md`

## Root cause

The ambient scope existed only to smuggle `(source, timestamps)` past the fixed property setter signature, but its dynamic extent made every downstream write (hooks, INPC handlers, derived recalculations) silently inherit a source stamp it never earned, which the outbound echo suppression then wrongly suppressed. Source authority applies to exactly one write; timestamps are the only genuinely dynamic-extent concern. The design splits them accordingly.

## Design highlights

- `ChangeOrigin` (readonly struct, byte-backed kind) with `default == Local`; stored flattened in `SubjectPropertyChange` so the struct does not grow.
- Internal `PendingOrigin` stamp: armed per write by `SetValueFromSource`, consumed at `PropertyWriteContext` construction only when the target property matches, restored (not cleared) on dispose so nested stamped writes compose.
- `context.Origin` is finalized at the terminal write: a stamped origin whose final value differs from the sent value becomes `Local`, so hook transforms and interceptor rewrites publish as locally computed and flow back to the source.
- Timestamps unchanged: `WithChangedTimestamp` stays, `WithTimestamps(changed, received)` replaces the timestamp half of the removed `WithState`.
- `ApplySubjectUpdate` takes a required typed `ChangeOrigin`; WebSocket applies migrate off the scope.
- Validators receive origin via `PropertyValidationContext<TProperty>`: capture-time validation sees `Local`, commit replay sees `Confirmed`.
- The echo skip in `ChangeQueueProcessor` stays a single reference comparison, behavior-identical.

## Scenarios fixed

1. Hook cascade writes during an inbound apply were echo-suppressed and never reached the source.
2. INPC handler write-backs inherited the ambient source.
3. Transformed values (`OnXChanging` reassigns `ref newValue`) now publish as locally computed so corrections flow back.
4. Manual `IRaisePropertyChanged` implementations no longer need a documentation-enforced local-origin contract.
5. A write interceptor rewriting `NewValue` during a source-scoped write no longer keeps the source stamp (previously a documented hole).
6. Lifecycle handlers that write no longer inherit ambient authority.
7. Validators no longer peek at thread-static state and can distinguish inbound from commit-confirmed origins.

The equality-suppressed correction case (model at 100, source sends 105, hook clamps to 100, source stays diverged) is scenario 8 and lands in the follow-up PR introducing the `Correction` kind, which closes #365. `TriggeredBy` and `Presumed` remain tracked in #342 as additive extensions.

## Breaking changes (0.0.x)

- `SubjectPropertyChange.Source` becomes `Origin` (`ChangeOrigin`); `WithSource` becomes `WithOrigin`.
- `SubjectChangeContext` loses `Source`, `WithSource`, and `WithState`; `WithTimestamps` added.
- `IPropertyValidator.Validate` takes `PropertyValidationContext<TProperty>`.
- `ApplySubjectUpdate` gains a required `ChangeOrigin origin` parameter.
- `SetValueFromSource` keeps its signature; all OPC UA and MQTT call sites are unchanged. One deliberate behavior nuance: passing `receivedTimestamp: null` now preserves an ambient received timestamp (matching `WithChangedTimestamp` semantics) where the removed `WithState` path cleared it, and there is intentionally no API left to explicitly clear an ambient received timestamp within a scope. XML-documented and test-pinned.

## Benchmarks

Write-path benchmark results (before/after) will be posted on this PR before merge, per the plan's Task 10.

Closes #345


src/Namotion.Interceptor/ChangeOrigin.cs
```diff
@@ -0,0 +1,41 @@
+namespace Namotion.Interceptor;
+
+/// <summary>
+/// The kind of origin of a property change. Byte-backed deliberately: inside
+/// <c>SubjectPropertyChange</c> the runtime can fold a byte into padding where a wider
+/// enum could grow the struct. Do not widen.
+/// </summary>
+public enum ChangeOriginKind : byte
+{
+    Local = 0,
```
RicoSuter, line 14: Add xml docs explaining each mber

RicoSuter: Done: added member-level XML docs for Local, FromSource, and Confirmed in 07396b85.

src/Namotion.Interceptor/PendingOrigin.cs
```diff
@@ -0,0 +1,79 @@
...
+/// intent-level APIs (SetValueFromSource, ApplySubjectUpdate, transaction replay).
+/// </summary>
+internal static class PendingOrigin
+{
+    [ThreadStatic] private static bool _armed;
+    [ThreadStatic] private static PropertyReference _target;
+    [ThreadStatic] private static ChangeOrigin _origin;
+    [ThreadStatic] private static object? _sentValue;
+
+    internal static PendingOriginScope Set(PropertyReference target, ChangeOrigin origin, object? sentValue)
```
RicoSuter, line 37: Do we need this scope when its reset by first read anyway?

src/Namotion.Interceptor/PendingOrigin.cs
```diff
@@ -0,0 +1,79 @@
...
+/// (a zero-allocation stack through nested ref structs, like SubjectChangeContextScope),
+/// so a cancelled write cannot leak the stamp and a nested stamped write cannot destroy
+/// an outer stamp. Same-property re-entry from OnChanging is unsupported (the inner
+/// invocation consumes the stamp). Thread-static by design: set and consume happen
+/// synchronously within one call frame, never across await. Internal: producers use
+/// intent-level APIs (SetValueFromSource, ApplySubjectUpdate, transaction replay).
+/// </summary>
+internal static class PendingOrigin
+{
+    [ThreadStatic] private static bool _armed;
```
RicoSuter: Wouldnt it be faster to put them in a single stread static struct field to not read multiple thread static fields?

src/Namotion.Interceptor/PendingOrigin.cs
```diff
@@ -0,0 +1,79 @@
...
+    [ThreadStatic] private static PropertyReference _target;
```
RicoSuter: Are the fields always reset to avoid memleaks (source and sub in prop)?

src/Namotion.Interceptor/PendingOrigin.cs
RicoSuter: Yes, the scope is still needed: TryConsume only covers the consumed path. The restore-on-dispose handles the three no-read paths: a cancelled OnXChanging (the chain never starts, so nothing consumes and the stamp would leak into the next write), a target mismatch that deliberately leaves the slot armed, and nested arming where the inner scope must restore the outer frame rather than clear it. All Set call sites dispose via using, so the slot is always returned to its previous (ultimately empty) frame.

src/Namotion.Interceptor/PendingOrigin.cs
RicoSuter: Yes: TryConsume resets all fields on a successful match, and every Set site disposes its scope via using, which restores the previous frame; the outermost restore is the empty frame, so the PropertyReference (subject) and source references never outlive the write call on that thread. The single-struct-field change from your other comment will make the reset one assignment, which makes this easier to see.

src/Namotion.Interceptor/PendingOrigin.cs
RicoSuter: Good call: refactored to a single [ThreadStatic] PendingFrame struct field in d2227c57. Set/TryConsume/Restore now do one TLS slot access plus field offsets instead of up to four slot lookups, and the reset is a single default assignment (which also makes the no-leftover-references property from the other thread trivially true). 5/5 stamp tests and 72/72 project tests green.

src/Namotion.Interceptor.Connectors.Tests/Updates/SubjectUpdateExtensionsTests.cs
```diff
@@ -476,14 +476,17 @@ public async Task WhenApplyingFromSource_ThenSourceIsTracked()
...
-        using (SubjectChangeContext.WithSource(externalSource))
-        {
-            target.ApplySubjectUpdate(update, DefaultSubjectFactory.Instance);
-        }
+        target.ApplySubjectUpdate(update, DefaultSubjectFactory.Instance);
+
+        // Applying an update no longer stamps a source on its writes (the update applier threads the
+        // origin per write in a follow-up step); assert source tracking through the supported per-write
+        // primitive, which is the mechanism the applier will use.
+        new PropertyReference(target, nameof(Person.FirstName))
```
RicoSuter: shouldnt ApplySubjectUpdate take source directly here and apply with this source and calling SetValueFromSource here is then not needed?

src/Namotion.Interceptor.Connectors.Tests/SubjectSourceBaseTests.cs
```diff
@@ -737,10 +737,7 @@ public async Task WhenRetryQueueDisabled_ThenInitializationSucceedsWithoutReappl
         {
             LoadInitialStateOverride = _ => Task.FromResult<Action?>(() =>
             {
-                using (SubjectChangeContext.WithSource(source!))
```
RicoSuter, line 740: can we just remove that?

src/Namotion.Interceptor.Connectors/Transactions/SourceTransactionWriter.cs
```diff
@@ -303,7 +303,7 @@ private static void MarkSnapshotSlots(Span<SubjectPropertyChange> snapshot, int[
         for (var i = 0; i < count; i++)
         {
             var slot = indices[i];
-            snapshot[slot] = snapshot[slot].WithSource(source);
+            snapshot[slot] = snapshot[slot].WithOrigin(ChangeOrigin.Confirmed(source));
```
RicoSuter, line 306: question: Why do we need to differentiate between FromSource and Confirmed? Is there a use case for that? (we keep it just wondering)

src/Namotion.Interceptor.Tests/SubjectChangeContextTests.cs
```diff
@@ -40,46 +40,52 @@ public void ResolveChangedTimestamp_WithNullTimestamp_ReturnsZero()
...
     {
-        Assert.Null(SubjectChangeContext.Current.Source);
+        // Arrange
+        var received = new DateTimeOffset(2025, 6, 15, 12, 0, 1, TimeSpan.Zero);
+        var changed = new DateTimeOffset(2025, 6, 15, 12, 0, 0, TimeSpan.Zero);
+
+        // Act & Assert - a null received argument preserves the ambient received timestamp
+        using (SubjectChangeContext.WithTimestamps(changed, received))
+        {
+            using (SubjectChangeContext.WithTimestamps(changed, null))
```
RicoSuter, line 67: question: is this new behavior (null does not overwrite with null ts), or just an additional test?

src/Namotion.Interceptor.Tests/VerifyChecksTests.PublicApi.verified.txt
```diff
@@ -133,13 +147,11 @@ namespace Namotion.Interceptor
     }
     public readonly struct SubjectChangeContext
     {
-        public readonly object? Source;
         public System.DateTimeOffset? ReceivedTimestamp { get; }
         public static Namotion.Interceptor.SubjectChangeContext Current { get; }
         public static System.Func<System.DateTimeOffset> GetTimestampFunction { get; set; }
         public static Namotion.Interceptor.SubjectChangeContext.SubjectChangeContextScope WithChangedTimestamp(System.DateTimeOffset? changed) { }
```
RicoSuter, line 153: should we remove WithChangedTimestamp and just provide WithTimestamps, and WithTimestamps(foo, null) is the same as WithChangedTimestamp(foo)? dont apply directly needs discussion

src/Namotion.Interceptor.Tracking/Transactions/SubjectPropertyChangeOperations.cs
```diff
@@ -118,7 +118,10 @@ private static bool TryApplyLocalChange(this SubjectPropertyChange change, out E
         try
         {
             var metadata = change.Property.Metadata;
-            using (SubjectChangeContext.WithState(change.Source, change.ChangedTimestamp, change.ReceivedTimestamp))
+            // Setting a Local origin is a no-op stamp (consume yields Local), so revert paths
+            // re-apply each change with its original provenance without branching.
+            using (SubjectChangeContext.WithTimestamps(change.ChangedTimestamp, change.ReceivedTimestamp))
+            using (PendingOrigin.Set(change.Property, change.Origin, change.GetNewValue<object?>()))
```
RicoSuter: perf: should we avoid reading `change.GetNewValue<object?>` twice here? use var?

src/Namotion.Interceptor.Tracking/Transactions/SubjectTransactionInterceptor.cs
```diff
@@ -69,11 +69,13 @@ public void WriteProperty<TProperty>(ref PropertyWriteContext<TProperty> context
                     $"Cannot modify property '{context.Property.Metadata.Name}': Transaction is bound to a different context.");
             }
 
+            // The ambient context is read only for the received timestamp; the origin comes from
```
RicoSuter: remove comment and inline SubjectChangeContext.Current or do we need currentContext somewhere else in this scope?

src/Namotion.Interceptor.WebSocket/Client/WebSocketSubjectClientSource.cs
```diff
@@ -293,10 +293,9 @@ private async Task ConnectCoreAsync(CancellationToken cancellationToken)
         return Task.FromResult<Action?>(() =>
         {
             var factory = _configuration.SubjectFactory ?? DefaultSubjectFactory.Instance;
-            using (SubjectChangeContext.WithSource(this))
-            {
-                _subject.ApplySubjectUpdate(_initialState, factory);
-            }
+            // TODO: Inbound updates are applied without a source origin until the update applier
```
RicoSuter: will this TODO be fixed/resolved in this pr later?

src/Namotion.Interceptor.WebSocket/Client/WebSocketSubjectClientSource.cs
```diff
@@ -534,13 +533,12 @@ private void HandleUpdate(SubjectUpdate update)
...
         propertyWriter.Write(
-            (update, subject: _subject, source: this, factory: _configuration.SubjectFactory ?? DefaultSubjectFactory.Instance),
+            (update, subject: _subject, factory: _configuration.SubjectFactory ?? DefaultSubjectFactory.Instance),
             static state =>
             {
-                using (SubjectChangeContext.WithSource(state.source))
-                {
-                    state.subject.ApplySubjectUpdate(state.update, state.factory);
-                }
+                // TODO: Inbound updates are applied without a source origin until the update applier
```
RicoSuter: will this TODO be fixed/resolved in this pr later?

docs/tracking.md
```diff
@@ -492,27 +492,18 @@ This is primarily used internally by the derived property change detection syste
...
 propertyReference.SetValueFromSource(
     source: mqttSource,
     changedTimestamp: DateTimeOffset.Now,
     receivedTimestamp: DateTimeOffset.Now,
     valueFromSource: newValue);
+// change.Origin is ChangeOrigin.FromSource(mqttSource)
 ```
 
-This prevents feedback loops where changes from external sources are written back to those same sources.
+Source marking is per write; a scope-based source no longer exists. This prevents feedback loops where changes from external sources are written back to those same sources.
```
RicoSuter: dont document past api, just current, also check other docs and xml docs that we only describe current state and not how it was at some point in the past.

src/Namotion.Interceptor.Connectors/Transactions/SourceTransactionWriter.cs
RicoSuter: Three concrete consumers of the distinction, all data-level (the echo skip indeed treats them identically today): (1) provenance-aware validators: capture-time validation sees Local, commit replay sees Confirmed, so a strict validator gates user input at capture without re-rejecting source-confirmed values at replay, and it can also distinguish merely-received (FromSource) from acknowledged (Confirmed) values; (2) the follow-up Correction PR (#365) synthesizes corrections only for FromSource writes, never Confirmed, because the commit protocol already guarantees the source holds the value; (3) the #342 reconciliation kinds (Presumed etc.) build on the same confidence ladder. So it is one byte buying the inbound-vs-acknowledged distinction that #365 explicitly called out as missing.

src/Namotion.Interceptor.Tests/SubjectChangeContextTests.cs
RicoSuter: Preserved behavior made explicit, plus a pin: the old WithChangedTimestamp always kept the ambient received timestamp, and WithTimestamps(changed, null) does the same, so the appliers keep master's semantics. What is genuinely new is that the contract is now stated in the doc comment and pinned by this test, because the deleted WithState used to OVERWRITE received with its argument, and a naive WithTimestamps implementation would have silently done the same and lost received timestamps on nested scopes.

src/Namotion.Interceptor.Tests/VerifyChecksTests.PublicApi.verified.txt
RicoSuter: Agreed it deserves discussion; my take: consolidate, in a small follow-up rather than this PR. WithTimestamps(x, null) is now semantically identical to WithChangedTimestamp(x) (both preserve ambient received), so keeping both is two names for one behavior. Removing WithChangedTimestamp costs a mechanical sweep of its call sites (appliers, PathExtensions, docs) and one PublicApi snapshot line, gains a single-scope API. Perf is identical (same struct copy either way). If you want it in this PR say the word; keeping it separate keeps this diff reviewable.

src/Namotion.Interceptor.WebSocket/Client/WebSocketSubjectClientSource.cs
RicoSuter: Already resolved: the TODOs were removed in 1a63106b (Task: typed ChangeOrigin on ApplySubjectUpdate) when the client sites got their origin stamping restored via ChangeOrigin.FromSource(this) / FromSource(state.source). No TODO remains in this file on the current head.

src/Namotion.Interceptor.WebSocket/Client/WebSocketSubjectClientSource.cs
RicoSuter: Same as the sibling thread: resolved in 1a63106b; the apply sites now stamp ChangeOrigin.FromSource directly and no TODO remains on the current head.

src/Namotion.Interceptor.Connectors.Tests/Updates/SubjectUpdateExtensionsTests.cs
RicoSuter: Checked: the test already exercises ApplySubjectUpdate(update, factory, ChangeOrigin.FromSource(source)) directly with no SetValueFromSource call remaining (it was reworked when the typed origin parameter landed). Nothing left to remove.

src/Namotion.Interceptor.Connectors.Tests/SubjectSourceBaseTests.cs
RicoSuter: Removed the dead split declaration in e1c8152b (the WithSource wrapper that forced it is gone, so it collapses to var source = new TestSubjectSource(...)). Kept writeRetryQueueSize: 0 and LoadInitialStateOverride: both are load-bearing for this test (it is the retry-queue-disabled scenario and the assertion checks the value the override sets).

src/Namotion.Interceptor.Tracking/Transactions/SubjectPropertyChangeOperations.cs
RicoSuter: Done in e1c8152b: hoisted to a single var newValue = change.GetNewValue<object?>() used for both the arming sentValue and the setter invocation.

src/Namotion.Interceptor.Tracking/Transactions/SubjectTransactionInterceptor.cs
RicoSuter: Done in e1c8152b: currentContext was single-use, inlined SubjectChangeContext.Current.ReceivedTimestamp into the CaptureChange call and removed the comment.

docs/tracking.md
RicoSuter: Done in e1c8152b: rewrote the history-style phrasing in docs/tracking.md, docs/connectors-opcua-client.md, and the WithTimestamps XML doc to describe only the current state. Swept all branch-changed docs and XML comments; the remaining 'no longer/previously' hits describe runtime or transaction state (values no longer matching, previously-written changes), not past API, and were left.

src/Namotion.Interceptor/Interceptors/IWriteInterceptor.cs
```diff
@@ -61,20 +61,43 @@ public struct PropertyWriteContext<TProperty>
...
+    /// The origin of this write. Before the terminal write executes this is the attempted
+    /// origin (what the caller declared when setting the pending origin); when the terminal write lands (the same
+    /// point <see cref="IsWritten"/> becomes true) it is finalized: a stamped origin whose
+    /// final value differs from the sent value becomes Local, because the stored value was
+    /// computed locally rather than taken from the source.
+    /// </summary>
+    public ChangeOrigin Origin { get; internal set; }
+
+    /// <summary>The value the source sent, valid when <see cref="Origin"/> is stamped.</summary>
+    internal object? SentValue { get; }
```
RicoSuter: should we merge this into the Origin, eg a struct ChangeOriginWithValue : ChangeOrigin or something, or just rename it SourceSentValue or something? i dont really like the design of this file change here. any proposals?

RicoSuter: Three options, with a recommendation:

1. NOT into ChangeOrigin itself, deliberately: ChangeOrigin is the published provenance type on every SubjectPropertyChange, so carrying the sent value there would grow every queued change by a reference slot, keep the sent value alive for the change's lifetime in queues, and leak write-pipeline evidence into the public model (the sent value is meaningless after finalization and for Local origins). Also structs cannot inherit, so ChangeOriginWithValue : ChangeOrigin would need composition anyway.

2. Recommended: group the pair as one internal readonly struct on the write context, e.g. internal struct AttemptedOrigin { ChangeOrigin Origin; object? SentValue; }, stored as a single field; the public Origin property projects from it and FinalizeOrigin replaces the field when demoting to Local. This is the same shape PendingOrigin's thread-static frame already has, so TryConsume can return it directly. No public API change, no size change, and the two things that only mean something together live together instead of as loose siblings.

3. Minimal: keep the layout and rename to SourceSentValue with a doc comment stating it is only meaningful while the origin is stamped (pre-finalization).

I would do 2. Say which and I will implement it.

RicoSuter: Implemented the unified shape in c2976a09, folding in your unification idea: one internal readonly struct AttemptedOrigin { Origin, SentValue } shared by both sides via composition. The thread-static frame nests it as { HasValue, Target, Attempted } (Armed renamed to HasValue per the other discussion), TryConsume returns the pair directly, and PropertyWriteContext holds a single _attempted field with the public Origin property projecting from it; FinalizeOrigin demotes by resetting the field to default. The pending, attempted, finalized lifecycle is now stated on both types' doc comments and explained user-facing in docs/connectors.md. No public API, snapshot, or size change; full suite green.

src/Namotion.Interceptor.Tests/VerifyChecksTests.PublicApi.verified.txt
RicoSuter: Tracked as follow-up: #367 (consolidate WithChangedTimestamp into WithTimestamps as the single timestamp scope).

RicoSuter:
## Full benchmark suite: branch vs master

`pwsh scripts/benchmark.ps1 -Stash -BaseBranch origin/master`, full suite (26 benchmarks, all classes), single process launch (DefaultJob), on an Intel i7-11700K pinned at 0.80 GHz (Ubuntu 24.04, .NET 9.0.11).

Absolute numbers are not comparable to the earlier M4 Max comment: this box is clock-locked, so everything is slower in nanoseconds. Only the within-run branch-vs-master deltas are meaningful. Each benchmark's error bars are tight (within-process), but master and branch run in separate processes, so between-process drift on this box is roughly a few percent. Treat sub-5% micro deltas as noise unless corroborated by a higher launch count.

### Deltas that stand out

| Benchmark | master | branch | delta | allocations |
|---|---:|---:|---|---|
| SubjectTransactionBenchmark.CommitChanges (Local) | 34.14 us | 37.07 us | **+8.6%** | 5072 B, unchanged |
| SubjectTransactionBenchmark.CommitChanges (SingleSource) | 37.94 us | 42.85 us | **+12.9%** | 15096 to 15496 B (**+400**) |
| SubjectTransactionBenchmark.CommitChanges (SingleSource+Local) | 37.83 us | 41.09 us | **+8.6%** | 15096 to 15496 B (**+400**) |
| SubjectTransactionBenchmark.CommitChanges (MultiSource) | 42.48 us | 47.13 us | **+10.9%** | 54976 to 56960 B (**+1984**) |
| RegistryBenchmark.WriteNoOp | 365.6 ns | 375.6 ns | +2.7% (+10 ns) | 0, unchanged |
| RegistryBenchmark.Write | 993.8 ns | 1011.4 ns | +1.8% (+18 ns) | 0, unchanged |
| RegistryBenchmark.WriteWithTimestampScope | 929.4 ns | 904.4 ns | -2.7% (faster) | 0, unchanged |
| SubjectSourceBenchmark.WriteToSource | 1.740 ms | 1.525 ms | -12.3% (faster) | 0, unchanged |
| SubjectSourceBenchmark.WriteToRegistrySubjects | 2.467 ms | 2.235 ms | -9.4% (faster) | 0, unchanged |

Everything else (ServiceOrderResolver, SourcePathProvider, SubjectUpdate, and the remaining RegistryBenchmark methods) is within run-to-run noise and allocation-identical.

### Read

- **Allocation-neutral on the hot write paths.** WriteNoOp / Write / WriteWithTimestampScope allocate 0 B on both branches. The only allocation change is on transaction commit with sources: +400 B (single source) and +1984 B (multi source), the per-change origin data carried through the commit records.
- **The write micro-path carries a fixed ~15 ns/write** (here +10 to +18 ns; the effect resolves cleanly at higher launch counts). A focused isolation traced this to the `PropertyWriteContext<T>` struct growing by the `AttemptedOrigin` field, not the thread-static pending-origin read: removing the TLS read leaves the regression intact, removing the field recovers it. Moving the origin out of the struct was tried and rejected, because real writes read `Origin` several times (finalization plus publishers), so out-of-band reads recover WriteNoOp but roughly double the cost on `Write`. The field is kept.
- **Transaction commit regresses ~8 to 13%** across all four modes, with the small allocation bump above. This is the feature's cost on the commit path (per-change origin finalization and capture).
- **Source-side macro benchmarks come out faster** on the branch (WriteToSource -12%, WriteToRegistrySubjects -9%), consistent with dropping the ambient source-scope reads at publish time. Both are high-variance and worth confirming with more launches.

Net: allocation-neutral on the common write path; a real but small fixed per-write CPU cost (~15 ns) inherent to carrying typed per-write origin; a ~10% cost on transaction commit (plus a few hundred bytes to two kilobytes per sourced commit); source paths neutral to faster.

<details>
<summary>Raw results, branch (feature/change-origin-design)</summary>

```
| Type                          | Method                    | Mode                 | Type        | Mean               | Error           | StdDev          | Gen0      | Gen1      | Gen2     | Allocated  |
|------------------------------ |-------------------------- |--------------------- |------------ |-------------------:|----------------:|----------------:|----------:|----------:|---------:|-----------:|
| ServiceOrderResolverBenchmark | NoDependencies            | ?                    | ?           |        947.0134 ns |       2.9779 ns |       2.7855 ns |    0.1945 |         - |        - |     1632 B |
| SourcePathProviderBenchmark   | TryGetPropertyFromPath    | ?                    | ?           |        268.3580 ns |       0.2878 ns |       0.2552 ns |    0.0753 |         - |        - |      632 B |
| SubjectSourceBenchmark        | WriteToRegistrySubjects   | ?                    | ?           |  2,235,278.3603 ns |   4,838.0331 ns |   4,525.4991 ns |         - |         - |        - |          - |
| SubjectUpdateBenchmark        | CreateCompleteUpdate      | ?                    | ?           |      5,812.3940 ns |      23.2676 ns |      21.7645 ns |    0.7019 |    0.0153 |        - |     5904 B |
| ServiceOrderResolverBenchmark | LinearChain               | ?                    | ?           |      1,036.4461 ns |       4.2191 ns |       3.9465 ns |    0.3147 |         - |        - |     2640 B |
| SourcePathProviderBenchmark   | VisitPropertiesFromPaths  | ?                    | ?           |        537.7294 ns |       1.7039 ns |       1.5939 ns |    0.1526 |         - |        - |     1280 B |
| SubjectSourceBenchmark        | WriteToSource             | ?                    | ?           |  1,525,217.6233 ns |  30,343.1334 ns |  42,536.8388 ns |         - |         - |        - |          - |
| SubjectUpdateBenchmark        | CreatePartialUpdate       | ?                    | ?           |      1,942.9559 ns |      10.1583 ns |       9.5021 ns |    0.3166 |         - |        - |     2656 B |
| ServiceOrderResolverBenchmark | MixedDependencies         | ?                    | ?           |      1,044.6199 ns |       3.6869 ns |       3.4487 ns |    0.2460 |         - |        - |     2064 B |
| SourcePathProviderBenchmark   | TryGetPropertyFromSegment | ?                    | ?           |         31.8120 ns |       0.0560 ns |       0.0524 ns |         - |         - |        - |          - |
| ServiceOrderResolverBenchmark | WithFirstLastGroups       | ?                    | ?           |      1,248.6222 ns |       4.3403 ns |       4.0599 ns |    0.3071 |         - |        - |     2576 B |
| SourcePathProviderBenchmark   | TryGetPath                | ?                    | ?           |        215.5926 ns |       2.9110 ns |       2.5805 ns |    0.0315 |         - |        - |      264 B |
| SubjectTransactionBenchmark   | CommitChanges Local             |                | ?           |     37,065.2328 ns |     146.8471 ns |     130.1761 ns |    0.5493 |         - |        - |     5072 B |
| SubjectTransactionBenchmark   | CommitChanges SingleSource      |                | ?           |     42,845.1382 ns |     203.8278 ns |     180.6880 ns |    1.8311 |         - |        - |    15496 B |
| SubjectTransactionBenchmark   | CommitChanges SingleSrc+Local   |                | ?           |     41,086.8156 ns |     150.6193 ns |     140.8893 ns |    1.8311 |         - |        - |    15496 B |
| SubjectTransactionBenchmark   | CommitChanges MultiSource       |                | ?           |     47,130.0769 ns |     124.1161 ns |     116.0983 ns |    6.7749 |    0.4883 |        - |    56960 B |
| RegistryBenchmark             | AddLotsOfPreviousCars     | ?                    | interceptor | 49,405,018.4400 ns | 501,370.0554 ns | 468,981.8611 ns | 3000.0000 | 2000.0000 | 500.0000 | 21696891 B |
| RegistryBenchmark             | IncrementDerivedAverage   | ?                    | interceptor |      5,312.9952 ns |      19.3826 ns |      18.1305 ns |    0.0153 |         - |        - |      128 B |
| RegistryBenchmark             | WriteNoOp                 | ?                    | interceptor |        375.5782 ns |       0.8050 ns |       0.7137 ns |         - |         - |        - |          - |
| RegistryBenchmark             | Write                     | ?                    | interceptor |      1,011.3745 ns |       3.7799 ns |       3.5357 ns |         - |         - |        - |          - |
| RegistryBenchmark             | WriteWithTimestampScope   | ?                    | interceptor |        904.3530 ns |       2.3368 ns |       2.1859 ns |         - |         - |        - |          - |
| RegistryBenchmark             | Read                      | ?                    | interceptor |        445.6852 ns |       1.3156 ns |       1.2306 ns |         - |         - |        - |          - |
| RegistryBenchmark             | DerivedAverage            | ?                    | interceptor |        289.4272 ns |       0.8686 ns |       0.8125 ns |         - |         - |        - |          - |
| RegistryBenchmark             | ChangeAllTires            | ?                    | interceptor |     12,610.1326 ns |      37.7293 ns |      35.2920 ns |    1.8463 |    0.1068 |        - |    15488 B |
| RegistryBenchmark             | GetOrAddSubjectId         | ?                    | interceptor |         21.0134 ns |       0.0125 ns |       0.0105 ns |         - |         - |        - |          - |
| RegistryBenchmark             | GenerateSubjectId         | ?                    | interceptor |        787.0427 ns |       1.0575 ns |       0.9892 ns |    0.0086 |         - |        - |       72 B |
| RegistryBenchmark             | KnownSubjectsSnapshot     | ?                    | interceptor |          0.0006 ns |       0.0006 ns |       0.0006 ns |         - |         - |        - |          - |
| RegistryBenchmark             | ReadParents               | ?                    | interceptor |          0.2612 ns |       0.0002 ns |       0.0002 ns |         - |         - |        - |          - |
```
</details>

<details>
<summary>Raw results, origin/master</summary>

```
| Type                          | Method                    | Mode                 | Type        | Mean               | Error           | StdDev          | Gen0      | Gen1      | Gen2     | Allocated  |
|------------------------------ |-------------------------- |--------------------- |------------ |-------------------:|----------------:|----------------:|----------:|----------:|---------:|-----------:|
| ServiceOrderResolverBenchmark | NoDependencies            | ?                    | ?           |        918.5989 ns |       3.4874 ns |       3.0915 ns |    0.1945 |         - |        - |     1632 B |
| SourcePathProviderBenchmark   | TryGetPropertyFromPath    | ?                    | ?           |        267.3423 ns |       0.5741 ns |       0.5370 ns |    0.0753 |         - |        - |      632 B |
| SubjectSourceBenchmark        | WriteToRegistrySubjects   | ?                    | ?           |  2,467,239.0319 ns |   3,986.3780 ns |   3,728.8605 ns |         - |         - |        - |          - |
| SubjectUpdateBenchmark        | CreateCompleteUpdate      | ?                    | ?           |      5,946.7478 ns |      19.3090 ns |      18.0616 ns |    0.7019 |    0.0153 |        - |     5904 B |
| ServiceOrderResolverBenchmark | LinearChain               | ?                    | ?           |      1,040.0829 ns |       5.2282 ns |       4.8904 ns |    0.3147 |         - |        - |     2640 B |
| SourcePathProviderBenchmark   | VisitPropertiesFromPaths  | ?                    | ?           |        542.9265 ns |       1.3166 ns |       1.2315 ns |    0.1526 |         - |        - |     1280 B |
| SubjectSourceBenchmark        | WriteToSource             | ?                    | ?           |  1,739,800.5010 ns |  19,186.4074 ns |  17,946.9774 ns |         - |         - |        - |          - |
| SubjectUpdateBenchmark        | CreatePartialUpdate       | ?                    | ?           |      1,948.9021 ns |       6.3910 ns |       5.9782 ns |    0.3166 |         - |        - |     2656 B |
| ServiceOrderResolverBenchmark | MixedDependencies         | ?                    | ?           |      1,034.0582 ns |       2.8889 ns |       2.5609 ns |    0.2460 |         - |        - |     2064 B |
| SourcePathProviderBenchmark   | TryGetPropertyFromSegment | ?                    | ?           |         34.1164 ns |       0.0609 ns |       0.0540 ns |         - |         - |        - |          - |
| ServiceOrderResolverBenchmark | WithFirstLastGroups       | ?                    | ?           |      1,258.7320 ns |       4.5635 ns |       4.2687 ns |    0.3071 |         - |        - |     2576 B |
| SourcePathProviderBenchmark   | TryGetPath                | ?                    | ?           |        200.3060 ns |       0.4277 ns |       0.3792 ns |    0.0315 |         - |        - |      264 B |
| SubjectTransactionBenchmark   | CommitChanges Local             |                | ?           |     34,139.1896 ns |     152.7960 ns |     142.9255 ns |    0.5493 |         - |        - |     5072 B |
| SubjectTransactionBenchmark   | CommitChanges SingleSource      |                | ?           |     37,941.4670 ns |     186.2078 ns |     174.1789 ns |    1.7700 |         - |        - |    15096 B |
| SubjectTransactionBenchmark   | CommitChanges SingleSrc+Local   |                | ?           |     37,825.0147 ns |     223.1236 ns |     208.7100 ns |    1.7700 |         - |        - |    15096 B |
| SubjectTransactionBenchmark   | CommitChanges MultiSource       |                | ?           |     42,481.4241 ns |     164.5500 ns |     153.9202 ns |    6.5308 |    0.4883 |        - |    54976 B |
| RegistryBenchmark             | AddLotsOfPreviousCars     | ?                    | interceptor | 50,701,956.9200 ns | 511,659.3826 ns | 478,606.5042 ns | 3000.0000 | 2000.0000 | 500.0000 | 21696574 B |
| RegistryBenchmark             | IncrementDerivedAverage   | ?                    | interceptor |      5,524.5553 ns |      52.4928 ns |      49.1018 ns |    0.0153 |         - |        - |      128 B |
| RegistryBenchmark             | WriteNoOp                 | ?                    | interceptor |        365.5628 ns |       0.6737 ns |       0.6301 ns |         - |         - |        - |          - |
| RegistryBenchmark             | Write                     | ?                    | interceptor |        993.8128 ns |       3.3024 ns |       3.0891 ns |         - |         - |        - |          - |
| RegistryBenchmark             | WriteWithTimestampScope   | ?                    | interceptor |        929.3874 ns |       2.3588 ns |       2.2064 ns |         - |         - |        - |          - |
| RegistryBenchmark             | Read                      | ?                    | interceptor |        438.0601 ns |       1.5929 ns |       1.4900 ns |         - |         - |        - |          - |
| RegistryBenchmark             | DerivedAverage            | ?                    | interceptor |        284.4580 ns |       0.9767 ns |       0.9136 ns |         - |         - |        - |          - |
| RegistryBenchmark             | ChangeAllTires            | ?                    | interceptor |     12,748.4977 ns |      34.2855 ns |      30.3932 ns |    1.8463 |    0.1068 |        - |    15488 B |
| RegistryBenchmark             | GetOrAddSubjectId         | ?                    | interceptor |         21.0548 ns |       0.0181 ns |       0.0169 ns |         - |         - |        - |          - |
| RegistryBenchmark             | GenerateSubjectId         | ?                    | interceptor |        784.5846 ns |       0.6514 ns |       0.6093 ns |    0.0086 |         - |        - |       72 B |
| RegistryBenchmark             | KnownSubjectsSnapshot     | ?                    | interceptor |          0.0004 ns |       0.0005 ns |       0.0004 ns |         - |         - |        - |          - |
| RegistryBenchmark             | ReadParents               | ?                    | interceptor |          0.2562 ns |       0.0003 ns |       0.0003 ns |         - |         - |        - |          - |
```
</details>

Issue #369, opened 2026-07-13
Make origin plumbing explicit: parameterized set method and outcome return instead of thread-static state


RicoSuter:
Follow-up to #366/#372. Replace the thread-static origin plumbing with explicit parameters and return values. Breaking changes are fine.

## Motivation

The one-shot pending stamp is correct and fully tested, but it is hidden state that polices itself: most review findings across the #366/#372design rounds targeted the mailbox mechanics (frame restore for re-entrancy, target matching, destructive consume, clear-on-entry, the constructor side effect, and the PR 2 outcome slot whose guards failed review twice). Explicit dataflow does not generate this class of finding: a nested setter simply never passes an origin, so "nested writes are Local" becomes syntax instead of an invariant. Measurements showed no meaningful perf difference either way, so this is a maintainability refactor, not an optimization.

## Proposed design

One generated method per property, not two. The C# property setter must exist syntactically but cannot take parameters, so it becomes a one-line forwarder:

```csharp
public partial string Name
{
    get => ...;
    set => SetName(value, ChangeOrigin.Local);
}

public void SetName(string value, ChangeOrigin origin) { /* full body: hooks, chain */ }
```

There is no duplicated body and no second real set method: the origin-carrying method IS the setter body, and the property setter forwards with Local.

Plumbing changes:

1. `SubjectPropertyMetadata.SetValue` changes signature to `Action<object, object?, ChangeOrigin>` (invokes the generated origin method). Plain local applies pass `ChangeOrigin.Local`.
2. `IInterceptorExecutor.SetPropertyValue` gains a `ChangeOrigin origin` parameter and returns a small readonly `WriteOutcome` struct (`IsWritten`, `ValueUnchanged`, `Origin`) instead of `bool`. The context constructor receives the origin directly; the sent-value evidence is the value argument captured at entry.
3. `ValueUnchanged` is set on the write context by the equality handler (the context already flows by ref through the chain), read from the returned outcome at exit. No thread-static outcome slot.
4. `SetValueFromOrigin` / `SetValueFromSource` keep their public signatures; internally they invoke the origin-carrying metadata delegate and read the returned outcome for correction detection (#372 semantics unchanged).
5. Timestamps stay ambient: one logical event shares one time across cascades, which is genuinely dynamic-extent. Only origin and outcome become explicit.

## Concepts removed

- `PendingOrigin`, `PendingOriginScope`, the thread-static frame, `HasValue`, `Set`/`TryConsume`
[...]

PR #372, opened 2026-07-13
Correction origin kind for equality-suppressed diverged source values

### LINKED ISSUE #365: Typed ChangeOrigin discriminator with causal TriggeredBy for change notifications

RicoSuter:
Adds the `Correction` origin kind: when an inbound value is projected by a hook to the value the model already stores, the equality check suppresses the write and, previously, nothing was published, leaving the source diverged indefinitely (the case from the #365 comment). Detection now synthesizes a `Correction(source)` change so the source converges to the model. Builds on #366 (typed ChangeOrigin); base is that branch and this PR retargets to master after #366 merges.

Design doc: `docs/superpowers/specs/2026-07-12-change-origin-design.md` (PR 2 sections; the superpowers docs are removed before merge).

## Three-outcome matrix

| Case | Outcome |
|---|---|
| stored value changed | normal write and change publication |
| stored value unchanged, sent value differs from stored | `Correction(S)` synthesized |
| stored value unchanged, sent value equals stored | nothing (pure echo or no-op) |

## How it works

- Detection lives in `SetValueFromOrigin` (no new interceptor, zero write-chain cost): the equality handler records the value-unchanged outcome for stamped writes, and a `FromSource` apply that ends unwritten with an unchanged projected value but a diverging sent value synthesizes a correction (observable value, old == new). Divergence is judged against the pre-transform sent value, so `ApplySubjectUpdate` transforms that project the sent value onto the stored value are detected too. A synthesis getter that throws propagates to the caller (stamped setters are already a throwing API via validators); a broken read path surfaces instead of becoming a silently diverged source. A set-only property has no getter and thus no observable value to assert, so synthesis is skipped. Cancelled writes, transaction captures, and `Confirmed` replays structurally self-exclude.
- Committed state under transactions: a correction only ever asserts committed state. On a flow that owns an active transaction the observable getter returns the transaction's pending overlay, so synthesis detaches the transaction for its read and compares against the committed value (the same value the equality check compared against). A real divergence is still corrected while a transaction is pending, and the uncommitted value cannot reach a source because delivery revalidates on the processor's flow, which carries no transaction (and detaches one if it ever does).
- Value equality: divergence detection and delivery revalidation compare with the declared property type's default equality (`EqualityComparer<T>` through the shared `PropertyValueEquality`), the same comparison the equality handler uses to suppress the write. A type that implements `IEquatable<T>` without overriding `object.Equals` is therefore not read as divergent from an equal echo, which a boxed `object.Equals` would misjudge into synthesizing a fresh correction on every echo. This subsumes the enum wire shape: an enum value delivered as its underlying integer (the OPC UA wire shape) compares and stores as the enum. `PropertyValueEquality` unwraps `Nullable<T>` and coerces the boxed underlying integer to the enum before comparing, so a nullable enum is not misread as divergence. It stays in sync with the origin survival check `SentValueEqualsAfterUnbox` (in master via #374), which applies the same unwrap, and diverges only on a genuinely incompatible box (detection throws, survival demotes to `Local`).
- Delivery is topology-based: corrections bypass the own-source echo skip and are delivered by every processor whose property filter matches (origin identity need not equal processor identity, e.g. WebSocket per-connection origins). Both buffered and immediate modes deliver corrections; an immediate-mode processor runs each correction through the same send-time revalidation.
- Correctness under concurrency: send-time revalidation drops a correction whose model value moved, and a bounded post-write revalidation loop (8 attempts) re-asserts the current value if an inbound apply lands mid-write, warning with a pointer to `RequestResynchronization` on exhaustion. A failed correction write is dropped, never retried (the retry queue filters corrections); recovery is the source's next inbound event. A revalidation read whose getter throws drops the correction instead of terminating the processor loop.
- Dedup: a normal change beats a queued correction regardless of order; a correction displaced from the flush slot by a normal change for the same property is routed through send-time revalidation rather than dropped, so a fresh correction queued behind an older normal change still converges the source. Two corrections keep the later one whole, preserving the old == new correction invariant. The observable never sees corrections; direct queue subscribers filter on `Kind`.
- `ProcessAsync` guards against a second call per instance (the flush scratch buffers assume a single consumer).

## Extracted to #374 (merged)

The origin-survival hardening (`FinalizeOrigin` no longer throws on null, type-mismatched, or nullable-enum-underlying sent values) and the Local-replay origin-arming perf change were split into #374 and merged to master, since neither depends on corrections. This PR is rebased on top; the `Correction` kind synthesized here relies on that survival behavior for echo suppression.

## Timestamps

A correction carries a local assertion timestamp: the local clock at synthesis, re-stamped at each delivery attempt so it reports when the assertion was sent. **`Correction.ChangedTimestamp` records when the assertion was created or sent. It does not guarantee monotonic wire ordering. A connector that requires monotonic versions must provide them in its own outbound writer**, the only layer that knows its endpoint's rule. No in-repo connector or endpoint rejects a write for carrying an older timestamp (inbound writes are unconditional assignments and the timestamp is metadata), so the correction layer fabricates no ordering above the local clock. The write-timestamp metadata is never advanced (an unchanged value is not a new write), so the model keeps the value's real last-change time and model and source may intentionally disagree on the timestamp of an identical value.

An earlier revision of this PR carried a Lamport lower bound, a processor-wide ratchet, and a `DateTimeOffset.MaxValue` ceiling guard to keep fabricated stamps acceptable to a hypothetical timestamp-ordering endpoint. Review established no such endpoint exists in this repo, so that machinery (about 90 lines and nine tests) was removed in favor of the local-clock contract above. Endpoint-specific wire ordering, if ever needed, belongs in that connector's outbound writer.

## Known limitation (follow-up #373)

Inbound applies are not ordered against outbound writes, so a notification a source emitted before an outbound write but delivered after it can revert the model to the older value, which is then echo-suppressed and does not flow back. This affects any outbound write, not just corrections, and is a pre-existing local-first residue rather than something this PR introduces. A correction's post-write revalidation closes the in-flight window only up to its read; it cannot observe an apply that follows the read. Recovery is the source's next inbound event, read-after-write where the connector supports it (OPC UA, on by default), or explicit resynchronization. Tracked as a follow-up in #373 (a clock-free fix via an in-band echo fence or read-after-write authority), with reconciliation (#342) as the general escape hatch.

## Tests

20 delivery tests including the gated in-flight window race, the value-cycle send-time re-stamp pin, the flush partition order, dedup revalidation routing (fresh correction behind a stale normal converges; stale correction ahead of a normal is dropped), bounded-loop exhaustion, the WebSocket identity-mismatch delivery shape, the dedup matrix, immediate-mode delivery, the throwing-getter drop, revalidation reading committed state under an active transaction, and retry filtering. 18 detection tests including the equality-suppressed `ApplySubjectUpdate` transform regression, committed-state synthesis under a pending transaction (echo and divergence directions), enum-over-underlying-type and nullable-enum echo (not misread as divergence), the `IEquatable`-without-`object.Equals` echo, the set-only (no getter) skip, the deliberate getter-throw propagation, the concurrency drop-or-fresh race, and transaction-capture and Confirmed self-exclusion. Origin survival across the null, type-mismatch, and nullable-enum cases is pinned in #374. The full unit suite (`Category != Integration`) passes. Reflection-based test seams were replaced with internal members via the existing `InternalsVisibleTo`. Public API grows by exactly the `Correction` kind and factory.

`TriggeredBy` and `Presumed` remain tracked in #342 as additive extensions.

Closes #365


RicoSuter:
## Write and transaction benchmarks: branch vs master (#366) vs pre-#366

`pwsh scripts/benchmark.ps1 -Stash -BaseBranch origin/master` with filters `*Write*` and `*CommitChanges*`, `-LaunchCount 5`, on the i7-11700K (Ubuntu 24.04, .NET 9.0.11 runtime, SDK 10.0.100). Clocks verified pinned at 0.80 GHz before the run (`scaling_max_freq` on all cores). Measured at branch commit b1fec297; the commits up to 18fa0e69 were audited and touch only correction synthesis/delivery plus a work-equivalent dedup branch, none of which these benchmarks execute, so the numbers represent the tip.

**Cross-session calibration.** Fresh master (= #366 merged) anchors against the #366 comment's branch numbers at a uniform 6.0x to 6.3x on every CPU-bound benchmark, which means that earlier session actually ran at turbo (~4.8 GHz) rather than the stated 0.80 GHz. Absolute nanoseconds are therefore not comparable across sessions. Within-session ratios are, and the uniformity of the anchor is the evidence: relative costs are preserved. The pre-#366 column below is chained accordingly: (pre-#366 to #366, old session) composed with (#366 to branch, this session). `WriteToSource` anchors at 3.9x because part of its time is async/queue-bound and does not scale with clock, so its chained value is low confidence.

### This session: branch vs master (pinned 0.80 GHz, LaunchCount 5)

| Benchmark | master (#366) | branch | delta | allocations |
|---|---:|---:|---|---|
| RegistryBenchmark.WriteNoOp | 2.274 us | 2.309 us | +1.5% | 0, unchanged |
| RegistryBenchmark.Write | 6.340 us | 6.250 us | -1.4% | 0, unchanged |
| RegistryBenchmark.WriteWithTimestampScope | 5.721 us | 5.805 us | +1.5% | 0, unchanged |
| SubjectSourceBenchmark.WriteToSource | 5.984 ms | 6.103 ms | +2.0% | 0, unchanged |
| SubjectSourceBenchmark.WriteToRegistrySubjects | 13.365 ms | 14.028 ms | +5.0% mean, -0.1% median | 0, unchanged |
| SubjectTransactionBenchmark.CommitChanges (Local) | 222.9 us | 207.0 us | **-7.1%** | 4.95 KB, unchanged |
| SubjectTransactionBenchmark.CommitChanges (SingleSource) | 258.6 us | 258.0 us | -0.2% | 15.13 KB, unchanged |
| SubjectTransactionBenchmark.CommitChanges (SingleSource+Local) | 249.4 us | 243.9 us | -2.2% | unchanged (KB rounding) |
| SubjectTransactionBenchmark.CommitChanges (MultiSource) | 285.7 us | 286.8 us | +0.4% | 55.63 KB, unchanged |

The WriteToRegistrySubjects mean is skewed by one noisy launch on the branch side (StdDev 1.55 ms vs 0.23 ms on master); medians differ by 0.1%, so it reads as neutral.

### Three-way deltas (pre-#366 chained through the anchor)

| Benchmark | pre-#366 to #366 (old session) | #366 to branch (this session) | pre-#366 to branch (chained) |
|---|---:|---:|---:|
| WriteNoOp | +2.7% | +1.5% | **+4.3%** |
| Write | +1.8% | -1.4% | +0.3% |
| WriteWithTimestampScope | -2.7% | +1.5% | -1.3% |
| WriteToSource | -12.3% | +2.0% | -10.6% (low confidence) |
| WriteToRegistrySubjects | -9.4% | ~0% (median) | ~-9% |
| CommitChanges (Local) | +8.6% | -7.1% | **+0.8%** |
| CommitChanges (SingleSource) | +12.9% | -0.2% | **+12.7%** |
| CommitChanges (SingleSource+Local) | +8.6% | -2.2% | +6.2% |
| CommitChanges (MultiSource) | +10.9% | +0.4% | +11.4% |

### Read

- **This PR is essentially free on top of #366.** Write micro-paths move within +/-1.5% (that is +/-35 to 90 ns at 0.8 GHz, single-digit nanoseconds at real clocks; the WriteNoOp +1.5% matches the predicted single branch in the equality handler for outcome recording). Allocations are bit-identical on every benchmark.
- **Local transaction commit is 7.1% faster than master**, recovering most of #366's +8.6% Local-commit regression (chained: +0.8% vs pre-#366). This is the skip-arming optimization for Local-origin replays (87666bf9).
- **The remaining regressions vs pre-#366 belong to #366, not this PR**: sourced commits at +6 to +13%, and the fixed per-write cost on the micro path (WriteNoOp chained +4.3%). The #366 isolation analysis attributed that fixed cost to the `AttemptedOrigin` field in `PropertyWriteContext`, not the thread-static read, which is relevant for the follow-up #369: removing the thread-static machinery will likely not return WriteNoOp to the pre-#366 baseline, since the field stays.

<details>
<summary>Raw results, branch (feature/change-origin-corrections @ b1fec297)</summary>

```
.NET SDK 10.0.100, Host/Job: .NET 9.0.11, X64 RyuJIT x86-64-v4, LaunchCount=5

| Type                   | Method                  | Mean          | Error       | StdDev        | Median        | Allocated |
|----------------------- |------------------------ |--------------:|------------:|--------------:|--------------:|----------:|
| SubjectSourceBenchmark | WriteToRegistrySubjects | 14,028.183 us | 630.1915 us | 1,545.8706 us | 13,241.964 us |         - |
| SubjectSourceBenchmark | WriteToSource           |  6,103.369 us | 116.5740 us |   292.4613 us |  6,038.133 us |         - |
| RegistryBenchmark      | WriteNoOp               |      2.309 us |   0.0340 us |     0.0853 us |      2.299 us |         - |
| RegistryBenchmark      | Write                   |      6.250 us |   0.0310 us |     0.0755 us |      6.221 us |         - |
| RegistryBenchmark      | WriteWithTimestampScope |      5.805 us |   0.0455 us |     0.1124 us |      5.748 us |         - |

| Method        | Mode                | Mean     | Error   | StdDev  | Median   | Gen0   | Gen1   | Allocated |
|-------------- |-------------------- |---------:|--------:|--------:|---------:|-------:|-------:|----------:|
| CommitChanges | Local               | 207.0 us | 0.79 us | 2.00 us | 207.7 us | 0.4883 |      - |   4.95 KB |
| CommitChanges | SingleSource        | 258.0 us | 1.01 us | 2.50 us | 257.6 us | 1.4648 |      - |  15.13 KB |
| CommitChanges | SingleSource+Local  | 243.9 us | 0.67 us | 1.66 us | 244.0 us | 1.4648 |      - |  15.14 KB |
| CommitChanges | MultiSource         | 286.8 us | 0.94 us | 2.37 us | 286.2 us | 6.3477 | 0.4883 |  55.63 KB |
```
</details>

<details>
<summary>Raw results, origin/master (f7fc9bf7, #366 merged)</summary>

```
.NET SDK 10.0.100, Host/Job: .NET 9.0.11, X64 RyuJIT x86-64-v4, LaunchCount=5

| Type                   | Method                  | Mean          | Error      | StdDev      | Median        | Allocated |
|----------------------- |------------------------ |--------------:|-----------:|------------:|--------------:|----------:|
| SubjectSourceBenchmark | WriteToRegistrySubjects | 13,364.566 us | 96.2769 us | 234.3513 us | 13,257.086 us |         - |
| SubjectSourceBenchmark | WriteToSource           |  5,984.028 us | 63.2323 us | 167.6831 us |  5,988.350 us |         - |
| RegistryBenchmark      | WriteNoOp               |      2.274 us |  0.0114 us |   0.0283 us |      2.270 us |         - |
| RegistryBenchmark      | Write                   |      6.340 us |  0.0636 us |   0.1549 us |      6.391 us |         - |
| RegistryBenchmark      | WriteWithTimestampScope |      5.721 us |  0.0622 us |   0.1560 us |      5.673 us |         - |

| Method        | Mode                | Mean     | Error   | StdDev  | Gen0   | Gen1   | Allocated |
|-------------- |-------------------- |---------:|--------:|--------:|-------:|-------:|----------:|
| CommitChanges | Local               | 222.9 us | 0.66 us | 1.63 us | 0.4883 |      - |   4.95 KB |
| CommitChanges | SingleSource        | 258.6 us | 1.05 us | 2.61 us | 1.4648 |      - |  15.13 KB |
| CommitChanges | SingleSource+Local  | 249.4 us | 0.92 us | 2.25 us | 1.4648 |      - |  15.13 KB |
| CommitChanges | MultiSource         | 285.7 us | 0.74 us | 1.81 us | 6.3477 | 0.4883 |  55.63 KB |
```
</details>

PR #377, opened 2026-07-18
Add per-property subscriptions & unify property change interceptors

RicoSuter:
## Overview

Merges the two property-change delivery interceptors (`PropertyChangeObservable` and `PropertyChangeQueue`) into a single `PropertyChangeInterceptor` with three delivery channels (Rx observable, pull queue, and new per-property subscriptions), sharing one delivery contract.

A feature that cares about one or a few specific properties (a heartbeat, a watchdog, a dirty flag) can now register an indexed, allocation-free callback without an observer, a queue, a thread, or a background service of its own.

## What changed

**Merged interceptor.** One `IWriteInterceptor` owns the Rx observable and the high-performance queue through an immutable `DispatchState` snapshot (a single volatile field, null when idle). The change payload is built once per write instead of twice, removing the duplicate `SubjectPropertyChange.Create` and the double derived-getter invocation that two separate interceptors incurred.

**Per-property subscriptions.** Registrations are stored on the subject itself (`IInterceptorSubject.Data`), the way `INotifyPropertyChanged` handlers live with their source object:

- `property.Subscribe(observer | callback)` (low level, via `PropertyReference`)
- `subject.SubscribeToProperty(x => x.Foo, observer | callback)` (strongly typed, strict direct-selector resolver)
- `IPropertyChangeObserver` / `PropertyChangeCallback`, delivering `SubjectPropertyChange` by `in` reference

Subscriptions survive detach and re-attach (dormant while unattached, revived on re-attach), follow the subject instance rather than a graph path, and use standard `IDisposable` ownership with mandatory disposal (a dropped handle keeps the process-wide gate open). Contexts that never subscribe pay only cache-hot loads on the write path.

**One delivery contract for all three channels.** Consumers are resolved after the write commits, behind a full memory fence (a Dekker/StoreLoad pairing with the subscribe-side barrier), so a subscription installed while a write is in flight is never missed: a write that commits after the subscribing call returns is always delivered; a racing write may be delivered with `OldValue` equal to `NewValue`; a write committed before may not be (read the property after subscribing). The observable subscribe joins the Rx subject before publishing the dispatch state within a single critical section, so a first observer racing a writer is covered too.

**Notification ordering pinned to lifecycle.** `PropertyChangeInterceptor` now declares `[RunsBefore(LifecycleInterceptor)]`, so notifications dispatch on the unwind after attach/detach reconciliation instead of before it. Previously the relative order was an unstated tie-break of registration order (master behaves the same way), which meant a subject assigned to a property was still dormant while consumers were told about it: a synchronous callback writing to that new child hit the zero-interceptor terminal and its write was silently never delivered. Value-typed writes are unaffected (reconciliation early-returns for non-subject types) and asynchronous consumers stop racing a half-attached graph. `DerivedPropertyChangeHandler` gained the matching `[RunsBefore(PropertyChangeInterceptor)]` so the whole chain is pinned rather than implied, and tests assert the resolved order directly. Three consequences are pinned and documented: a removal now reaches consumers with the departing subject already detached (writes to it are stored but untracked, which is intended), a throwing `ILifecycleHandler` suppresses that write's notification, and a handler that writes properties while attaching emits those changes before the structural change. Audited as unused in-repo: there are no production synchronous consumers and no lifecycle handler writes properties.

**Correctness hardening from review.** Only committed writes publish (an inner interceptor veto no longer produces phantom notifications, a latent master bug); null arguments are rejected at every subscribe entry point; a foreign value under the reserved subject-data key fails loud instead of livelocking; three latent `PropertyChangeQueueSubscription` races fixed (enqueue-versus-dispose `ObjectDisposedException`, completion lost-wakeup, non-atomic dispose).

## Breaking changes

The library is released on NuGet, so both halves below belong in the release notes.

**Compile level.** Removed: the `PropertyChangeObservable` and `PropertyChangeQueue` types, `WithPropertyChangeObservable()` and `WithPropertyChangeQueue()` (replaced by `WithPropertyChangeSubscriptions()`), and `PropertyChangeQueueSubscription`'s public constructor. The removed `PropertyChangeQueue` implemented `IDisposable`; its replacement `PropertyChangeInterceptor` does not, because it releases nothing and per subscription disposal is the teardown path. `CreatePropertyChangeQueueSubscription()` lost its unused scheduler parameter. `WithFullPropertyTracking()`, `GetPropertyChangeObservable()` and `PropertyChangeQueueSubscription.TryDequeue` are unchanged, and a scan of an external consumer of this library (about 4000 files) found exactly one affected line, a `WithPropertyChangeObservable()` call in a test.

**Compile level, `Namotion.Interceptor.Registry`.** The expression overload `TryGetRegisteredProperty<T>(Expression<Func<T, object?>>)` becomes `TryGetRegisteredProperty<T, TValue>(Expression<Func<T, TValue>>)`. The added generic parameter is binary breaking, and source breaking for callers that pass explicit type arguments or a pre-built `Expression<Func<T, object?>>`; lambda call sites still infer unchanged. The overload also gained null-safe registry-walked resolution for chained selectors, with index, dictionary key and method segments compiled and evaluated against the object graph.

**Behavioral.** These compile but act differently:

- Vetoed writes no longer publish. An interceptor that returns without calling `next` previously still produced a notification carrying a value that was never stored.
- Notifications dispatch after lifecycle reconciliation rather than before it, so for subject typed writes: writes a callback makes to a departing subject are no longer tracked (it is already detached, which is intended); a lifecycle handler that writes properties while attaching now emits those changes before the structural change that introduced the subject; and a lifecycle handler that throws suppresses that write's notification, though handlers are already required to be exception free. Value typed writes are unaffected.
- Null arguments to the subscribe entry points now throw `ArgumentNullException` at the call site instead of surfacing a `NullReferenceException` later on a writer thread.
- `GetPropertyChangeObservable()` returns the interceptor through `AsObservable()` instead of `Publish().RefCount()`. Delivery is equal or stronger (a second observer on one returned instance now gets the delivery guarantee it previously lacked), but the returned object's runtime type changed.
- `Namotion.Interceptor.Registry`: `TryGetRegisteredProperty(subject, propertyName, registry)` returns `null` when the context has no registry, instead of throwing `InvalidOperationException`.

Of these, only the attach time handler write ordering can affect code that follows every documented rule; the rest either fix a defect or tighten an already documented requirement.

**Added:** `PropertyChangeInterceptor`, `WithPropertyChangeSubscriptions()`, `IPropertyChangeObserver`, `PropertyChangeCallback`, and the `Subscribe` / `SubscribeToProperty` extension methods. The core `Namotion.Interceptor` package gains one internal per write dedup flag and no public surface.

## Testing

- Full unit suite green (Tracking at 407); the merge is behavior-preserving for existing observable, queue, transaction, and source tests.
- New coverage: dual-channel dispatch, gate closure and re-closure after churn, the three queue race fixes, per-property dispatch and exactly-once dedup under aggregation, dormancy and revival, the install-versus-write race for all three channels (deterministic via a blocking interceptor), vetoed-write suppression, null-subscribe guards, foreign-data-key fail-loud, the queue drain-after-disposal and cancellation-priority contracts, the resolved write-chain order (including aggregated contexts) and the lifecycle-ordering consequences, dispatch-versus-dispose safety, CAS subscribe/dispose/write races, the strict resolver negatives, derived-property firing and the derived-detection prerequisite, and transaction commit-replay visibility including asymmetric aggregation.
- Public API snapshot updated.

## Performance: static cost model

The two delivery interceptors (`PropertyChangeObservable`, `PropertyChangeQueue`) merge into one `PropertyChangeInterceptor`, so the write chain is one interceptor frame shorter. Reads are not affected: this is a write-only interceptor. What each write path does, versus master:

| Write path | Change vs master | Direction |
|---|---|---|
| Same-value write (no-op) | none; the `[RunsFirst]` equality check stops it before this interceptor | unchanged |
| Committed value write, no consumers | adds a post-commit memory fence plus a consumer-count and dispatch-state re-read; removes one interceptor frame | slightly higher |
| Committed write, one channel active | payload built once; master ran the idle channel as a pass-through frame | same or lower |
| Committed write, both channels active | payload built once instead of twice (one old-value capture, one `GetFinalValue`, one `Create`); fewer reference-typed allocations | lower, fewer allocations |
| Property read | not intercepted | unchanged |

Per-property subscriptions are new, with no master equivalent:

| Subscription state | Added cost per write |
|---|---|
| Listener on the written property | one dictionary lookup plus array dispatch |
| Any subscription elsewhere in the process | one count-gated lookup miss |
| After all subscriptions disposed | none; the write returns to the idle path |

The one cost the branch adds to the common write path is a **post-commit full memory fence** (a Dekker / StoreLoad pairing with the subscribe side, in `ResolveListeners`) that runs on every committed value-changing write, even with zero consumers. It buys the guarantee that a subscription installed while a write is in flight is never missed, and it is partly shadowed by the terminal write's existing `Interlocked.Exchange` on the write timestamp, which already drains the store buffer before this code runs. Everything else the branch does to the write path is neutral or cheaper: one fewer interceptor frame, and a single payload build once any consumer is active.

Because the fence sits on the committed-write path, the added cost is isolated there: same-value writes (stopped earlier) and reads (not intercepted) cross no changed code and act as measurement controls. Measured numbers are in the [benchmark comment](https://github.com/RicoSuter/Namotion.Interceptor/pull/377#issuecomment-5022333187).

## Deferred follow-ups

- #378: expression resolver consolidation plus recorded future options (asymmetric process-wide fence design with break-even analysis, per-subject listener gating, dedicated listener field).
- Three harder concurrency and lifetime tests: GC rootless-cycle collection, shared-observer stress, and best-effort commit-failure apply-and-revert visibility.
- Benchmark hardening polish (drain-loop cancellation guard, setup-method naming).


src/Namotion.Interceptor.Tracking/Transactions/SubjectTransactionInterceptor.cs
```diff
@@ -11,8 +11,6 @@ namespace Namotion.Interceptor.Tracking.Transactions;
 /// Also manages the per-context transaction lock for serialized transactions.
 /// </summary>
 [RunsBefore(typeof(DerivedPropertyChangeHandler))]
-[RunsBefore(typeof(PropertyChangeObservable))]
-[RunsBefore(typeof(PropertyChangeQueue))]
```
RicoSuter, line 15: Dont we need

[RunsBefore(typeof(PropertyChangeInterceptor))]

here?

src/Namotion.Interceptor.Tracking/Change/PropertyChangeSubscriptions.cs
```diff
@@ -0,0 +1,26 @@
+namespace Namotion.Interceptor.Tracking.Change;
+
+/// <summary>
+/// Process-wide gate for the per-property listener lookup. Counts subscriptions created and not yet
+/// disposed. It is a long (not int) so cumulative fire-and-forget increments (which never decrement)
+/// cannot wrap to zero and spuriously close the gate; 2^63 registrations is unreachable.
+/// </summary>
+internal static class PropertyChangeSubscriptions
+{
+    private static long _liveCount;
```
RicoSuter: should we rename livecount => subscriptioncount here?

src/Namotion.Interceptor/Interceptors/IWriteInterceptor.cs
```diff
@@ -39,6 +39,13 @@ public struct PropertyWriteContext<TProperty>
     // trigger's already-resolved value.
     private long _writeTimestamp;
 
+    /// <summary>
+    /// Set by the first PropertyChangeInterceptor instance that dispatches per-property listeners for this
+    /// write, so deeper interceptor instances in an aggregated fallback chain do not double-deliver.
+    /// Internal: Tracking sets it; no external interceptor may.
+    /// </summary>
+    internal bool ArePropertyListenersClaimed;
```
RicoSuter: maybe rename to something better, eg is property listeners published (find better name?) also compact the xml docs, it's clear it's internal from the internal keyword

src/Namotion.Interceptor.Tracking/Change/PropertyChangeSubscriptionExtensions.cs
```diff
@@ -0,0 +1,100 @@
...
+    {
+        return subject.SubscribeToProperty(propertySelector, new DelegateObserver(callback));
+    }
+
+    private static string ResolveDirectPropertyName<TSubject, TValue>(Expression<Func<TSubject, TValue>> propertySelector)
+    {
+        var body = propertySelector.Body;
+
+        // Unwrap a Convert/ConvertChecked boxing or numeric-cast node (e.g. Expression<Func<T, object>>).
+        while (body is UnaryExpression { NodeType: ExpressionType.Convert or ExpressionType.ConvertChecked } unary)
```
RicoSuter, line 89: dont we have this code already in another expression helper method we could reuse? or is it not the same?

src/Namotion.Interceptor.Tracking/Change/PropertyChangeQueueSubscription.cs
```diff
@@ -33,30 +31,27 @@ internal void Enqueue(in SubjectPropertyChange item)
...
-    /// Should only be called from a single consumer thread per subscription (not thread-safe for concurrent TryDequeue calls).
-    /// </summary>
-    /// <param name="item">The dequeued property change if available.</param>
-    /// <param name="cancellationToken">Cancellation token to abort the wait.</param>
-    /// <returns>True if an item was dequeued; false if the subscription is completed or cancelled.</returns>
     public bool TryDequeue(out SubjectPropertyChange item, CancellationToken cancellationToken)
     {
         while (true)
         {
-            // Check cancellation before dequeuing so that kill/shutdown signals
```
RicoSuter, line 51: why remove comments everywhere here?

src/Namotion.Interceptor.Tracking/Change/PropertyChangeQueueSubscription.cs
```diff
@@ -4,27 +4,25 @@
...
+    internal PropertyChangeQueueSubscription(PropertyChangeInterceptor interceptor)
     {
-        _changeQueue = queue;
+        _interceptor = interceptor;
     }
 
-    /// <summary>
-    /// Enqueues a property change. Thread-safe and can be called concurrently from multiple threads.
-    /// </summary>
-    /// <param name="item">The property change to enqueue.</param>
```
RicoSuter, line 27: why remove xml docs in whole class? not necessary? we should try to keep diff as minimal as possible and only change code when necessary (incorrect, needed)

src/Namotion.Interceptor.Tracking/Change/PropertyChangeInterceptor.cs
```diff
@@ -0,0 +1,270 @@
...
+            {
+                return; // one-shot
+            }
+
+            _inner.Dispose();
+            owner.RemoveObservableConsumer();
+        }
+    }
+
+    // ----- Write path -----
```
RicoSuter: remove these --- comments

docs/tracking-transactions.md
```diff
@@ -522,7 +522,7 @@ Note that concurrent `CommitAsync` calls on the same transaction are rejected: o
 
 1. **Always use `using` blocks** - Ensures proper disposal even on exceptions
 2. **Keep transactions short** - Long transactions hold pending changes in memory
-3. **Register transactions before notifications** - Call `WithTransactions()` before `WithPropertyChangeObservable()`
+3. **Register transactions before notifications** - Call `WithTransactions()` before `WithPropertyChangeNotifications()`
```
RicoSuter: with the runbefore etc. i'm not sure whether this is even needed?

docs/tracking.md
```diff
@@ -1,6 +1,6 @@
 # Tracking
 
-The `Namotion.Interceptor.Tracking` package provides comprehensive change tracking for interceptor subjects, including property value changes, derived property updates, subject lifecycle events, and parent-child relationships. It offers two mechanisms for observing changes: **Observable** (Rx-based) and the **(high performance) queue**, with the queue being the preferred choice for high-throughput scenarios.
+The `Namotion.Interceptor.Tracking` package provides comprehensive change tracking for interceptor subjects, including property value changes, derived property updates, subject lifecycle events, and parent-child relationships. A single `PropertyChangeInterceptor`, enabled with `WithPropertyChangeNotifications()`, delivers every property change through three facets that share one write path: an **Rx observable** for composition and UI, a **high-performance queue** for high-throughput consumers, and **per-property subscriptions** for observing one property on one subject instance.
```
RicoSuter: now i question whether WithPropertyChangeNotifications is really a good name as the interceptor is called PropertyChangeInterceptor, we never use the word Notifications again, so wondering whether the registration method should really use that name?

docs/tracking.md
```diff
@@ -52,25 +51,25 @@ var person = new Person(context)
...
+**Observable limitations:**
 - Higher memory overhead per change event
 - Slightly lower throughput in high-frequency scenarios
 - Subject synchronization overhead
 
-### Property Changed Queue (High Performance)
+### Property Change Queue (High Performance)
 
-The queue approach uses a lock-free, allocation-conscious queue and is optimized for maximum throughput with minimal allocations. This is the preferred mechanism for high-performance scenarios such as background services, IoT data processing, and source synchronization:
+The queue facet uses a lock-free, allocation-conscious queue and is optimized for maximum throughput with minimal allocations. This is the preferred mechanism for high-performance scenarios such as background services, IoT data processing, and source synchronization:
```
RicoSuter: not sure about the word "facet", seems uncommon, no? if we rename we should search all occurences of it in repo

src/Namotion.Interceptor.Tracking/InterceptorSubjectContextExtensions.cs
```diff
@@ -108,27 +112,19 @@ public static IObservable<SubjectPropertyChange> GetPropertyChangeObservable(thi
...
+    /// Under concurrent writes to the same property, changes may be enqueued out of commit order because
+    /// dispatch runs outside the subject lock; if you need the current value, re-read the property rather
+    /// than relying on the delivered new value.
     /// </summary>
     /// <param name="context">The context.</param>
-    /// <param name="scheduler">The scheduler to run the callbacks on (defaults to Scheduler.Default).</param>
-    /// <returns>The observable.</returns>
+    /// <param name="scheduler">Unused. Retained for signature compatibility; the pull-based queue has no scheduler.</param>
+    /// <returns>The queue subscription.</returns>
     public static PropertyChangeQueueSubscription CreatePropertyChangeQueueSubscription(this IInterceptorSubjectContext context, IScheduler? scheduler = null)
```
RicoSuter: remove IScheduler? scheduler = null, it was already unused before, right?

src/Namotion.Interceptor.Tracking/Change/PropertyChangeSubscriptions.cs
RicoSuter: Done in 2ddea7e7: renamed to subscriptionCount / ReadSubscriptionCount.

src/Namotion.Interceptor/Interceptors/IWriteInterceptor.cs
```diff
@@ -39,6 +39,9 @@ public struct PropertyWriteContext<TProperty>
     // trigger's already-resolved value.
     private long _writeTimestamp;
 
+    // Set by the PropertyChangeInterceptor instance that publishes this write's per-property listeners, so deeper aggregated instances do not double-deliver.
+    internal bool ArePropertyListenersPublished;
```
RicoSuter: also dont like ArePropertyListenersPublished, becuase it does not really publish listeners, right? better name?

src/Namotion.Interceptor.Tracking/Change/PropertyChangeSubscriptionExtensions.cs
```diff
@@ -0,0 +1,100 @@
...
+    /// Under concurrent writes to the same property, notifications may arrive out of commit order because
+    /// dispatch runs outside the subject lock; if you need the current value, re-read the property rather
+    /// than relying on the delivered new value.
+    /// A write that commits after this subscription is installed is always delivered; a write that committed
+    /// before the install may not be, and reading the property after subscribing observes that earlier state;
+    /// a write that raced the install may be delivered with OldValue equal to NewValue.
+    /// </remarks>
+    public static IDisposable Subscribe(this PropertyReference property, IPropertyChangeObserver observer)
+    {
+        if (!property.Subject.Properties.TryGetValue(property.Name, out var metadata)
```
RicoSuter: why not property.Metdata?

src/Namotion.Interceptor.Tracking/Change/PropertyChangeSubscriptionExtensions.cs
```diff
@@ -0,0 +1,100 @@
...
+            || !(metadata.IsIntercepted || metadata.IsDerived))
+        {
+            throw new ArgumentException(
+                $"Property '{property.Name}' on {property.Subject.GetType().Name} cannot be subscribed to: it is not an intercepted or derived property, so its changes never enter the interception chain.",
+                nameof(property));
+        }
+
+        return PropertyChangeSubscription.Install(property, observer);
```
RicoSuter: Install => Register or Create? or bad idea?

src/Namotion.Interceptor.Tracking/Change/PropertyChangeSubscription.cs
```diff
@@ -0,0 +1,132 @@
+namespace Namotion.Interceptor.Tracking.Change;
+
+/// <summary>
+/// A subject-stored per-property subscription. Stored in Subject.Data[(propertyName, ListenersKey)]
+/// as an element of an immutable copy-on-write array. Strong references (standard IDisposable
+/// ownership); dispose or drop the handle.
+/// </summary>
+internal sealed class PropertyChangeSubscription : IDisposable
+{
+    internal const string ListenersKey = "Namotion.Interceptor.PropertyChangeListeners";
```
RicoSuter: check other keys, maybe we should also use a short one for better perf?

src/Namotion.Interceptor.Tracking/Change/IPropertyChangeObserver.cs
```diff
@@ -0,0 +1,12 @@
...
+
+/// <summary>
+/// Synchronous per-property change observer. OnChange runs on the writing thread, inside the write,
+/// outside the subject lock. Implementations MUST be thread-safe (they may be invoked concurrently),
+/// fast, non-blocking, and MUST NOT throw. Deliveries may arrive out of commit order under concurrent
+/// writes to the same property; re-read the property if you need the current value.
+/// </summary>
+public interface IPropertyChangeObserver
+{
+    void OnChange(in SubjectPropertyChange change);
```
RicoSuter, line 15: add xml docs?

src/Namotion.Interceptor.Benchmark/PropertyChangeNotificationBenchmark.cs
```diff
@@ -0,0 +1,230 @@
...
+/// build-once win and of the zero-allocation idle and listener-elsewhere fast paths.
+///
+/// Each state runs in its own process (BenchmarkDotNet default), set up by a targeted
+/// <see cref="GlobalSetupAttribute"/>. This isolation matters: the per-property listener live count
+/// is a process-wide static, so a listener created for one state must not be visible to another.
+/// The context is registered with WithPropertyChangeSubscriptions only (no equality check), so every
+/// write reaches the interceptor even when the value does not change.
+/// </summary>
+[MemoryDiagnoser]
+public class PropertyChangeNotificationBenchmark
```
RicoSuter: should also rename to PropertyChangeSubscription(s)Benchmark

src/Namotion.Interceptor.Tracking/Transactions/SubjectTransactionInterceptor.cs
RicoSuter: Intentional, no change needed. The ordering is expressed target-side: `PropertyChangeInterceptor` carries `[RunsAfter(typeof(SubjectTransactionInterceptor))]`, which builds a per-instance edge and so orders the transaction interceptor before *every* aggregated `PropertyChangeInterceptor` instance. A source-side `[RunsBefore]` here would bind only the last instance per type (`ServiceOrderResolver` keeps one index per type), so it is the weaker form under aggregation. The transaction ordering tests verify this under both registration orders.

src/Namotion.Interceptor/Interceptors/IWriteInterceptor.cs
RicoSuter: Done in 2ddea7e7: renamed to `ArePropertyListenersPublished` and compacted the doc to a single line.

src/Namotion.Interceptor.Tracking/Change/PropertyChangeSubscriptionExtensions.cs
RicoSuter: Not a reusable helper here, unfortunately. There are four similar `x => x.Foo` resolvers (this one in Tracking, plus Registry, OpcUa and Mqtt), but all three others live in packages that depend on Tracking (Registry references Tracking; OpcUa/Mqtt are downstream) and core has none, so this one cannot reference them without a package cycle. It is also deliberately stricter than the connector ones (it rejects chained, captured, static and field selectors). Kept it Tracking-local; extracting one shared resolver into core that all four consume would be a good separate follow-up.

src/Namotion.Interceptor.Tracking/Change/PropertyChangeQueueSubscription.cs
RicoSuter: Restored in 2ddea7e7 (see the sibling thread): rebuilt from master keeping all docs and comments; the diff is now only the necessary changes.

src/Namotion.Interceptor.Tracking/Change/PropertyChangeQueueSubscription.cs
RicoSuter: Restored in 2ddea7e7: rebuilt the file from master, keeping all its XML docs and comments, so the diff is now only the necessary changes: the ctor becomes `internal (PropertyChangeInterceptor)`, a new `Complete()` for interceptor-side teardown, the `_completed` re-check after `Reset()` (lost-wakeup fix), and the one-shot atomic `Dispose` that no longer disposes `_signal` (enqueue-vs-dispose fix).

src/Namotion.Interceptor.Tracking/Change/PropertyChangeInterceptor.cs
RicoSuter: Done in 2ddea7e7: removed the three `// -----` divider comments.

docs/tracking-transactions.md
RicoSuter: Done in 587c772a: deleted the item. Registration order is irrelevant now that ordering is enforced by the `[RunsAfter]` attribute, and the actual behavior (notifications suppressed during capture, fired at commit) is already documented above in the "Notification suppression" bullet and the commit-flow section.

docs/tracking.md
RicoSuter: Done in 587c772a: renamed `WithPropertyChangeNotifications` to `WithPropertyChangeSubscriptions` everywhere. It matches the existing `*Subscription` type names and covers all three mechanisms (observable `Subscribe`, queue subscription, per-property `Subscribe`) without underselling the queue.

docs/tracking.md
RicoSuter: Done in 587c772a: renamed `facet` to `channel` across the five files (docs, interceptor, extensions, benchmark, test).

src/Namotion.Interceptor.Tracking/InterceptorSubjectContextExtensions.cs
RicoSuter: Done in 053f717f: removed the unused `scheduler` parameter (and its doc line); snapshot updated.

src/Namotion.Interceptor/Interceptors/IWriteInterceptor.cs
RicoSuter: Done in 053f717f: renamed to `ArePropertyObserversNotified`.

src/Namotion.Interceptor.Tracking/Change/PropertyChangeSubscriptionExtensions.cs
RicoSuter: Done in 053f717f: now uses `property.Metadata` directly. An unknown property name throws `InvalidOperationException` (from `Metadata`) and a non-trackable property throws `ArgumentException`; retargeted the unknown-member test accordingly.

src/Namotion.Interceptor.Tracking/Change/PropertyChangeSubscriptionExtensions.cs
RicoSuter: Done in 053f717f: renamed `Install` to `Create`.

src/Namotion.Interceptor.Tracking/Change/PropertyChangeSubscription.cs
RicoSuter: Done in 053f717f: `ListenersKey` value is now the short `ni.pcl`, matching the `ni.dpd` convention.

src/Namotion.Interceptor.Tracking/Change/IPropertyChangeObserver.cs
RicoSuter: Done in 053f717f: added `<summary>` and `<param>` to `OnChange`.

src/Namotion.Interceptor.Benchmark/PropertyChangeNotificationBenchmark.cs
RicoSuter: Done in 053f717f: renamed the class and file to `PropertyChangeSubscriptionsBenchmark`.

RicoSuter:
## Final benchmarks (updated with a memoryRandomization pass): CPU-pinned, `666bb0d9` vs master `56e382fc`

**This revises my earlier verdict.** I previously reported a confirmed +3 to +5% write-path regression. A `--memoryRandomization` pass shows that delta is not separable from per-binary layout bias on this machine, so the honest conclusion is: no reliably-measurable regression. Details below. (The two benchmark comments above this one are superseded: they used a stale local master and a CPU throttled to 0.80 GHz.)

### Methodology

CPU pinned to a fixed 3.60 GHz (performance governor, turbo disabled), verified holding 3.6 GHz under load. Local master synced to `56e382fc`. Full suites at LaunchCount 3, an LC15 confirmation of the methods that moved, and an LC15 `--memoryRandomization` pass over the write and read methods to separate real cost from code/data layout.

### Write and read deltas do not survive as a real regression

| Method | non-MR Delta | MR Delta | Crosses changed code? |
|---|---:|---:|---|
| WriteWithTimestampScope | +5.3% | +2.4% | yes (post-commit fence) |
| Write | +3.6% | +2.7% | yes (post-commit fence) |
| Read | +3.1% | +2.5% | no (control) |
| WriteNoOp | +1.5% | -1.1% | no (control, stopped before the interceptor) |
| ReadParents | +1.7% | -0.4% | no (control) |

Reading this: memoryRandomization peeled a layout component off the write rows (`WriteWithTimestampScope` fell from +5.3% to +2.4%), but the residual write deltas (+2.4% to +2.7%) land right at the top of the unchanged-control spread (-1.1% to +2.5%). Crucially, `Read`, which the PR never touches (this is a write-only interceptor), also sits at +2.5% and did not collapse. So the branch's larger binary carries a stable few-percent delta even on code it does not change, and pinning plus randomization do not remove it. If a real fence cost existed on top of that floor, `Write` would exceed `Read`; it does not. The write-path cost is therefore not separable from layout bias here.

### Conclusion

- **No reliably-measurable hot-path regression.** What looked like +3 to +5% earlier was mostly memory and binary layout, with the residual indistinguishable from the same-size delta on an unchanged read.
- **Allocations: byte-identical on every method.** No allocation regression anywhere.
- **New subscriptions feature: well-behaved.** Listener on another property 0.98x an idle write (free, allocation-free), back to 1.00x baseline after unsubscribe, and 48 B only on the two boxed-observable paths.

**Merge-safe on performance.**

### On the redundant write-side fence

The `Interlocked.MemoryBarrier()` in `ResolveListeners` is provably redundant with the terminal write's existing `SetWriteTimestamp` -> `Interlocked.Exchange` (a full fence on every committing write, which runs before the post-commit dispatch). But this data does not promise a measurable recovery from removing it, and there is no cleanly-measurable fence cost on this hardware to recover.

**Update (tested and reverted).** The barrier removal was implemented and measured directly, isolated with-fence vs without-fence on the same commit, pinned: no measurable impact (`Write` +0.4%, `WriteWithTimestampScope` -0.7%, `WriteNoOp` control +0.2%, all inside the noise band). The `Interlocked.Exchange` fully shadows it, as predicted. It was reverted: beyond the null result, the removal is not universally safe. A committed write through a custom terminal or a direct `IsWritten = true` (both public API) bypasses the terminal `Exchange`, and the `BlockingWriteInterceptor` race tests establish their own happens-before via `ManualResetEventSlim`, so they would not catch a weakened fence. The explicit barrier stays.

Environment: i7-11700K pinned at 3.60 GHz (no turbo, performance governor), .NET SDK 10.0.100, host .NET 9.0.11, BenchmarkDotNet 0.15.5. The BDN header prints a low `Max:` value; that is an idle-frequency snapshot at process start, not the measurement frequency, which was a verified stable 3.6 GHz.

PR #381, opened 2026-07-19
Add expression-based path subscriptions (SubscribeToPath)

RicoSuter:
Adds **expression-based path subscriptions** (`SubscribeToPath`) to `Namotion.Interceptor.Tracking`: observe "the value at `x.Foo.Bar[3].Baz`" while the path itself changes over time. `Foo` may be null and appear later, `Bar` may be replaced with a collection holding a different subject at index 3, and the leaf value changes. The subscriber observes the current state of the path (resolved with a value, or unresolved) and receives one event per observable transition, each carrying the real property write that caused it.

This PR now contains the **full implementation** (tracker, tests, benchmarks, and docs), not only the design document. It is based directly on `master`: the per-property change subscription primitive it composes is already merged (#377), so this branch is release-safe on its own. The public surface is purely additive (seven new public types under `Namotion.Interceptor.Tracking.Paths`); no core API changed and there is no `Namotion.Interceptor.Registry` dependency. The design and implementation went through several independent adversarial review rounds; the carve-outs and edge cases below reflect that.

## What is included

- **Production**: the new `Paths/` folder in `Namotion.Interceptor.Tracking` (expression decomposer, never-throwing validating walker, typed and lenient value accessors, a per-segment accessor cache for the steady-state walk, and the `SubjectPathSubscription` tracker with subscribe-before-read chain build, divergent retrack, exclusive-drain delivery, thread-affine reentrancy guard, dispose contract, and transaction ambient-suppression).
- **Tests**: expression validation, `Current`, transitions, concurrency (missed-write window, cyclic path, cross-path deadlock-freedom, reentrancy), dispose and teardown-on-throw, transactions (including the retrack-stranding regression), dormancy and the five carve-outs, the property-type matrix, the chain invariant, cause/provenance, and fire-and-forget collection.
- **Benchmarks**: `SubjectPathSubscriptionBenchmark` (path delivery versus a bare per-property listener at depths 1 and 4, the hot-leaf suppressed case, structural retrack, `Current` read, subscribe/dispose churn, and a deep mixed path), with allocation-free delivery gated by `[MemoryDiagnoser]`. Needs the `requires-benchmark` label to run in CI.
- **Docs**: a slimmed consumer-facing Path Subscriptions section in `docs/tracking.md`, plus a permanent internal design doc at `docs/design/tracking-path-subscriptions.md` (mechanism, concurrency model, dispose/transaction internals, the five quiescent-consistency carve-outs, and the escalation path).

## API

```csharp
var subscription = root.SubscribeToPath(x => x.Foo.Bar[3].Baz, (in SubjectPathChange<double> change) =>
{
    if (change.NewState.TryGetValue(out var value)) { ... } else { /* path is unresolved */ }
}, SubjectPathValidation.Full);

var current = subscription.Current; // SubjectPathValue<double>: IsResolved + Value, always fresh
```

- `SubjectPathChange<TValue>` carries `Kind` (`ValueChange` or `PathChange`), the observed transition `OldState`/`NewState` (each a `SubjectPathValue<TValue>` with `IsResolved` and `Value`), and `Cause` (the unmodified `SubjectPropertyChange` of the triggering write).
- Typed expression paths only: property access, collection index, dictionary key. Static shape defects (casts, field selectors, multi-argument indexers, captured-object chains, and similar) throw at subscribe with clear errors; a segment that is not resolvable or not intercepted at runtime resolves to unresolved rather than throwing. Index arguments are evaluated once at subscribe time.
- No initial callback; the handle's side-effect-free `Current` serves the initial state.
- A required `SubjectPathValidation` argument chooses how a leaf write is validated: `Full` re-validates the whole path from the root on every callback (so a divergence created while a segment was dormant heals on the next callback of any kind), while `LeafOnly` re-reads only the leaf on a leaf write and skips the from-root walk (a divergence created while dormant then heals on a later structural write, not on the leaf write). Structural writes always re-validate and retrack in both modes, and `Current` is always a fresh full walk. `LeafOnly` cuts roughly half the tracker's per-delivery walk overhead on a deep path (benchmarked at depth 4: about 182 ns to 132 ns), so it is the choice for deep, high-frequency leaf-write paths that can accept the bounded dormancy carve-out; `Full` is the safe choice otherwise.

## Mechanism

- One per-property subscription per path segment on the currently resolved subject. Collections change only by reassignment in this library, so segment subscriptions capture all structural change; no Registry or lifecycle coupling (the feature compiles without a Registry reference).
- A structural change disposes and rebuilds only the suffix below the changed segment (subscribe-before-read per segment).
- Every processed callback revalidates the path with a fresh bounds-checked walk from the root and retracks from the first divergence, so even a callback from a subject that silently left the path heals the chain instead of delivering off-path state. `NewState` is a fresh re-read, which is what makes delivery converge regardless of dispatch order.
- Delivery is serialized per subscription through an exclusive-drain queue: events are computed and ordered under a small per-tracker lock, user callbacks run outside it. Per subscription, events are totally ordered and chained (event N+1's `OldState` equals event N's `NewState`).

## Guarantees

- **Missed-write closure**: the primitive resolves listeners post-commit behind a full fence, so a write committing after a segment subscription's install is always delivered; combined with subscribe-before-read the chain cannot keep watching a subject the graph has left (for observed writes).
- **Quiescent consistency**: once observed writes settle, the last event's `NewState` and `Current` agree with the real graph.
- **`Current` is always accurate**: an independent fresh walk, correct even mid-divergence, and it never throws.
- **Never throws** into the writer's chain or out of `Current`: every traversal failure (getter, container operation, type mismatch) resolves to unresolved.
- **Allocation-free delivery** for inline value types and strings, including a path with an `ImmutableArray<T>` intermediate segment (typed accessors, a reused per-subscription walk buffer, and an `ArrayPool`-backed `Current`). A non-`IEquatable<T>` struct leaf is the exception, since it boxes in the suppression comparison.

## Where it is not guaranteed (five documented carve-outs)

Quiescent consistency holds except in these cases; in all five `Current` still reflects truth and the event stream heals on the next delivered segment callback:

1. Throwing callback (abandons the drain).
2. Dormant divergence (a change while a segment subject is outside every notifying context).
3. A foreign throwing synchronous observer of the same write aborting dispatch.
4. A getter or container operation throwing during the walk.
5. An `Equals`-suppressed derived subject-valued intermediate (a distinct-but-equal recompute the equality handler drops).

It is deliberately **not a write log**: concurrent writes coalesce (intermediate values may be skipped). For per-write completeness use the per-property or queue channel.

## Transactions

`SubscribeToPath` throws inside an active non-committing transaction (its build walk would read staged values and strand the chain on rollback). The event-computation walk reads the committed graph (suppresses an active ambient transaction) so it never retracts from staged values. `Current` inside a transaction returns that transaction's read-your-writes view (benign, no persistent state).

## Cost

A path subscription costs nothing on writes outside the watched path beyond the primitive's active-lookup state. A write to a watched segment pays, inline on the writing thread, a short per-subscription lock, a validating walk (one property read per segment), and the callback; structural changes additionally pay re-subscription of the segments below the change. Leaf delivery is allocation-free for inline value types and strings. Benchmark-gated: idle write, listener-elsewhere, leaf at depths 1 and 4 including the hot-leaf suppressed case, structural retrack, `Current` read, subscribe/dispose churn, and a deep mixed path.

## Open tasks (this PR)

- [x] Removed the temporary superpowers design spec now that review is through; its content lives in `docs/design/tracking-path-subscriptions.md` and `docs/tracking.md` (the firehose-rejected rationale it carried was migrated into the design doc Overview).
- [x] Add the `requires-benchmark` label so the benchmark job runs in CI.
- [x] Performance: cache each segment's resolved `(subject, accessor)` binding so the steady-state walk skips the per-walk `Properties` lookup and accessor-cache probe, falling back to by-name resolution only on a subject mismatch (the divergence signal). Benchmarked before merging: delivery drops 15 to 23 percent and `Current` reads 35 to 49 percent across depths and leaf types, delivery stays allocation-free, and the only cost is a small allocation on the rare structural-retrack and subscribe/dispose paths. Included in this PR.

## Follow-ups (out of scope here)

- **#385**: commit-order sequence numbers for the change channels (consumer-side reordering) and optional per-channel FIFO delivery.
- **#387**: dynamic (runtime-named) path segments (`Dynamic<T>` / `Dynamic()` / string paths) and late-added-property discovery. The tracker's segment-list contract is kept decoupled from the expression decomposer so this composes without reworking the walker.
- Other future extensions documented in the spec: prefix sharing, cast support in selectors, paths ending in an indexed element, in-place collection mutation tracking, asynchronous delivery, and a `Refresh()` re-sync hook.


RicoSuter:
Question: Does a leaf change event also trigger a read of every segment of the path? Should not probably as it will not change the subscription, right?

src/Namotion.Interceptor.Tracking/Paths/SubjectPathChange.cs
```diff
@@ -0,0 +1,30 @@
...
+        Kind = kind;
+        Old = old;
+        New = @new;
+        Cause = cause;
+    }
+
+    public SubjectPathChangeKind Kind { get; }
+
+    /// <summary>Observed state before this event.</summary>
+    public SubjectPathValue<TValue> Old { get; }
```
RicoSuter: should we rename to OldValue and NewValue? Old/new seems a bit off, no?

src/Namotion.Interceptor.Tracking/Paths/SubjectPathSubscription.cs
```diff
@@ -0,0 +1,509 @@
+using System;
```
RicoSuter, line 1: Could we do a focused structural refactor of these two classes without using partial classes or arbitrary file splits?

  - Extract the type-reflection helpers from PathExpressionDecomposer into an internal PathTypeResolver.
  - Split SubjectPathSubscription into cohesive collaborators. For example, a PathSubscriptionChain<TValue> could own segment handles, resolved subjects/accessors,
    building, retracking, and suffix disposal. A separate delivery component could own the pending queue and exclusive callback drain.

  - Extract the remaining ProcessSegmentCallback protocol into named steps such as TryComputeEvent, DrainCallbacks, and optionally an ambient-transaction suppression
    scope.

  The subscription should remain the coordinator for locking, reentrancy, disposal, and transaction ordering. These invariants are load-bearing, so their comments and
  ordering should move with the extracted code unchanged. This would reduce both class and method size while making the concurrency protocol easier to verify.

src/Namotion.Interceptor.Tracking/Paths/SubjectPathChange.cs
RicoSuter: Done in 520a7bbd, renamed to `OldState`/`NewState`. I went with `...State` rather than `...Value` because both are `SubjectPathValue<TValue>` observed states that can be unresolved (the leaf value lives at `.Value` inside them), so `OldValue`/`NewValue` would wrongly imply they are the raw values. The public API snapshot, docs, and tests are updated.

src/Namotion.Interceptor.Tracking/Paths/SubjectPathSubscription.cs
RicoSuter: Done across two commits.

- 520a7bbd extracts the type-reflection helpers into an internal `PathTypeResolver`.
- fa1aa169 splits `SubjectPathSubscription` into `PathSubscriptionChain<TValue>` (segment handles, resolved subjects and accessors, build, retrack, suffix disposal) and `PathDeliveryQueue<TValue>` (the pending queue plus the exclusive callback drain), extracts `PathSegmentObserver`, and decomposes `ProcessSegmentCallback` into `TryComputeEvent`, the deferred-revalidation loop, and an `AmbientTransactionScope`.

The subscription stays the sole coordinator for locking, reentrancy, disposal, and transaction ordering, and the load-bearing invariant comments moved with the code unchanged. No public-API change (the snapshot is unchanged), all tests green, and an independent review confirmed every concurrency invariant is preserved with net-better verifiability. A quick benchmark confirmed no hot-path regression from the collaborator indirection.

src/Namotion.Interceptor.Tracking/Paths/PathTypeResolver.cs
```diff
@@ -0,0 +1,119 @@
...
+            }
+        }
+
+        var listInterface =
+            FindClosedInterface(collectionType, typeof(IList<>)) ??
+            FindClosedInterface(collectionType, typeof(IReadOnlyList<>));
+
+        // No generic list interface: not a subscribe-time error (runtime validity is lenient); both
+        // types stay null so the walk indexes the boxed collection via IndexReferenceCollection.
+        return listInterface is null
```
RicoSuter, line 63: This fallback does not preserve the indexer semantics represented by the expression. For an indexable IEnumerable<T> that implements neither IList<T> nor IReadOnlyList<T>, x => x.Items[0].Name can bind a custom get_Item that returns one element while the walker enumerates and watches another. I reproduced this with a reverse-indexing collection: the expression selected the last element, but Current returned the first enumerated element. Please either compile/store the actual indexer or reject this shape instead of silently traversing a different child.

src/Namotion.Interceptor.Tracking/Paths/SubjectPathValue.cs
```diff
@@ -0,0 +1,52 @@
...
+
+    /// <summary>Suppression equivalence: same resolvedness and, when both resolved, values equal under <see cref="EqualityComparer{TValue}.Default"/>.</summary>
+    internal static bool AreEquivalent(in SubjectPathValue<TValue> a, in SubjectPathValue<TValue> b)
+    {
+        if (a.IsResolved != b.IsResolved)
+        {
+            return false;
+        }
+
+        return !a.IsResolved || EqualityComparer<TValue>.Default.Equals(a._value, b._value);
```
RicoSuter, line 50: EqualityComparer<TValue>.Default.Equals executes arbitrary user equality code under the subscription lock, and an exception here is not contained. With a context using WithPropertyChangeSubscriptions, I reproduced a custom IEquatable<T>.Equals causing a leaf assignment to commit and then throw from the path subscription machinery. That conflicts with the never-throws delivery guarantee. Please contain comparison failures, for example by treating them as non-equivalent, so path bookkeeping cannot escape into the writer.

RicoSuter:
Input from a downstream consumer, in case it is useful while implementing this.

We are building reactive machine conditions over a subject graph, and this PR is exactly the primitive we want: the value at `x.Unit.Inner.IsManualMode`, still correct while intermediates are attached late, replaced or removed. We are hand-rolling per-hop subscriptions in the meantime and would drop them for this.

One addition would let us delete our resolution code entirely rather than keep half of it: **expose the resolved leaf `PropertyReference` on `SubjectPathValue<TValue>`**, non-null exactly when `IsResolved`, so it fits the existing tri-state rather than widening it.

Reasons it seems worth having:

- The tracker already resolves the leaf to read its value and to install the per-property subscription, so this surfaces state it is holding anyway.
- `PropertyReference` lives in the core assembly, so it does not invert the `Registry -> Tracking` dependency direction. `RegisteredSubjectProperty` would, which is presumably why the API is value-only today, but the reference is the layer-safe half: a consumer calls `TryGetRegisteredProperty()` on it from the Registry layer.
- Without it, any consumer that needs metadata rather than the bare value has to resolve the path a second time itself, which largely defeats the subscription. For us that is the leaf's `IsAvailableComputed` attribute and the source that owns it (`GetSourceState()`), but the same applies to units, minimum/maximum and permissions.

Two things we do not think belong in this PR, for the record:

- Watching a sibling **attribute** property of the leaf. That is a Registry-layer concern, and with the `PropertyReference` in hand a consumer can subscribe to it directly and re-establish that subscription on `PathChange`, which is where `SubjectPathChangeKind` already helps.
- Source synchronization state, which is a Connectors concern.

PR #397, opened 2026-07-31
Add batched OPC UA browse/read session primitives and split status classification

RicoSuter:
## Summary

First of three stacked PRs splitting #357. This one carries the protocol-level layer only, with no behavior change: nothing on master calls the new primitives yet, so it is purely additive.

- **`OpcUaSessionExtensions`** (internal): `BrowseNodesAsync` and `ReadNodesAsync` batch by the server's reported operation limits, follow BrowseNext continuation points under a bounded round cap, split and retry when the server rejects the batch size, and keep results positionally aligned with the request by padding short responses. Continuation points are released on every abort path so unfollowed pages do not leak server-side. `DistinctByResolvedNodeId` resolves and deduplicates browse references so consumers can rely on a non-null `BrowseName`.
- **`OpcUaStatusCodeClassifier`**: splits the single permanent-code set into `SessionPermanentCodes` and `LoadSkipCodes`. The access-scoped codes (`BadUserAccessDenied`, `BadNotReadable`, `BadNotImplemented`) answer the two questions oppositely. A monitored item must be retried because role permissions and `AccessLevel` can change mid-session, but a browse or read must skip the node rather than crash-loop the whole load. `IsTransientError` keeps its existing behavior and code set; `ThrowIfTransientError` and `IsBatchTooLarge` are new.
- **`OpcUaTransientServiceException`**: carries operation, node id, and status code so consumers can branch on transient load failures via `OpcUaClientDiagnostics.LastError`.

### Browse batches are capped by the continuation-point quota, not just the operation limit

`MaxNodesPerBrowse` is not the binding constraint for a browse that pages. The server's continuation-point quota is a separate and much smaller limit (SDK servers default to 10, against a `MaxNodesPerBrowse` of 2500), enforced both per Browse call and per session. Sizing a batch by the operation limit alone means holding more continuation points open than the server allows, which either fails the call with `BadNoContinuationPoints` or silently evicts the oldest points so the follow-up BrowseNext fails with `BadContinuationPointInvalid`. Neither is in `LoadSkipCodes`, so the load aborts, reconnect retries with the identical batch size, and it fails the same way: a loop that never converges.

Browse and BrowseNext batches are therefore sized at `min(MaxNodesPerBrowse, MaxBrowseContinuationPoints)`, which is what `Opc.Ua.Client.Browser` does for the same reason. Release calls still use the plain operation limit, since releasing frees continuation points rather than consuming them.

### Result contract

A NodeId present in the browse result was browsed to completion, possibly to an empty reference list. A NodeId that failed this round is omitted rather than reported with a truncated child list. That covers three cases: a permanent bad status on the first page, pagination stopping at the round cap, and a permanent bad status on a BrowseNext page. Consumers align collection children positionally, so a short list would replace a collection with the wrong contents instead of leaving it alone for the next load to retry.

## Public API

One additive type, `OpcUaTransientServiceException`. No removals, no signature changes. Snapshot updated.

## Test Plan

- [x] `dotnet build src/Namotion.Interceptor.slnx` clean
- [x] OPC UA test project: 313 passed
- [x] `OpcUaSessionExtensionsTests` cover the continuation-point quota cap, split-and-retry, continuation-point release on abort, both partial-pagination drops, multi-round paging, short-response padding, and transient-versus-permanent status handling
- [x] Public API snapshot accepted

## Stack

1. **This PR**: session primitives and status classification
2. #398: subscription callback gating and sweep ordering
3. #357: the four-phase loader itself

### DIFF
--- a/src/Namotion.Interceptor.OpcUa.Tests/Client/OpcUaSessionExtensionsTests.cs
+++ b/src/Namotion.Interceptor.OpcUa.Tests/Client/OpcUaSessionExtensionsTests.cs
@@ -1,0 +1,447 @@
+using Microsoft.Extensions.Logging.Abstractions;
+using Moq;
+using Namotion.Interceptor.OpcUa.Client;
+using Opc.Ua;
+using Opc.Ua.Client;
+
+namespace Namotion.Interceptor.OpcUa.Tests.Client;
+
+/// <summary>
+/// Tests for the bad-status handling in <see cref="OpcUaSessionExtensions"/>.
+/// Browse aborts on a transient per-NodeId status by throwing
+/// <see cref="OpcUaTransientServiceException"/> so the structural graph is never
+/// loaded incomplete; permanent statuses are logged and skipped. The read path is
+/// a best-effort primitive: it never classifies or throws, returning every result
+/// positionally so each caller applies its own policy (value loads keep the good
+/// values, type resolution aborts on transient itself).
+/// </summary>
+public class OpcUaSessionExtensionsTests
+{
+    [Fact]
+    public async Task WhenBrowseReturnsTransientBadStatus_ThenThrowsTransientServiceException()
+    {
+        // Arrange
+        var nodeId = new NodeId(2001, 2);
+        var mockSession = CreateMockSession();
+        mockSession
+            .Setup(s => s.BrowseAsync(
+                It.IsAny<RequestHeader>(),
+                It.IsAny<ViewDescription>(),
+                It.IsAny<uint>(),
+                It.IsAny<BrowseDescriptionCollection>(),
+                It.IsAny<CancellationToken>()))
+            .ReturnsAsync(new BrowseResponse
+            {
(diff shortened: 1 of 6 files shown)


RicoSuter:
Superseded by #414, which carries this branch's content as its second commit (`e6e2e1b0`).

Two changes were folded in there rather than landing this as-is:

1. The continuation-point cap is now conditional on `maxReferencesPerNode != 0`. Unconditional capping cost 7 of 16 browse round-trips on a representative address space, and the quota only matters when a browse can actually leave continuation points open.
2. `BadNoContinuationPoints` now halves the batch and retries. It was not in `LoadSkipCodes`, so it threw transient and reconnect retried the identical batch size forever.

The reason for consolidating rather than merging this first: this branch has no callers of its own, so the browse-completion contract cannot be reviewed without the consumer that depends on it, and the cap change above is motivated entirely by the loader's measurements.

PR #398, opened 2026-07-31
Gate OPC UA data-change callbacks during subscription setup and sweep before read-after-write

RicoSuter:
## Summary

Second of three stacked PRs splitting #357. Stacked on #397. Two ordering bugs in subscription setup, both of which let work reach a subject that is not ready for it.

**Callback gating.** Data-change notifications could be applied while subscription setup was still running, including during reconnection re-setup. A `_callbacksEnabled` flag now closes at the start of `CreateBatchedSubscriptionsAsync` and opens only as the final statement, so a notification can never write into a subject mid-setup. Closing it first also prevents an in-flight notification from passing on the previous setup's stale-true flag. It reopens only when the manager is not shutting down, so a `DisposeAsync` racing with setup cannot have its gate closure undone.

**Sweep before registration.** Read-after-write was registered per subscription inside the creation loop, so a subject that detached during setup could still be registered. Registration now happens once, after `SweepDetachedSubjects` has removed monitored items for subjects no longer in the registry, and only for handles that survived the sweep.

The sweep, the registration, and the gate opening are one method, `CompleteSetup`, because the ordering is the whole point and three statements at a call site invite reordering.

Also: `EscalatePersistentlyFailedItemsAsync` now skips items the sweep already dropped. They remain in the SDK subscription, so escalation could otherwise resurrect a detached subject's item into polling.

## Notes for review

The tests drive production code paths rather than parallel copies. `OnFastDataChangeForTesting` calls the real `FastDataChangeCallback` handler (it reads neither the subscription nor the string table, so there is nothing to fake), which means deleting the gate check in production fails the gating test. The ordering test calls `CompleteSetupForTesting`, so swapping the sweep and the registration in production fails it too, which was verified by making that swap.

`IReadAfterWriteRegistrar` is extracted purely so a recording spy can be injected in tests. Both it and `SubscriptionManager` are internal, so there is no public API change.

`SweepDetachedSubjects` enumerates the monitored-item dictionary directly instead of its `Values` property, which takes every bucket lock and allocates a copy, and checks its seen-set before the registry lookup so the lookup runs once per distinct subject rather than once per monitored item.

## Test Plan

- [x] `dotnet build src/Namotion.Interceptor.slnx` clean
- [x] OPC UA test project: 315 passed
- [x] New tests: `OpcUaSubscriptionCallbackGatingTests` (notification before the gate opens is dropped, applied after), `OpcUaSubscriptionSweepOrderingTests` (detached subject is swept and never registered)
- [x] Mutation-checked: reversing the sweep and registration order fails the ordering test

## Stack

1. #397: session primitives and status classification
2. **This PR**: subscription callback gating and sweep ordering
3. #357: the four-phase loader itself

### DIFF
--- a/src/Namotion.Interceptor.OpcUa.Tests/Client/OpcUaSubscriptionCallbackGatingTests.cs
+++ b/src/Namotion.Interceptor.OpcUa.Tests/Client/OpcUaSubscriptionCallbackGatingTests.cs
@@ -1,0 +1,27 @@
+namespace Namotion.Interceptor.OpcUa.Tests.Client;
+
+public class OpcUaSubscriptionCallbackGatingTests
+{
+    [Fact]
+    public void WhenDataChangeArrivesBeforeSetupCompletes_ThenNotificationIsIgnored()
+    {
+        // Arrange
+        var harness = SubscriptionManagerTestHarness.Create();
+        harness.RegisterMonitoredItem(clientHandle: 7, propertyName: "Value");
+
+        var timestamp = DateTimeOffset.UtcNow;
+
+        // Act - apply before gate is open (_callbacksEnabled is false)
+        harness.Manager.ApplyDataChange(7, timestamp, 42d);
+
+        // Assert - gate blocked the write
+        Assert.NotEqual(42d, harness.GetValue("Value"));
+
+        // Act - open the gate, then apply the same change
+        harness.Manager.EnableCallbacksForTesting();
+        harness.Manager.ApplyDataChange(7, timestamp, 42d);
+
+        // Assert - write is now applied
+        Assert.Equal(42d, harness.GetValue("Value"));
+    }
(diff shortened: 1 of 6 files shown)


RicoSuter:
Superseded by #414, which carries this branch's content as its third commit (`749ff8c2`).

One thing changed. The `CompleteSetup` guard here:

```csharp
// Never re-open a gate that DisposeAsync closed: it may have run concurrently with setup.
_callbacksEnabled = !_shuttingDown;
```

was dead as written, because `CreateBatchedSubscriptionsAsync` reset `_shuttingDown` to false earlier in the same method. That reset is a pre-existing bug on master (disposal is terminal, so the flag must be monotonic), and it is fixed in #414's first commit `9af09f53`, which lands before this one. The guard is live there.

PR #399, opened 2026-07-31
Monotonic per-subject commit revision on every property change

### LINKED ISSUE #385: Commit-order sequence numbers for property changes (and investigate FIFO channel delivery)
## Summary

Property change dispatch reorders relative to commit order under concurrency. Commits to a subject are already totally ordered by `SyncRoot`, but that order is discarded before dispatch. This issue proposes:

1. **Near-term (agreed): commit sequence number.** Stamp a monotonic sequence on each change so consumers can reorder and reconstruct a complete sequence without requiring in-order delivery.
2. **Investigate: optional FIFO in-order delivery per channel** (queue, observable, per-property, and the path observables built on them).

## Root cause

The terminal field write is the only thing under a lock (`src/Namotion.Interceptor/Cache/WriteInterceptorFactory.cs`):

```csharp
lock (context.Property.Subject.SyncRoot)
{
    innerWriteValue(subject, context.NewValue);
    context.IsWritten = true;
    context.FinalizeOrigin();
    context.Property.SetWriteTimestamp(...);
}
```

`SyncRoot` is scoped to the field write only and is released as the interceptor chain unwinds. `PropertyChangeInterceptor.WriteProperty` (`src/Namotion.Interceptor.Tracking/Change/PropertyChangeInterceptor.cs`) dispatches to all channels after `next()` returns, i.e. outside `SyncRoot`. Two threads writing the same property commit in `SyncRoot` order but can dispatch in the opposite order. That gap is the reordering window. The commit order exists; it is simply not carried to dispatch.

## Proposal 1 (near-term): commit sequence number

Stamp a monotonic sequence at commit time (inside the `SyncRoot` critical section, next to `SetWriteTimestamp`) and expose it on `SubjectPropertyChange`.

- **Granularity:** a per-subject counter is the cheapest (one increment while `SyncRoot` is already held, no new lock, no cross-core contention). It is dense over committed writes, since vetoed or no-op writes never reach the terminal write and so leave no gaps.
- **What it buys:** any consumer can sort by sequence to recover commit order (fixing same-property out-of-order for a current-value view without a re-read), and a full-stream consumer (queue channel) can detect completeness by observing a contiguous per-subject sequence.
- **Tradeoff to decide:** per-property completeness detection would need per-property sequences (more state); global ordering would need a shared atomic (contention). Per-subject is the recommended default.
- **Cost:** one increment under an existing lock, one field on the change struct (public API surface update), one counter slot per subject. Idle path unaffected.

## Proposal 2 (investigate): FIFO in-order delivery per channel

Order can only be captured inside the terminal write critical section, so this needs a core post-commit hook. Tracking alone cannot reach the ordered section.

**Granularity (per-property vs per-subject FIFO):** the goal is per-property FIFO (each property's writes delivered in commit order). Because `SyncRoot` orders all of a subject's commits together, the cheap mechanism naturally yields per-subject FIFO, which is strictly stronger than per-property (it also orders across a subject's properties). So per-property FIFO comes for free as a consequence, with no per-property state. Finer per-property coordinators would reduce cross-property serialization at dispatch, but since `SyncRoot` already serializes all of a subject's commits, there is little parallelism to recover, so per-subject is the recommended unit. Whether any workload justifies per-property granularity is an open investigation point.

- **Queue channel: cheapest.** Move the `Enqueue` into the `SyncRoot` critical section (via the hook); the consumer already pulls in enqueue order, so no new drain machinery is needed. Prototype and benchmark first.
- **Synchronous channels (per-property callbacks, `ImmediateScheduler` Rx):** callbacks cannot run under `SyncRoot`. FIFO needs enqueue-under-lock plus an exclusive drain outside the lock, with an uncontended inline fast path (the same pattern as the path-subscription drain). The uncontended common case stays near current performance; concurrent writes to the same subject pay a bounded extra cost (a record copy plus a drain-flag CAS). Scheduler-based Rx observers cannot be ordered, since the scheduler reorders downstream regardless.
- **Constraints found while investigating:**
  - Cannot hold `SyncRoot` across `next()`: nested `LifecycleInterceptor` acquires child subjects' `SyncRoot`s, so a parent-then-child ordering could deadlock a concurrent child-then-parent path.
  - Cannot build the full change under `SyncRoot`: `GetFinalValue()` runs derived getters, which `FinalizeOrigin` documents must not run under the lock. The drainer builds the change outside the lock, in commit order, so derived getters stay off the lock.
- **Per-subject state:** the coordinator (pending FIFO plus drain flag) is per-subject and allocated only when the subject has a subscriber, keeping idle subjects zero-cost. The per-property listener path already does a `subject.Data` lookup; the queue and observable channels would add one.

## Out of scope (not fixable at this layer)

- **Cross-property ordering for path subscriptions:** a path value depends on multiple properties and subjects, so a single coherent ordered stream needs a global order, which is deliberately avoided.
- **Dormancy loss:** writes made while a subject is outside every notifying context never enter interception, so no dispatch guarantee can deliver them.

## Relation to path subscriptions (#381)

The path subscription design does not depend on this and is not changed by it. Per-subject sequence numbers do not simplify the path design: its complexity is the moving cross-subject path, not single-property ordering, and the fresh validating walk already provides convergence, structural resolution, and on-path validation in one pass, which a sequence number cannot replace. A future optional optimization could let the path tracker converge via sequence comparison instead of re-reading the leaf on the pure-leaf case, but that is an opt-in mode, not a simplification.

## Suggested order of work

1. Add the per-subject commit sequence (Proposal 1) with a benchmark confirming the idle and single-writer paths are unaffected.
2. Prototype queue-channel FIFO (Proposal 2, queue only) behind a benchmark.
3. Evaluate synchronous-channel FIFO only if a concrete consumer needs it.

RicoSuter:
## Summary

Stamps every committed property write with a monotonic per-subject `Revision` and exposes it on `SubjectPropertyChange`, so a consumer can tell which of two changes to the same property committed later.

That ordering is not otherwise available. A change is dispatched after its commit and outside the subject lock, so under concurrent writers a change that committed later can reach a consumer first, and nothing in the payload says so.

Phase 1 of a two-phase design. The connector work that consumes this is stacked in #420; ordered exactly-once delivery follows after that. Nothing in this repository consumes the revision yet, which is deliberate: the label and its machinery are reviewable on their own, and every defect found while developing the two together was in the consumer rather than here.

Fixes #385 in part.

## Public API

**`SubjectPropertyChange.Revision` is public.** Shipping a consumer-facing revision label is the point of this PR. `docs/tracking.md` documents how to converge on the current value with it.

**Adding the optional `long revision = 0` parameter to `SubjectPropertyChange.Create` is binary breaking.** A consumer compiled against the six-parameter method gets a `MissingMethodException`, and reflection callers break outright (this branch had to fix its own reflection-based test, because `MethodInfo.Invoke` does not fill optional parameters). The library ships a source generator, so consumers recompile against every version and do not bind to a stale signature in practice. A forwarding six-parameter overload would remove the break, at the cost of a permanently doubled surface; the single method with a default is the better shape to live with.

**`InterceptorExecutor.GetOrCreate(ref IInterceptorExecutor?, IInterceptorSubject)` is added.** Needed by both the generator template and `DynamicSubject`, and the generator emits the call into the consumer's assembly, so it cannot be internal.

Those three lines are the entire public API delta.

## Behavior change

A derived recalculation now publishes the stabilized value rather than a re-read of the getter. The old behavior could pair an old value with a new value that never coexisted atomically, and it ran user code on the publish path where a throw suppressed the notification entirely.

Two consequences, neither reachable from anything in this repository: the derived getter is invoked once per recalculation instead of twice, and an interceptor that rewrites `NewValue` on a derived recalculation now changes what is published, where the getter re-read used to mask it.

## How the revision is produced

Both write terminals do `context.Revision = ++context.Executor.Revision;` with the subject's `SyncRoot` held, so a plain increment is exclusive and needs no `Interlocked`. The executor is threaded through `PropertyWriteContext`, the same technique commit 765ad647 used for `Terminal`, which avoids an interface dispatch and a type test on every committed write.

The increment is only safe because the enclosing lock is the subject's `SyncRoot` **and** the executor belongs to that subject. The lock half is lexically enclosing and compiler-guaranteed; the pairing half is asserted with `Debug.Assert` in both terminals, since the internal constructors are visible to the test and benchmark assemblies. In Release that pairing is convention rather than mechanism, which is the one place this design relies on a construction-site invariant.

**A revision is consumed exactly when the terminal runs.** A vetoed write and a write stopped by the equality check never reach it and consume nothing; a derived property's recalculation does reach it and takes a revision of its own.

**Revisions are per subject.** Two changes to one subject are ordered by comparing them; revisions of different subjects are not comparable. A change constructed outside a terminal write carries 0, which orders against nothing.

## A bug found on the way

**The generated context initializer was racy.** `_context ??= new InterceptorExecutor(this)` is not atomic: two threads racing the first `Context` access each built an executor and one was silently discarded along with its state. Under revision semantics that means a discarded counter, so one subject could hand out duplicate revisions and defeat the ordering entirely. It is also unsafe publication independent of revisions, since `??=` publishes through a plain store while `Interlocked.CompareExchange` is a full fence.

Reproduced reliably before the fix, and covered for both the generated accessor and `DynamicSubject`.

## Performance

CPU pinned at 3.6 GHz, turbo off, `--memoryRandomization`, `LaunchCount=10`, on an otherwise idle machine with the .NET build servers shut down first. `RegistryBenchmark` is the suite that matters here: it covers the always-on write path this branch instruments.

**Allocations are the one real cost.** `SubjectPropertyChange` grows 136 to 144 bytes, so allocations grow wherever a change is stored and nowhere else:

| Benchmark | master | branch | delta |
|---|---|---|---|
| `CommitChanges/MultiSource` | 41,088 B | 43,072 B | +4.83% |
| `CommitChanges/SingleSource` | 12,296 B | 12,696 B | +3.25% |
| `ChangeAllTires` | 14,464 B | 14,496 B | +0.22% |
| `AddLotsOfPreviousCars` | 19,904,619 B | 19,944,724 B | +0.20% |
| `CommitChanges/Local`, `Write`, `WriteNoOp`, `Read` | | | unchanged |

The two large cases pin the mechanism exactly: `ChangeAllTires` grows 32 B for the 4 tires it changes, `AddLotsOfPreviousCars` 40,105 B for about 5,000 changes, both 8 bytes each. `CommitChanges/Local` is unchanged despite being a transaction, because it buffers no changes. The `CommitChanges` rows come from an earlier full-suite run; the rest are from the final `RegistryBenchmark` pass.

`PropertyWriteContext<int>` grows 72 to 88 bytes and `PropertyWriteContext<object>` 80 to 96, but it is a stack struct always passed by `ref`, so it allocates nothing and changes no calling convention. Measured with `Unsafe.SizeOf` on both branches. The `SubjectPropertyChange` size is pinned by a test at its exact measured value, so the next field cannot slip through unnoticed.

A derived recalculation with a queue consumer allocates **less** than before, 112 B down to 80 B, because publishing no longer re-invokes the getter.

**No timing regression.** Every write benchmark came out nominally faster than master:

| Method | master | branch | delta |
|---|---|---|---|
| `Write` | 1,199.96 ns | 1,176.06 ns | -2.0% |
| `WriteNoOp` | 423.67 ns | 416.07 ns | -1.8% |
| `WriteWithTimestampScope` | 1,077.84 ns | 1,064.18 ns | -1.3% |
| `AddLotsOfPreviousCars` | 57.01 ms | 56.24 ms | -1.4% |

The mechanism predicts a small cost on these, not a gain (the branch adds a store per commit and 16 bytes of context zeroing per write), so read them as inside the noise floor rather than as an improvement. Two benchmarks the change cannot touch landed within 0.2 ns across branches, which is the calibration for how tight this run was: `IncrementDerivedAverage` 5,903.07 against 5,903.22 ns, and `DerivedAverage` 342.68 against 342.69 ns.

`Read` measured +3.0% here (468.41 to 482.44 ns) with overlapping intervals. It is the control this change cannot affect, and it has now been measured three times on this branch at +4.9%, +6.7% and +3.0%, twice with non-overlapping intervals. Same untouched code, a different answer each run, which is what a code layout artifact looks like: placement is fixed by the build and identical across all launches of a branch, so raising `LaunchCount` cannot average it away. There is also no mechanism for it. `GetPropertyValue` is untouched, `Revision` is declared on the derived `InterceptorExecutor` so the inherited field offsets reads use are unchanged, and the class is now `sealed`, which helps dispatch rather than hurting it.

`Debug.Assert` was confirmed to compile out by scanning the shipped Release assembly for the assert message: absent, while a control literal from the same assembly is present.

## Known limits

**The executor-owns-subject pairing is enforced only in Debug.** See above. Every construction site in the repository was verified to pair correctly.

**A revision orders changes within one subject only.** Consumers that need a global order across subjects do not get one from this, and `docs/tracking.md` says so.

**Revision 0 orders against nothing.** Any change built through the public `SubjectPropertyChange.Create` rather than a terminal write carries it, including transaction capture and rollback inverses, which are descriptions rather than commits.

## Direction change worth flagging

`IInterceptorExecutor.cs` carries a `TODO: Get rid of the executor completely.` Threading a concrete `InterceptorExecutor` through `PropertyWriteContext` makes that removal harder than it was.

A deliberate trade, raised so the reviewer weighs it rather than discovers it: the executor is what makes the increment a plain field increment under a lock the caller already holds, and the alternative was an interface dispatch and a type test on every committed write.

## Tests

2,598 unit tests pass, 0 failures, build clean with warnings as errors.

Moving the revision increment outside the subject lock destroys the ordering scheme, and every other test in the repository still passed with it moved. `WhenOnePropertyIsWrittenConcurrently_ThenEveryCommitTakesADistinctRevision` covers it with eight writers over six rounds, asserting one distinct nonzero revision per committed write and that the highest revision carries the value the property ended up with. It passes in 209 ms and fails in 67 ms with the increment moved out.

Every guard this branch adds was verified by removing it and confirming a specific test fails: the increment's lock, the compare-and-swap executor publication (for both the generated accessor and `DynamicSubject`), and the derived publish short-circuit.

## Review

Reviewed across six dimensions independently (correctness, concurrency, performance, simplification, API and docs and tests, connector integration), on top of the passes taken during development. No correctness or thread-safety defect was found in the revision mechanism: the increment is genuinely exclusive across every write path, executor publication is race-free, revisions are never compared across subjects, and the derived publish change never skips a notification master would send or publishes a value that was never committed.

Every defect those reviews did find was in the connector layer consuming the revision, which is why that work is now a separate pull request.


src/Namotion.Interceptor.Generator/SubjectCodeGenerator.cs
```diff
@@ -139,7 +139,23 @@ private static void EmitInterceptorSubjectImplementation(StringBuilder builder)
...
+                    {
+                        get
+                        {
+                            var context = _context;
+                            if (context is not null)
+                            {
+                                return context;
+                            }
+
+                            // Compare-and-swap rather than ??=: two threads racing the first access
```
RicoSuter: remove comment block, should be clear from

src/Namotion.Interceptor.Dynamic/DynamicSubject.cs
```diff
@@ -27,7 +27,24 @@ protected DynamicSubject(IEnumerable<SubjectPropertyMetadata> properties)
...
+            var context = _context;
+            if (context is not null)
+            {
+                return context;
+            }
+
+            // Compare-and-swap rather than ??=: two threads racing the first access
+            // would otherwise each publish an executor and one would be discarded
+            // along with its state, including the per-subject revision counter.
+            var created = new InterceptorExecutor(this);
```
RicoSuter: can we maybe extract this logic into `return InterceptorExecutor.Create(ref _context)`so it's easy to use here and in the generator? (ensure no perf regression)

RicoSuter: Done in a2968cba as `InterceptorExecutor.GetOrCreate(ref _context, this)`, used by both this file and the generator template. I named it `GetOrCreate` rather than `Create` because it returns the existing instance on every call after the first, and `Create` reads as if it always makes a new one. Happy to rename if you prefer yours.

It has to be public: the generator emits the call into the consumer's assembly, so `internal` is not an option. Additive only, core public API snapshot updated.

No perf regression. The accessor is now a load and a branch, with the allocation behind a separate `NoInlining` method, so it is smaller and more inlinable than the inline version it replaces. Measured on the construction-heavy benchmark: allocations identical (19,944,623 B vs 19,944,612 B), and the timing delta (+1.09%) is smaller than the control benchmark's own drift (-4.5%) in the same run, so nothing is resolvable there.

Worth noting for context: this accessor is not on the property read/write path at all. The generated `GetPropertyValue`/`SetPropertyValue` read the `_context` field directly and skip interception when it is null, so this is the single creation site rather than a hot path.

Verified the extraction did not weaken the guarantee: reverting the shared helper to `??=` fails both `ExecutorPublicationTests` and `DynamicSubjectExecutorPublicationTests` in under 20 ms each.

src/Namotion.Interceptor.Generator/SubjectCodeGenerator.cs
RicoSuter: Removed in a2968cba. The whole block is gone: the generated accessor is now a single expression body, `IInterceptorSubjectContext IInterceptorSubject.Context => InterceptorExecutor.GetOrCreate(ref _context, this);`, which took 17 lines of generated code per subject down to one.

The rationale for the compare-and-swap moved onto `InterceptorExecutor.GetOrCreate` itself, which is the one place it now needs to be stated rather than duplicated into every generated file.

All 16 generator snapshots re-accepted.

docs/tracking.md
```diff
@@ -145,6 +145,25 @@ Dispatch starts on the writing thread, outside the subject lock, and shares one
...
 - **Transactions replay on commit**: with `WithTransactions()`, writes captured inside a transaction do not notify during capture. They replay through the interceptor on commit and notifications fire then. If the transaction is rolled back (disposed without commit), the changes are discarded, no notifications fire, and the property keeps its pre-transaction value. If a best-effort commit partially applies and then reverts, listeners observe the apply-and-revert pair, so a consumer such as a watchdog or dirty flag must not treat the revert as a user change.
 
+### Delivery Guarantees
+
+Every committed write carries a `SubjectPropertyChange.Revision`: a counter that is monotonic **per subject** over committed writes, so two changes to the *same* subject are ordered by comparing it, the higher revision committed later. Revisions of *different* subjects are **not** comparable, and a change constructed outside a terminal write carries `0`, which orders against nothing.
```
RicoSuter, line 151: > change constructed outside a terminal write carries

how can this happen? is this a valid use case even?

RicoSuter: Good question, and the wording it is quoting was misleading. Corrected in d949925a.

**Where revision-0 changes actually come from.** Two internal sites, both building something that is not a commit:

- `SubjectTransaction.cs:132` captures writes made inside a transaction, before commit. Nothing has committed yet, so there is nothing to order against.
- `SubjectPropertyChangeOperations.cs:153` `ToRollbackChange` builds the inverse of a change (old and new swapped) to undo an applied one. An inverse is not a commit either.

Plus any external caller of the public `SubjectPropertyChange.Create`, since the new `revision` parameter defaults to 0.

So yes, it is a valid use case, just not an *ordering* one. Revision 0 means "this is not a commit", which is why it orders against nothing rather than sorting as the oldest.

**Can such a change reach the flush deduplication? No.** A change queue is fed by exactly two call sites, `PropertyChangeInterceptor.cs:196` and `:244`. Both are gated on `context.IsWritten` and both pass `context.Revision`, and `IsWritten = true` is assigned on the same two lines that assign the revision (`WriteInterceptorFactory.cs:30` and `:52`, inside the subject lock). Transaction changes replay through that same write path on commit and get stamped there. So every change a `ChangeQueueProcessor` sees carries a revision of 1 or more, and the fallback is defensive code rather than a live path.

Two independent reviewers reached that conclusion separately, one by reading the write path and one empirically (8 concurrent writers, 35,487 delivered changes, zero revision-0 and zero duplicate revisions).

`docs/tracking.md` now says this explicitly instead of presenting the fallback as something you can hit. I also added a caveat that the same paragraph was missing: revisions decide *which* change's old value survives, not that it is the value held at the preceding revision, because the old value is captured by the generated setter outside the subject lock.

.github/workflows/build.yml
```diff
@@ -40,6 +40,25 @@ jobs:
         id: filter
         with:
           filters: |
+            # The chain every connector is built on: Interceptor -> Tracking -> Registry ->
```
RicoSuter: Remove comments should be clear why

docs/tracking.md
```diff
@@ -145,6 +145,29 @@ Dispatch starts on the writing thread, outside the subject lock, and shares one
...
+
+(a) A throwing lifecycle handler or a throwing earlier observer suppresses delivery for the rest of that write's consumers, so delivery is exactly-once only while those no-throw contracts hold.
+
+(b) Deduplication is scoped to one flush batch, so an inversion straddling a flush tick can still emit the older value last. Compare `Revision` in the write handler if that matters.
+
+**Flush deduplication semantics**: a `ChangeQueueProcessor` with a buffer time collapses each flush batch to one change per property. Per property, the surviving old value comes from the change with the *lowest* revision in that batch and the new value from the change with the *highest*, so the survivor spans the batch even when the enqueue order was inverted. The survivor's `Revision`, `Origin`, `ChangedTimestamp` and `ReceivedTimestamp` come from that same highest-revision change, so under an inverted batch they follow commit order rather than the last arrival: a consumer that keys off `Origin.Source`, for example to suppress echoes, sees the origin of the newest commit. Emit order is the arrival order of each property's last occurrence.
```
RicoSuter: Right place to document? Maybe should be in connectors.md (where cqp docs live and a link here?)

docs/connectors.md
```diff
@@ -113,6 +113,16 @@ Provenance-aware validators receive the origin via `PropertyValidationContext` a
...
+
+Per property, the surviving old value comes from the change with the *lowest* `SubjectPropertyChange.Revision` in that batch and the new value from the one with the *highest*, so the survivor spans the batch even when changes were enqueued in the opposite order they committed. Enqueuing happens after the commit and outside the subject lock, so under concurrent writers that inversion is real rather than theoretical. The survivor's `Revision`, `Origin`, `ChangedTimestamp` and `ReceivedTimestamp` all come from that same highest-revision change, so a handler that keys off `Origin.Source`, for example to suppress echoes, sees the origin of the newest commit. Emit order is the arrival order of each property's last occurrence.
+
+Revisions are monotonic per subject and are not comparable across subjects; see [Delivery Guarantees](tracking.md#delivery-guarantees) for the full contract, including what the old value does and does not promise.
+
+Deduplication covers one flush batch only. An inversion that straddles a flush tick can still deliver the older value last, so compare `Revision` in `WriteChangesAsync` if your source needs strict convergence.
```
RicoSuter: can we fix this somehow so that we are correct and consumers dont need to compare themselves? eg overlapping buffers or something? simplest solution possible?

.github/workflows/build.yml
RicoSuter: Removed in be95eb60, down to one line.

I kept that single line because it is the one thing the paths cannot say themselves: the last four entries (`Testing`, `Hosting`, `Validation`, `AspNetCore`) are not connector dependencies at all. No connector references them. They are referenced only by the connector *integration test projects*, which is why they gate connector suites.

Without that line a maintainer has every reason to read them as a copy-paste mistake and delete them, and the failure mode is silent: the suites simply stop running, and a skipped job looks identical to "no relevant changes". That is exactly the bug this commit exists to fix.

docs/tracking.md
RicoSuter: Agreed, moved in be95eb60.

`docs/connectors.md` gains a **Change Batching and Deduplication** section right after Write Consistency Guarantees, which is where `ChangeQueueProcessor` and its `bufferTime` already live. Links both ways.

Two things stayed in `tracking.md` deliberately:

- The channel comparison table, including the `ChangeQueueProcessor` row. Its value is having all four channels side by side, so splitting it would make it worse.
- The old-value caveat, because it applies to every channel and not just the flush path. I reworded it to say so explicitly.

That section went from 33 lines to 20.

docs/connectors.md
RicoSuter: Yes, fixable, and simply. But not in the way the buffering suggests, and I would not fold it into this PR. Detail below.

**Why it happens.** Thread A commits revision 8 and is preempted before enqueuing. Thread B commits revision 10 and enqueues immediately. The flush fires and emits 10. Thread A resumes and enqueues 8, which the next flush emits. The consumer sees 10 then 8 and writes the older value to the source.

**Why overlapping buffers do not fix it.** Holding changes back so late arrivals catch up only narrows the window, it does not close it: there is no bound on how long a thread can be preempted between committing and enqueuing. It would buy latency on every flush in exchange for a smaller race rather than no race. Strictly worse than the alternative.

**The simplest thing that actually works** is a per-property last-emitted-revision filter on the deduplicator. After the batch collapses, drop any survivor whose revision is not greater than the last one already emitted for that property, then record the new one:

```csharp
if (survivor.Revision != 0 &&
    _lastEmitted.TryGetValue(survivor.Property, out var seen) &&
    survivor.Revision <= seen)
{
    continue; // a newer commit for this property already went out
}
```

Roughly a dozen lines. It makes the source path genuinely convergent, so `WriteChangesAsync` never has to compare anything.

**I had recorded "needs an eviction story" as the blocker, and having now checked, that objection is weaker than I wrote.** `SubjectSourceBase` builds the processor with `propertyReference.TryGetSource(out var source) && source == this` (`SubjectSourceBase.cs:90`), so the dictionary is bounded by the properties that source owns, the same order as its own mapping table. Not unbounded. Graph churn can leave residue for detached subjects, and a cap plus `Clear()` handles that safely, since losing the filter degrades to exactly today behaviour rather than to anything incorrect.

**One non-obvious trap** if we do it: dropping survivors shrinks `_count`, and `Reset` clears only `[0, _count)`, so the dropped tail slots would keep their subjects and boxed values alive in the pooled buffer. They have to be cleared explicitly. That is the kind of detail that makes this a change wanting its own tests rather than a footnote on this one.

**Worth correcting something I wrote earlier:** I had assumed Phase 2 supersedes this. Re-reading the plan, it does not. Phase 2 has `bufferTime = 0` ride the Ordered channel while `bufferTime > 0` stays Immediate, so the batching path keeps flush deduplication and keeps this gap. So this is a real remaining hole on the buffered path, not something the next PR absorbs.

**Recommendation:** small follow-up PR. This one has been through seven review passes and is measured and green; adding a new stateful filter now means new state, new eviction behaviour and new tests on top of that. Happy to do it immediately after, or now if you would rather have it in one go.

src/Namotion.Interceptor.Connectors/ChangeDeduplicator.cs
```diff
@@ -0,0 +1,226 @@
...
+/// captured: the generated setter reads it outside the subject lock, so picking by revision decides
+/// WHICH change's old value survives, not that it is the value committed at the preceding revision.
+/// Both ends are picked by <see cref="SubjectPropertyChange.Revision"/> (commit
+/// order) rather than by arrival position, because a change is enqueued after its commit and outside the
+/// subject lock, so concurrent writers to one property can enqueue in the opposite order they committed.
+/// Owns the pooled scratch buffers so that a flush allocates nothing per batch.
+/// Not thread-safe: the caller must serialize all calls, which <see cref="ChangeQueueProcessor"/> does
+/// by holding its flush gate for the whole deduplicate, write and reset cycle.
+/// </summary>
+internal sealed class ChangeDeduplicator : IDisposable
```
RicoSuter: maybe rename to ChangeQueueDeduplicator or ChangeBufferDeduplicator? Deduplicator seems to be a strange word... ideas?

RicoSuter: Renamed in 636c1953, and you were right that the word was off. It does not remove duplicates, it merges them: the surviving old value comes from the oldest commit in the batch and everything else from the newest. The class body calls `SubjectPropertyChange.MergeWithNewer` four times, so "merger" is the vocabulary the code was already using.

- `ChangeDeduplicator` is now `ChangeMerger`, and `Deduplicate` is `Merge`.
- `EmittedRevisionTracker` is now `DeliveredRevisionFilter`, matching the "delivered" wording the docs use.

I had suggested `ChangeCoalescer`, and you were right to push back on that too: in a .NET codebase "coalesce" collides with the null-coalescing operator and SQL `COALESCE`, both of which mean *first non-null*. Wrong connotation.

I left the noun unqualified rather than taking `ChangeQueueDeduplicator` or `ChangeBufferDeduplicator`, since those keep the inaccurate verb, and `ChangeBufferMerger` reads slightly oddly next to the pooled `_buffer` the class owns internally. Also swept the stale "dedup" vocabulary out of the surrounding comments, the log message and a local variable.

On merging the two types: I would keep them separate. The immediate path (`bufferTime = 0`) uses the filter but has no batch to merge, so a merged type would have that path depending on batch-collapsing machinery to collapse nothing. Their state also has opposite lifetimes: the merger's is per batch and cleared by `Reset`, the filter's must survive it. One class holding both means two clearing rules in one place, and getting that backwards silently breaks convergence.

docs/connectors.md
```diff
@@ -113,6 +113,20 @@ Provenance-aware validators receive the origin via `PropertyValidationContext` a
...
+
+Revisions are monotonic per subject and are not comparable across subjects; see [Delivery Guarantees](tracking.md#delivery-guarantees) for the full contract, including what the old value does and does not promise.
+
+Delivery converges across flushes as well as within one. A change is enqueued after its commit and outside the subject lock, so a writer preempted between the two can land an older commit in a later batch than a newer one. The processor remembers the newest commit already delivered per property and drops anything that commit has superseded, so `WriteChangesAsync` does not have to compare revisions itself. Nothing bounds how long that preemption lasts, which is why buffering longer would not have closed it. Values the source itself pushed in count as delivered, since the source already holds them.
+
```
RicoSuter: “ preemption lasts, which is why buffering longer would not have closed it” remove as its a discussion and does not belong here? Not sure

docs/connectors.md
```diff
@@ -113,6 +113,20 @@ Provenance-aware validators receive the origin via `PropertyValidationContext` a
...
+That memory ages out, so the guarantee covers properties written within a recent window rather than forever. A property that has been quiet for thousands of writes cannot have a straggler in flight anyway, which is the case the window is sized for.
+
+[Source transactions](tracking-transactions.md) write to the source themselves and then apply locally, and that local apply arrives here as a confirmation. Normally it is not sent on, because the source already has it. The exception is when this processor wrote to the same property in between: that write can reach the source after the transaction's and leave it holding an older commit than the subject, so the confirmation is sent out to restore it. A property written only through transactions therefore costs nothing extra, while one written both ways pays a redundant write per transaction to stay converged.
```
RicoSuter: Maybe we should mention a failure in this converge write to be looked at in the transaction epic issue?

docs/tracking.md
```diff
@@ -145,6 +145,25 @@ Dispatch starts on the writing thread, outside the subject lock, and shares one
...
+| `ChangeQueueProcessor`, buffer > 0 | no, latest-state-wins | arrival of survivors; per-property newest within a flush (b) | processor thread |
+
+(a) A throwing lifecycle handler or a throwing earlier observer suppresses delivery for the rest of that write's consumers, so delivery is exactly-once only while those no-throw contracts hold.
+
+(b) Per property, only the newest commit is delivered: a batch collapses to its newest, and a commit that a delivered one already superseded is dropped rather than emitted late. See [Change Batching and Deduplication](connectors.md#change-batching-and-deduplication) for what a survivor carries.
+
+Note what the old value is and is not, on every channel. Revisions decide *which* change's old value survives a collapse, not that it is the value the property held at the preceding revision: the old value is captured by the generated setter at the call site, outside the subject lock, so under concurrent writers it can be a value that was already superseded. The new value is exact, the old value is a best-effort diff baseline. Compare `Revision` or re-read the property if you need more than that.
```
RicoSuter, line 166: > the old value is captured by the generated setter at the call site, outside the subject lock

Would it make sense to read under lock so that old is really correct? Bad idea for perf or deadlocks?

docs/connectors.md
RicoSuter: Agreed, removed in 8aa86661.

It was arguing against a rejected alternative rather than stating the guarantee, and the reasoning already lives on `DeliveredRevisionFilter` where a maintainer changing the mechanism will actually see it. One canonical location instead of two.

docs/connectors.md
RicoSuter: Added in 8aa86661, and checking it turned up something worth stating explicitly.

**What actually happens on failure.** The repair write is an ordinary write on the ordinary path, so it inherits whatever the source is configured with. `SubjectSourceBase` wires the processor to `WriteChangesViaRetryQueueAsync`:

- With a `WriteRetryQueue`, a failure calls `WriteRetryQueue.Enqueue(...)` and the change is retried. Convergence is restored.
- Without one (`WriteRetryQueue is null`), it is logged and dropped. The source stays on the older value until the property is written again.

The second case is not new to this PR. `ChangeQueueProcessor.TryFlushAsync` has always caught, logged `"Failed to write changes."` and continued, so a batch that fails without a retry queue was already lost on master. The transaction repair write just joins that set.

**The part that is new, and is the reason this was worth asking about.** `TryAdmit` records the revision as delivered inside `Merge`, which runs *before* the write is attempted. So after a failed write the recorded baseline is ahead of what the source actually holds.

That degrades safely rather than silently: a newer commit has a higher revision, so it is still admitted and repairs the source. Only an older straggler is suppressed, and sending that would have been the wrong value anyway. So the failure mode is "no repair from an older write" rather than "wrong value written", which is the same outcome as master. But it is a real asymmetry between what the filter believes and what the source holds, and the docs now say so instead of leaving a reader to assume recording implies delivery.

**For the epic.** The durable question underneath is the one you raised earlier: on a failed write-back, should the subject revert to the source value rather than leave local ahead of remote? That is a transaction-semantics decision, not a connector one, so it belongs in #347 rather than here. I will add it there.

Making the record conditional on write success is not the fix, incidentally. The write is asynchronous and batched, the filter is single-threaded scratch state owned by the flush, and deferring the record would reopen the cross-flush window for every change in flight during a slow write.

docs/tracking.md
RicoSuter: Traced it. Short version: the full version is a bad idea for both reasons you named, but a narrow version is not, and it is cheap enough to be worth a follow-up issue rather than accepting the caveat outright.

**Where the old value comes from today.** The generated setter reads the backing field at the call site and passes it down: `SetPropertyValue(nameof(X), newValue, _X, setter)` (`SubjectCodeGenerator.cs:313`). That read is outside any lock. `PropertyValueEqualityCheckHandler` is `[RunsFirst]` and compares `context.CurrentValue` against `context.NewValue` to skip unchanged writes, so the racy read is also what gates the entire chain. Only the terminal takes `lock (subject.SyncRoot)`, and it holds it for the field write, the revision stamp and the timestamp.

**Why reading under the lock wholesale is the bad version.** The equality check would have to move inside the lock too, because comparing outside and writing inside is precisely the race. That means:

- Every no-op write pays a lock acquire plus a full chain traversal. Today `[RunsFirst]` short-circuits before any of that, and `WriteNoOp` is a benchmark because no-op writes are common.
- Every interceptor reading `context.CurrentValue` (validation, tracking, derived recalculation) is observing a value only stabilized later, so they would have to move under the lock as well.
- That is the deadlock half of your question, and it is real rather than theoretical. `DerivedPropertyChangeHandler` cascades into writes on *other* subjects, so running the chain under `SyncRoot` creates cross-subject lock ordering. Today the lock covers a field store and two integer writes and nothing that can re-enter.

**The narrow version that does work.** Keep the equality check where it is (racy, but it only decides whether to skip a no-op, which is benign) and have the terminal capture the true pre-write value under the lock purely for the published change. The terminal already has the subject and the lock; it would need a read delegate alongside `innerWriteValue`, and the generator already emits a static getter lambda for `GetPropertyValue`.

This works because both publish sites read `context.CurrentValue` *after* the terminal returns (`PropertyChangeInterceptor.cs:190` and `:235`, and `context.Revision` is only assigned in the terminal), so a value the terminal writes back into the context does reach the published change.

**Cost, stated honestly.** One more delegate field on `PropertyWriteContext`, which this PR already grew from 72 to 88 bytes and which the description flags as an always-on write-path cost. Plus one delegate invocation per committed write. That is not free, and it is why I would not fold it into this PR.

**Why I would not rush it either.** Concurrent writes to the same property have no single correct "previous value" at the API level. The final value is well defined (last commit under the lock wins, and the revision now says which), but the old value is a question that only has an answer if the reads and writes are serialized against each other, which is the expensive version above. So the narrow fix buys accuracy for change consumers doing delta computation, not correctness for the model.

Recommendation: follow-up issue, sized as "capture the old value in the terminal", with the delegate cost measured against `WriteNoOp` and `Write` before accepting. Happy to open it. The `docs/tracking.md` caveat stays either way, since it describes what the field promises rather than a defect.

RicoSuter: Opened as #413, scoped to the narrow version (capture in the terminal, leave the equality check where it is) with the benchmark gate and the "is this worth paying for" question stated up front rather than assumed.

`docs/tracking.md` keeps its caveat unchanged, since it describes what the field promises rather than a defect.

docs/connectors.md
RicoSuter: Refining my previous reply, because checking the retry queue changed the answer and my wording oversold it. Corrected in 543a9dc4.

**Are we going out of sync with remote? Yes, in the failure path.** Concretely:

1. The transaction writes `T` directly to the source. Source = `T`.
2. Our processor's write of an older local commit `W` lands afterwards. Source = `W`.
3. The transaction applies `T` locally. Subject = `T`, published as `Confirmed`.
4. The write-back sends the confirmation. On success, source = `T` and it is converged.
5. On failure, source = `W` and subject = `T`, out of sync, with nothing actively reconciling.

**What I got wrong the first time.** I described the retry queue as if it always rescues this. `WriteRetryQueue.Enqueue` is a ring buffer that drops its oldest entries at capacity (default 1000), so when writes fail across many properties at once a pending repair can be evicted before it is ever retried. The queue makes the failure much less likely, not impossible.

**How it heals.** Only through the next successful write to that property, or a source reconnect reloading initial state. If the property is never written again, local and remote stay diverged indefinitely.

**Why "closed" here means documented rather than fixed.** On master the confirmation was never written back at all, so source = `W` with subject = `T` was the unconditional outcome for this interleaving. This PR repairs the success path and inherits master's behaviour on the failure path, so it is strictly better while not being complete.

I also dropped the claim that a suppressed straggler "would have been the wrong value anyway". It is a value the subject has superseded, not a value that is wrong to hold, and in the narrow case where the newer write failed, sending it would have left the source closer to current than suppressing it does. That needs a write failure plus an inversion plus an evicted or disabled retry queue plus no subsequent write, so it is not worth engineering against, but it should not have been waved away.

The durable question, whether a failed write-back should revert the subject to the source value instead of leaving local ahead of remote, is transaction semantics rather than connector behaviour. Happy to add it to #347 on your go-ahead.

RicoSuter: Closing the loop: the epic note is now on #347 (https://github.com/RicoSuter/Namotion.Interceptor/issues/347#issuecomment-5176899367).

Reading the epic first changed what was worth writing there. The repair polarity is **not** the same decision as the failed-revert case row 5 already designs: #349 classifies a write failure and requests a *source-wins* resync, which is right when the transaction did not commit. Here it did commit, so `T` is the committed value and `W` is a superseded one that only reached the source because of the interleaving. A source-wins resync would resurrect `W` and discard a committed transaction value, which is worse than the divergence it repairs.

So it needs the same `WriteFailureKind` machinery with the opposite polarity, which makes "which side wins" a per-failure-kind decision rather than a uniform one. That is the part I suggested pinning in row 4 contract wording, with row 5 enumerating this as a classifiable failure.

I also noted that `SourceOwnershipManager.ReapplyRetryQueue` already implements a reconciliation on reconnect that is closer to what this case wants than source-wins is, and listed the four gaps that stop it covering this one (it keys off the best-effort `OldValue` from #413, the retry queue evicts oldest at capacity, a rejected write that does not drop the connection never restarts the listen loop, and a disabled queue skips it entirely).

Nothing further needed on this PR for it; the connector-side behaviour is documented in `docs/connectors.md`.

docs/tracking.md
```diff
@@ -143,8 +143,28 @@ Dispatch starts on the writing thread, outside the subject lock, and shares one
...
 - **Throwing synchronous observers suppress later deliveries**: each interceptor dispatches its queue first, then its Rx observable, then any per-property listeners it resolved. With aggregated contexts, the innermost interceptor resolves the per-property listeners, so they may run before outer contexts' queue and Rx channels. An exception from any synchronous observer propagates out of the write and prevents later deliveries in that order; queue items already enqueued remain available. Keep synchronous observers exception-free. The exception surfaces from the setter after the value was committed and nothing is rolled back; the property keeps the new value. For scheduler-based Rx observers, delivery means the change was accepted by the channel, not that the callback has already run.
+- **A derived recalculation publishes the stabilized value**: the change carries the value the recalculation committed rather than a fresh read of the getter. The getter therefore runs once per recalculation instead of twice, a throwing getter no longer suppresses the notification, and an interceptor that rewrites `NewValue` on that path now changes what is published.
```
RicoSuter: Dont say no longer, we should describe now not history

PR #400, opened 2026-07-31
Replace context locking with copy-on-write state snapshots

### LINKED ISSUE #401: Pure delegation cycle recurses until stack overflow
Two contexts that delegate to each other with zero services form a pure delegation cycle. `GetServices` and the three `ExecuteIntercepted*` methods follow `DelegationTarget` without a visited set on that path, so the recursion never terminates and the process dies with a `StackOverflowException`, which cannot be caught.

This is pre-existing and predates #400. The old `_noServicesSingleFallbackContext` fast path had the identical behavior on the identical path, so #400 preserves it deliberately rather than introducing it.

## Reachability

It is constructible, not merely theoretical. `ContextInheritanceHandler` removes a fallback only on `{ ReferenceCount: 0, IsPropertyReferenceRemoved: true }`, so a context detach that leaves property values set strands the fallback edges. Orderings exist where a mutually referencing subject pair is orphaned from the root and then partially detached, leaving `A.Context` with `[B.Context]` and `B.Context` with `[A.Context]` and no services on either. The next property access on either subject crashes the process.

A cycle with at least one service on either side is fine: the delegation fast path does not apply, and the walk's visited set terminates it.

## Related, milder symptom

The same stranded-edge mechanism lets a still-attached subject retain a fallback to a fully detached context and keep resolving services through it.

## Options

- Add a depth guard or a visited set to the delegation path. Costs something on the hottest path in the library, so it needs measuring.
- Fix the detach path so it strips stale fallback edges, which addresses the root cause and both symptoms.
- Reject a cycle at `AddFallbackContext` when both contexts would be pure delegators.

The second looks most promising since it also fixes the stale-resolution symptom, but it needs a reproduction from the registry detach path first.

### LINKED ISSUE #405: Invalidation walk recurses per using context and overflows the stack on very deep graphs
`InterceptorSubjectContext.InvalidateUsingContexts` and `InvalidateUsingContext` recurse once per using context, so a sufficiently deep context graph overflows the stack. Verified: `AddService` on the root of a 100,000 deep delegation chain dies with a `StackOverflowException`, which cannot be caught and takes the process with it.

Pre-existing, and unrelated to the resolving path. Found while reviewing the fix for #401.

## Why it is worth fixing separately

#401 removed recursion from the four resolving entry points, so `GetServices` and intercepted property and method access now handle a chain of any depth without touching the stack. The invalidation walk in the same class still recurses, so the practical depth ceiling moved rather than disappeared: resolution is unbounded, but any topology mutation that has to invalidate upwards through a deep graph is not.

Depth here is the depth of the subject graph, since every attached child registers into its parent's used-by set.

## Shape of a fix

Convert the upward walk to an explicit stack or queue. It already carries a `[ThreadStatic]` visited set (`_invalidationVisited`) that prevents revisiting and terminates cycles, so the change is mechanical: replace the recursive call with pushing onto a worklist and looping until it drains. The visited set already gives the termination argument.

Worth measuring afterwards, since the walk runs on every topology mutation, though a worklist is likely cheaper than the call frames it replaces. The existing 0/1/many snapshot shapes that avoid an array allocation in the common cases should be preserved.

RicoSuter:
Replaces the mutable-fields-plus-locks design of `InterceptorSubjectContext` with a single immutable state snapshot published atomically, then makes delegation resolution cost the same at any depth. This closes eight concurrency defects and two crashes, and makes reads about 20 percent faster against `master` on a shallow graph and about 4 times faster on a deep one.

The PR started as two targeted locking fixes. Fixing the first two defects exposed a third, whose fix exposed a fourth and a fifth, and the sixth had no contained fix at all. At that point the patch was maintaining five hand-written invariants across four mutators. The rewrite replaces all of them with one.

Closes #401. Closes #405.

## The defects

| # | Defect | Consequence |
|---|---|---|
| 1 | Lock order inversion between the downward service query and the upward invalidation walk | Hard hang, no exception, no timeout |
| 2 | First-attach fast path skipped invalidation propagation | Permanently stale compiled chain |
| 3 | `_noServicesSingleFallbackContext` observable as stale while `_lock` was held | `TryAddService` can admit a duplicate |
| 4 | `_methodInvocationFunction = null` performed outside `_lock` | Invalidation lost to a concurrent chain build |
| 5 | Cyclic fallback graphs, and `factory()` invoked under `_lock` | Deadlock, reachable with ordinary factories |
| 6 | Cache insert landing after the invalidating `Clear` | A compiled chain missing an interceptor, cached permanently |
| 7 (#401) | A pure delegation cycle recursed until the stack overflowed | Uncatchable process crash |
| 8 (#405) | The invalidation walk recursed per using context | Uncatchable process crash on a deep graph |

Defect 6 is the most serious. It is not limited to the service cache: `GetReadInterceptorFunction` and `GetWriteInterceptorFunction` build a chain and then `TryAdd` it, so an invalidation in between leaves a chain that silently skips an interceptor for the lifetime of the context.

Defect 5 was confirmed by measurement rather than argued from the code. With a cyclic fallback graph, two threads and entirely ordinary `() => new Cross()` factories, `master` deadlocks on the first round; this branch completes 20,000 of 20,000. Cycles are not exotic here: `ContextInheritanceHandler` gives a newly attached child its parent's context as a fallback, so any back-reference forms one.

## The design

All topology and everything derived from it lives in one immutable `ContextState`: services, fallback contexts, the delegation target, and the caches. Four rules govern it, each carried as a comment on the code it governs.

**R1. Queries take no context lock.** A query pins one snapshot with a single volatile read and walks other contexts' snapshots the same way. With no context lock on the query path there is no lock cycle, so defects 1 and 5 cannot occur, including in cyclic graphs.

**R2. Mutators publish with one interlocked exchange under `_mutationLock`, no CAS loop.** The only lock-free writer is the invalidation CAS, which never changes topology, and a mutator's state is born cache-free, so overwriting a concurrent invalidation preserves its intent.

**R3. Invalidation is one unconditional CAS attempt.** No early-out when caches look absent, because a reader may be lazily creating one concurrently. No retry on failure, because every competing write also publishes cache-free state.

**R4. `_usedByContexts` is always a superset of the true using set.** `AddFallbackContext` registers upward before publishing; `RemoveFallbackContext` unregisters after. A missing entry is defect 2.

Three defects stop being things the code fixes and become things it cannot express: the delegation target is derived in the constructor so it cannot disagree with the topology it comes from (defect 3); caches belong to the state that produced them, so a late insert lands in an abandoned object (defect 6); publishing a new state is itself the invalidation, so there is no `Clear()` and no null-out whose ordering could be wrong (defect 4).

## Delegation depth

Every subject has its own executor context, and `ContextInheritanceHandler` gives each child its parent's context as its only fallback. An executor registers no services on itself, so the default topology is thousands of pure proxies delegating toward one base context that holds everything. Chain length equals subject-graph depth, and `master` re-walks that chain on **every** intercepted read, write and method call.

Each state now records the context its chain ends on, so the chain is walked once per state instead of once per access, and the walk records what it found on every state it passed. Depth becomes free.

Two rules make it correct, and both were established the hard way, by shipping the wrong one twice and having review catch it:

> An installed state always describes the topology of **its own** context correctly, so `DelegationTarget`, `FallbackContexts` and `Services` may be read from any state a walk reaches.
>
> A recorded chain end is a **derived claim about contexts further down**. It may be trusted only for the context being resolved, never for a context reached during a walk.

The second rule is not obvious, and the reason is R4 itself. Because a using set is deliberately a superset, `RemoveFallbackContext` publishes before it unregisters, so for a moment a context that now delegates still sits in the using set it is leaving. An invalidation arriving over that entry gives it its final state, and marks it visited so it is never invalidated again in that walk, while contexts further down its chain still carry pre-mutation records. A walk trusting one of those records caches a resolution of a context the graph no longer reaches, on a state nothing invalidates again. That is defect 6 rebuilt, and it is why the resolution walk reads topology only.

Cycle detection moved off the query path with it. Running once per state, it can afford to be exact, so the Floyd walk, the separate exact re-walk and the hop budget are gone. The confirmation compares `ContextState` object identity rather than the fallback contexts states point at: a state is never installed twice, so one still in place has been in place since it was pinned, which proves the loop edges existed at the same moment rather than merely one after another. That also closes a residual the previous confirmation had, where a rolling sequence of rewirings could produce a repeat that never was a cycle.

The retry that a failed confirmation triggers backs off with a `SpinWait` before walking again. Reaching it means a mutator is rewiring the chain at that moment, since the walk had to observe a repeat and then find one of that loop's states already replaced, and nothing else bounds the loop. Temporary instrumentation confirmed the path is live rather than theoretical: under a mutator toggling a cycle-closing edge against four resolving threads it fires roughly 200,000 times a second. A walk that never retries, which is every walk on a graph nobody is mutating, does not reach it.

The verdict is recorded on the confirmed loop and on nothing ahead of it. The confirmation re-reads the states of the loop only, so a context on the run leading into it reaches that loop according to an edge read earlier, which a concurrent rewiring can have moved since. Recording the verdict there would make every query on that context raise until a pending invalidation replaced its state, and a caller cannot tell that from a cycle it really is on. This is the one place the asymmetry matters: a stale terminal resolves the wrong services for a moment and then converges, while a stale raise does not read as something transient. `WhenAcyclicPrefixLeadsIntoDelegationCycle_ThenOnlyTheCycleRecordsTheVerdict` pins it and fails on the first iteration without it.

Compiled chains are found by indexing an array with a dense per-property-type index instead of hashing a `Type`. The arrays exist only on contexts a chain ends on, so a graph of thousands of delegating subjects has as many arrays as it has contexts that answer.

## Memory model

Review found two ordering hazards in the new code, neither present on `master`, both fixed before merge. They are why `PublishState` uses `Interlocked.Exchange` rather than `Volatile.Write`, and why the invalidation walk does not test the using set for emptiness outside its lock.

A mutator publishes, releases `_mutationLock`, then reads using sets and other contexts' states to drive the upward walk. A store followed by a load is the one pair release semantics does not order, and both `Monitor.Exit` and a volatile write are release-only, so neither helps. This is not an ARM64 detail, as an earlier version of this description said: store-load is reordered under every mainstream memory model, x86-64 TSO included, where the store sits in the store buffer while the later load is already satisfied. The consequence would be an invalidation skipped against a stale used-by set while the current state keeps accepting cache fills computed from pre-mutation topology.

The second hazard was an unlocked emptiness test. `HashSet.Count` is `_count` minus `_freeCount`, two independently mutated fields, so an unlocked read can compute a count that was never true at any instant, and a using context that never left the set could be skipped for good.

## Verification

**Tests.** 2651 unit tests pass. The 4-case deadlock theory written for the earlier patch passes unchanged, and fails on `master` on the first or second attempt of all four cases.

A correction to an earlier version of this description, which claimed `TryAddService` atomicity at 0 of 3000 violations against 1658 of 3000 on `master`. That second number does not reproduce: the test as written measures 0 of 3000 on `master` too, over four runs and against a build of `master` itself. It cannot distinguish the two implementations, because both serialize the exists-check and the add on one context lock, so it guards a lock-free `TryAddService` that neither branch has. Defect 3 is left in the table on the strength of the code path rather than that measurement, and it needs a test that mutates the fallback concurrently to be guarded at all.

**Tests proven able to fail.** Four defects were deliberately reintroduced to check the tests that guard them, rather than assuming they would catch anything:

| Reintroduced defect | Caught by | Rate |
|---|---|---|
| Trusting a recorded chain end in the service walk | `WhenChainIsRewiredBelowADiamond` | first iteration, 14 ms |
| Throwing on a cyclic chain reached as a fallback | `ContextConcurrencyFuzzTests` | 8 of 8 seeds |
| Skipping the visited check on a jump target | `ContextServiceWalkOrderTests` | 3 of 8 seeds |
| Reusing the caller's pin when a walk restarts | `WhenCallerPinnedCyclicStateWasReplacedWithTerminal` | deterministic |

**Randomized fuzzing.** A model-based fuzzer builds random fallback graphs, hammers them from several threads with topology mutations, service registrations, queries and intercepted access, then asserts quiescent consistency against a single-threaded walk of the final graph. Four oracles: resolved services, reached interceptors, which contexts must raise instead of resolving, and the recorded chain end of every context compared against the final topology directly. Validated at 8,000 topologies and roughly 77 million operations, on the commit this merges.

The corpus itself was the finding that mattered. It previously could not build a delegation chain longer than two hops, and its workers gave nearly every context a service within a few hundred operations, so no final topology delegated at all and the quiescent oracle never saw the delegation path. Contexts that never receive a service now take part, chained head to tail the way one executor per level is, and coverage assertions fail a run that produces no chain of three or more hops, no chain that is a cycle, or no recorded chain end to compare.

**Known limits of the fuzzing.** Three things here rest on argument rather than on a test, and a green fuzz run is evidence about none of them.

The out-of-order invalidation window is the first. With that defect deliberately reintroduced, 8,000 topologies and 77 million operations pass, as does a directed 400-iteration concurrent test. The window is the few instructions between a publish and an unregister. It is guarded by construction, and the equivalent defect on the collecting walk is guarded by a test only because its window is an entire invalidation walk wide.

R2 and R3 are the other two. The publish fence closes a store followed by a load, and the single unretried invalidation CAS is sound because a competing state can only win it after the mutation was published. Both are ordering claims about interleavings no test in this repository can schedule, and neither would fail on the hardware CI runs on even if it were wrong. They are stated where the code relies on them and verified by review only.

The cycle verdict is recorded on the confirmed loop rather than on the run leading into it for the same reason: the test that pins it asserts the code shape, not the race that motivates it.

**Review.** Five independent adversarial passes, all instructed to refute rather than confirm, and told not to trust the comments. They found the publish fence, the two unsound record shortcuts, unbounded thread-static buffer retention, a type-check helper call on the hot path, and an array element helper call. Two comments that were load bearing and false were removed.

**Stress.** A 150,000-context chain rewired continuously under 64 oversubscribed walkers, scripted so the graph is acyclic at every instant: 1.13 billion calls, zero false cycle reports, no hang. Note that the walk itself runs about 81,000 times there rather than once per call, because the recorded chain end absorbs the rest, so that is the number of adversarial walks the zero is measured over.

## Benchmarks

`RegistryBenchmark` against `master`, CPU pinned to 3.6 GHz with turbo disabled, memory randomization on, five process launches per benchmark.

| Benchmark | master | branch | delta |
|---|---:|---:|---:|
| DerivedAverage | 344.9 ns | 238.7 ns | -30.8% |
| WriteNoOp | 428.6 ns | 335.2 ns | -21.8% |
| Read | 471.7 ns | 369.1 ns | -21.8% |
| WriteWithTimestampScope | 1,085.2 ns | 950.7 ns | -12.4% |
| IncrementDerivedAverage | 5,855.9 ns | 5,195.2 ns | -11.3% |
| Write | 1,177.9 ns | 1,088.9 ns | -7.6% |
| AddLotsOfPreviousCars | 57.17 ms | 55.13 ms | -3.6% |
| ChangeAllTires | 14,514 ns | 14,790 ns | +1.9% |

Control benchmarks this change cannot affect moved by -0.9% to -0.2%, so everything outside that is real. Allocations improve by 416 KB on the attach-heavy benchmark and 384 B on `ChangeAllTires`.

That graph is three levels deep. `ContextDelegationDepthBenchmark` measures what depth costs, compared against a branch that is `master` plus this benchmark file unchanged, so both sides run identical benchmark code:

| | Depth 1 | Depth 8 | Depth 64 |
|---|---:|---:|---:|
| Read, master | 195.7 ns | 287.0 ns | 1,816.6 ns |
| Read, branch | 152.5 ns | 158.8 ns | 152.3 ns |
| | -22.1% | -44.7% | -91.6% |
| Write, master | 680.5 ns | 873.1 ns | 2,467.5 ns |
| Write, branch | 652.4 ns | 660.6 ns | 654.4 ns |
| | -4.1% | -24.3% | -73.5% |

The point is not the 11.9x at depth 64, it is that the branch is flat: 152.5, 158.8 and 152.3 ns for the same work at one, eight and sixty-four levels, where `master` grows ninefold. Depth stops being something a subject graph pays for on every property access.

`ChangeAllTires` regresses by about 1.9 percent, and it is the one benchmark dominated by attach and detach. Measured over three A/B runs rather than one, because a single run cannot separate this from noise: `master` spans 14,596 / 14,514 / 14,486 ns across runs of identical code, while the branch sits at 14,790 and 14,829 ns after this fix, above every `master` sample. The same benchmark measured +0.5 percent before it, so the fix is what moved it.

It is the price of the second rule above. A cold walk can no longer stop at the first context that already recorded where its chain ends, so it takes one more hop, and each hop costs an entry in the walk's visited set and path list rather than just a pointer read. Attaching four tires pays that four times.

Only cold walks are affected, meaning the first resolution on a context and the first after something on its chain changes. Steady state is untouched, which is why every read and write benchmark improves. Allocations on this benchmark improve by 384 B.

The obvious way to win it back is to defer the visited-set bookkeeping for the first few hops and detect a repeat by scanning the path list instead, which is cheap while the path is short and is sound because it trusts no recorded state. That is left out of this PR deliberately: it is a new optimization on a hot correctness path that has already absorbed five review rounds, and it belongs in its own change with its own review.

## One behaviour change beyond the defects

Fallback contexts moved from a `HashSet` to an `ImmutableArray`, which changes their order after a removal. A `HashSet` reuses the freed slot on the next add, an array appends:

```
{A, B, C}, remove B, add D
HashSet:        A, D, C
ImmutableArray: A, C, D
```

Fallback order feeds frame order in the service walk, so it decides the relative order of services that no ordering attribute separates, and therefore the execution order of such interceptors. It is reachable in normal use, because `ContextInheritanceHandler` removes a fallback on every detach and adds one on every attach, though it needs a context with at least two fallbacks to be visible at all.

This reads as a fix rather than a regression: the old order was an artifact of `HashSet` free-list reuse, never designed and never documented, while the new one is registration order. Anything that needs a defined order declares it with the ordering attributes, which both versions honour identically. It is called out here because nobody could have depended on the old order deliberately, but something may have depended on it by accident.

## Out of scope

Six follow-up issues are filed separately. #402 through #404 and #406 concern behaviour that predates this PR. #409 and #410 were opened from the review of this branch.

* #402: `InterceptorExecutor`'s fallback overrides run their lifecycle callbacks in the wrong order relative to `base`. Concurrent removes can run the detach callbacks twice, a remove racing an add can silently undo the add, and on a context whose chain is a cycle the remove path throws before it removes while the add path leaves a half-attached subject. The remove half is fixed by the same inversion, which also turns out not to need the workaround the issue originally proposed.
* #403: `TryAddService` at two delegation levels concurrently can admit a duplicate, which later makes `TryGetService` throw.
* #404: `TryAddService` runs its factory and predicate under the mutation lock, so a factory that mutates a different context, raced by another thread, deadlocks. Documented as a contract here. This PR narrows the hazard considerably, since `master` deadlocks in a cyclic graph with ordinary factories and no contract violation at all.
* #406: Service equality can reenter a cold service walk and cache a partial result. The same follow-up also tracks duplicate service retention and makes the intended equality semantics explicit.
* #409: The two memory trade-offs this design makes, neither affecting correctness and neither measured. Compiled chain arrays are sized by a process-wide property type index rather than by what a context uses, and invalidation now allocates one state per delegating context where `master` allocated none for those.
* #410: `ContextInheritanceHandler` strands fallback edges on a partial detach. This is the mechanism that builds the cycles defect 7 is about, and it also lets a still-attached subject keep resolving through a fully detached context, which raises nothing. Split out of #401 so that closing it here does not lose the part this PR does not fix.

The public and protected API of the Interceptor libraries is unchanged, verified by the untouched `PublicApi` snapshots. One app-level signature changes: HomeBlaze's `SubjectPathResolver` constructor no longer takes the context, because its self-registration was redundant.


src/Namotion.Interceptor.Benchmark/DelegationDepthBenchmark.cs
```diff
@@ -0,0 +1,89 @@
...
+/// <summary>
+/// Measures the intercepted read and write fast path as a function of delegation-chain depth.
+/// Every attached child subject inherits the context of its parent as its only fallback context,
+/// so a subject graph of depth N resolves every property access through a chain of N delegating
+/// contexts. The other benchmarks only cover graphs 2 to 3 levels deep, so this is the one that
+/// sees the cost of chain resolution. <see cref="Depth"/> counts the proxy contexts between the
+/// subject and the context holding the services; the subject's own executor adds one more hop.
+/// </summary>
+[MemoryDiagnoser]
+public class DelegationDepthBenchmark
```
RicoSuter: Should we rename to ContextDelegationDepthBenchmark so its clear its about context for someone scanning it later?

RicoSuter:
## How to review this

`InterceptorSubjectContext.cs` is about 1,080 lines, of which roughly 280 are comments. Reading it top to bottom is not a good use of attention, and I would rather say why than let someone try.

The defects that mattered here were not findable by reading. Five adversarial passes found two, and both needed a specific interleaving constructed across three threads before they were visible. A linear read will not surface that class of bug. So the suggestion is to review the reasoning and the evidence, and to leave the interleavings to the fuzzing and the adversarial passes.

### 1. Six invariants, each stated where it is relied on

Everything else follows from these. Checking that each one is argued where it is used, and that nothing nearby quietly violates it, is the highest value pass.

| Invariant | Stated at | Pinned by |
|---|---|---|
| R1, queries take no context lock | class comment, top of file | the deadlock theory in `ContextLockingTests` |
| R2, publish is one interlocked exchange under the mutation lock | `PublishState`, line 768 | argument only, no test can reach it |
| R3, invalidation is one unconditional CAS with no retry | `InvalidateState`, line 783 | the fuzzer |
| R4, a using set is always a superset of the true using set | `AddFallbackContext` line 111, `RemoveFallbackContext` line 147 | `WhenVeryDeepChainHasMultiFallbackNodes` |
| A `ContextState` is installed exactly once | `WithoutCaches`, line 990 | `WhenContextIsInvalidated_ThenItsStateObjectIsReplaced` |
| Topology may be read from any state a walk reaches; a recorded chain end only from the context being resolved | `ResolveDelegationChain` line 308, `TryEnterContext` line 645 | `WhenChainIsRewiredBelowADiamond` |

The last two carry the most weight. Install-once is what makes the cycle confirmation exact and what makes a record on a still-installed state trustworthy. The topology rule is the one that was got wrong twice during this PR, in both walks, with a different wrong justification each time, so it is worth reading `ResolveDelegationChain` and `TryEnterContext` specifically to see that neither reads a record from a context it reached rather than started from.

### 2. The public API and the semantics

This is the part a human reads better than any tool, and where review here actually found something: that a subject registering its own services, scoped to itself and its children, is a load bearing extension point and not dead code. No amount of adversarial review found that, because it is a question about intent rather than about the code. `ContextSubtreeServiceTests` now pins the behaviour.

### 3. Whether the tests can fail

Six defects were deliberately reintroduced to check that the test guarding each one actually fails. The evidence is in the commit messages, with detection rates where they are not deterministic. That is more checkable than the code, and three tests written during this PR looked reasonable and caught nothing until they were checked this way.

Worth knowing: one defect class, an invalidation arriving out of chain order, is guarded by construction and by no test. With it deliberately reintroduced, 8,000 fuzz topologies and 77 million operations pass. That is stated in the description so a future green fuzz run is not misread as evidence about it.

### 4. The numbers

Read the deltas against the control benchmarks that this change cannot affect, which moved by -0.9% to -0.2%. The one regression, `ChangeAllTires` at +1.9%, was reproduced across three A/B runs before being reported, because a single run cannot separate it from noise: `master` alone spans 0.77% across runs of identical code.

### What to skip

The interleavings. `CollectServices`, `InvalidateUsingContexts`, `QueueUsingContexts` and the walk bodies are correct for reasons that are stated above them and were attacked from several directions. Verifying them by inspection is not realistic, and the invariant table is the tractable version of the same question.

src/Namotion.Interceptor/InterceptorSubjectContext.cs
```diff
@@ -9,149 +9,209 @@ namespace Namotion.Interceptor;
...
-                return true;
+            // R4: register into the fallback BEFORE publishing so that its _usedByContexts is
+            // always a superset of the true using set. A missing entry would leave a compiled chain
+            // above permanently stale. An extra entry costs a spurious invalidation, and lets the
+            // invalidation walk reach a context out of chain order, which is why no walk may trust
+            // what a context further down recorded about its chain (see ResolveDelegationChain).
+            var usedByContexts = contextImpl.GetOrCreateUsedByContexts();
+            lock (usedByContexts)
+            {
+                usedByContexts.Add(this);
```
RicoSuter, line 138: [P1] Use reference identity for context topology

`InterceptorSubjectContext` is public and unsealed, but this set uses the default comparer. Two distinct context subclasses can therefore compare equal, causing the second reverse-edge registration to be discarded while its fallback edge is still published.

I reproduced a permanent stale cache with equality-equal contexts `A` and `B`, each carrying a local service and falling back to `D`: after `B` cached one service, adding a service to `D` invalidated only `A`, and `B` continued returning one service instead of two.

This is pre-existing, but it directly violates R4, which requires `_usedByContexts` to be a superset of the true using set. Equality callbacks can also run arbitrary code while this set lock is held, invalidating the leaf-lock argument.

Please use an internal reference comparer based on `ReferenceEquals` and `RuntimeHelpers.GetHashCode` for `_usedByContexts` and all context visited sets, and use reference comparisons for fallback duplicate detection and removal.

RicoSuter: Resolved by #407, now merged into this branch via c83efd4d. `InterceptorSubjectContext` has a `private protected` constructor and `InterceptorExecutor` is sealed, so external code cannot create a context subtype that overrides equality. This enforces the reference-identity invariant without additional comparers.

src/Namotion.Interceptor.Benchmark/DelegationDepthBenchmark.cs
RicoSuter: Renamed the file and class to `ContextDelegationDepthBenchmark` in 65fb466b. The benchmark methods are now `ReadWithDelegation` and `WriteWithDelegation`, paired with the existing `WithoutDelegation` methods.

RicoSuter:
## Performance impact of commits after the benchmark review

The [benchmark review comment](https://github.com/RicoSuter/Namotion.Interceptor/pull/400#issuecomment-5151285530) was posted on August 1 at 11:47 UTC. Eight commits followed it.

| Later change | Performance relevance |
|---|---|
| [Merge of current master](https://github.com/RicoSuter/Namotion.Interceptor/commit/c83efd4d97e12a7003f9abc79a62d600dff1f66b), including [#407](https://github.com/RicoSuter/Namotion.Interceptor/commit/32af6f995b044d5b3e3665ec48db73c40b9064e5) | Potentially relevant. It seals `InterceptorExecutor`, which may alter JIT devirtualization or inlining on intercepted-access paths. Both branches receive it, but the published A/B results no longer represent the current base/head pair. |
| `65fb466b` naming clarification | Renames only. |
| [HomeBlaze lifecycle registration fix](https://github.com/RicoSuter/Namotion.Interceptor/commit/f8955ee4377998ca0680c4aadbd5e6274c8b51ea) | Only relevant to HomeBlaze DI/startup performance, not the reported context benchmarks. |
| `98c95fd3` and `aa52bf4d` | Tests only. |
| `fc0eb21b` | Comments only. |
| [Cycle-verdict fix](https://github.com/RicoSuter/Namotion.Interceptor/commit/ee5e4cbb0f006f9ba52992a4ddc309498123c940) | The steady-state fast path is unchanged, but cold delegation resolution was touched. That could affect `ChangeAllTires`, whose reported regression was only +1.9% and is dominated by attach/detach and cold resolution. |
| [Latest coverage commit](https://github.com/RicoSuter/Namotion.Interceptor/commit/abd9f1b9eb90d412df3371c343a727a9f7a4ff5e) | No Release performance impact. Runtime additions are comments and a `Debug.Assert` in the cycle-only path. |

Recommendation: rerun `RegistryBenchmark` and `ContextDelegationDepthBenchmark` A/B against current `master`. At minimum, rerun `ChangeAllTires`, the read/write fast paths, and the depth benchmarks. This is mainly to refresh the published evidence, not because there is a likely new regression.

Issue #411, opened 2026-08-03
An add is refused during the two-phase fallback removal window and the edge then disappears


RicoSuter:
Split out of #402 so it is not lost when that issue closes. Not closed by #402's fix, and widened by it.

## The window

`InterceptorExecutor.RemoveFallbackContext` removes an edge in two phases, because the detach callbacks have to run while the edge is still resolvable and they cannot run under the mutation lock:

```
phase one: claim the ownership record, leave the edge
           detach callbacks run here (user code, arbitrary duration)
phase two: publish the removal
```

Between the phases the edge exists with no record. A concurrent `AddFallbackContext` sees the edge present and returns `false`, which per `IInterceptorSubjectContext` means the context was already registered. The in-flight removal then drops the edge.

Quiescent state: no edge, and a caller that was told the edge was already there.

```
A: TryTakeFallbackAttachment -> claims the record, edge still present
C: AddFallbackContext        -> edge present, returns false, attaches nothing
A: detach callbacks, publish the removal -> edge gone
```

## Why it is not just pre-existing

The shape existed before #402's fix: `RemoveFallbackContext` always ran the callbacks between deciding and committing. What changed is the width. Previously the window spanned the callbacks between a lock-free `HasFallbackContext` check and the commit; now it spans the callbacks between two locked sections. The duration is the same user code either way, so this is not a regression in kind, but it is now a deliberate part of the design rather than an accident of the ordering, and it should be recorded as such.

## Why the obvious fixes do not work

- **Blocking the add until the removal completes** deadlocks. The removing thread is inside detach callbacks that take `LifecycleInterceptor._attachedSubjects`, and the adding thread may already hold it. This is the same wall that made a refusing removal unworkable in #402, which is why that path hands the removal to the attaching thread instead.
- **Making the add defer symmetrically**, the mirror of #402's handoff, is the plausible fix: the add would register its intent against the claimed record, and the removing thread would perform the add after publishing phase two. It needs the record to survive phase two long enough to carry that intent, and it needs a decision about what `AddFallbackContext` returns when its work has been handed to another thread. #402 chose `true` for the deferred removal case, meaning "committed, not necessarily already applied", and the same contract would apply here.

## Impact

[...]

Issue #384, opened 2026-07-20
Lifecycle handler exceptions corrupt reconciliation state and can be fatal; make violations recoverable


RicoSuter:
## Summary

Lifecycle handlers (`ILifecycleHandler.HandleLifecycleChange`, `IPropertyLifecycleHandler.AttachProperty`) are contractually required to be exception-free. The contract is reasonable, but it is currently unenforced and uncontained: a single handler that throws corrupts the in-memory reconciliation state and propagates the exception to whatever triggered the attach, which in a hosted app can be a fatal location. This issue proposes making a violation recoverable rather than catastrophic.

## Current behavior

`LifecycleInterceptor` reconciles the object graph under `lock (_attachedSubjects)`. For each subject being attached, `AttachToProperty` / `AttachToContext` perform a multi-step, non-transactional mutation:

1. add the subject to `_attachedSubjects`
2. increment its reference count
3. invoke lifecycle handlers (`InvokeAddedLifecycleHandlers`, called bare, no try/catch)
4. raise `SubjectAttached` and attach the subject's properties

If a handler in step 3 throws:

- the lock unwinds cleanly, but the bookkeeping is left half-updated: the subject is in `_attachedSubjects` with an incremented ref count, but `SubjectAttached` never fired and its properties were never attached. The lifecycle state for that subtree is now inconsistent.
- the exception propagates out of the property write / graph construction to the caller.

Nothing between the handler and the caller catches it, so the caller inherits both the exception and a graph whose lifecycle bookkeeping no longer matches reality.

## Why a violation is catastrophic, not local

The blast radius depends entirely on who triggered the attach, and that is often a place with no isolation. Concrete example: in HomeBlaze, the root graph is built inside `RootManager.LoadAsync`, which runs from `RootManager.ExecuteAsync`, and `RootManager` is a `BackgroundService`. An unhandled exception faulting `ExecuteAsync` hits .NET's default `BackgroundServiceExceptionBehavior.StopHost`, so the entire host shuts down. One misconfigured subject (for example a property carrying both `[State]` and a secret `[Configuration]`, which `PropertyAttributeInitializer` rejects by throwing during attach) takes down every other working subject with it, with the failure surfacing far from the misconfigured attribute that caused it.

This is the worst diagnostic profile available: state corruption plus non-local fatal failure, triggered from an error path (exactly where graceful degradation matters most).

## Why the contract exists (and why it is not simply wrong)

[...]

Issue #207, opened 2026-03-02
Leaked fallback context when constructor context differs from lifecycle parent context


RicoSuter:
## Problem

When a subject is created with `new Subject(context)` and then placed under a parent whose context is **different** from `context`, the constructor's fallback is never cleaned up, causing a memory leak via `_usedByContexts`.

### Reproduction

```csharp
var rootContext = InterceptorSubjectContext.Create().WithLifecycle().WithContextInheritance();
var root = new Person(rootContext);
var parent = new Person(); // no context — inherits from root when placed
root.Children = [parent];

// Bug: child created with ROOT context, but placed under non-root parent
var child = new Person(rootContext);
parent.Children = [child];

// child has rootContext as fallback (from constructor).
// ContextInheritanceHandler does NOT add parentContext because
// IsContextAttach is false (child was already in _attachedSubjects
// from the constructor's AttachSubjectToContext call).

parent.Children = []; // remove child

// ContextInheritanceHandler tries RemoveFallbackContext(parentContext)
// but child only has rootContext → removal fails → rootContext leaks
// → root._usedByContexts retains a reference to the dead child executor forever
```

### Unit Test

In `FallbackContextLeakTests.cs`:

```csharp
[Fact]
public void WhenChildCreatedWithRootContextAndPlacedUnderNonRoot_ThenRemovalCleansUpAllFallbacks()
{
    var context = InterceptorSubjectContext
        .Create()
        .WithLifecycle()
        .WithContextInheritance();

    var root = new Person(context) { FirstName = "Root" };
    var parent = new Person { FirstName = "Parent" };
    root.Children = [parent];

    var child = new Person(((IInterceptorSubject)root).Context) { FirstName = "Child" };
    parent.Children = [child];
    Assert.Equal(1, child.GetReferenceCount());

    parent.Children = [];

    Assert.Equal(0, child.GetReferenceCount());

    var rootContext = ((IInterceptorSubject)root).Context;
    var childContext = ((IInterceptorSubject)child).Context;
    var hadLeakedFallback = childContext.RemoveFallbackContext(rootContext);
    Assert.False(hadLeakedFallback,
        "Root context still in child's fallback list after full detach — "
        + "root._usedByContexts leaks a reference to the dead child.");
}
```

### Root Cause

[...]

PR #412, opened 2026-08-03
[superseded] Own each fallback edge by the interceptors its attach resolved

### LINKED ISSUE #402: InterceptorExecutor fallback overrides: callback ordering undoes a concurrent add and blocks detach on a cyclic chain
`InterceptorExecutor` wraps both fallback mutators with lifecycle callbacks. Both re-derive at detach time what the matching attach already knew: they resolve `ILifecycleInterceptor` from the fallback context again, against a chain that may have changed or broken in between. Every defect below follows from that.

```csharp
public override bool AddFallbackContext(IInterceptorSubjectContext context)
{
    var result = base.AddFallbackContext(context);          // commit first
    if (result)
    {
        var array = context.GetServices<ILifecycleInterceptor>();   // then work that can throw
        // ... AttachSubjectToContext for each
    }
    return result;
}

public override bool RemoveFallbackContext(IInterceptorSubjectContext context)
{
    if (HasFallbackContext(context))                        // check
    {
        var array = context.GetServices<ILifecycleInterceptor>();   // work that can throw
        // ... DetachSubjectFromContext for each
        return base.RemoveFallbackContext(context);         // commit last
    }
    return false;
}
```

## 1. Remove racing add silently undoes a live registration

`RemoveFallbackContext` checks and acts non-atomically and runs the callbacks in the gap:

```
A: HasFallbackContext(ctx) -> true          committed to detaching, has not yet
C: RemoveFallbackContext(ctx)               detaches, removes
C: AddFallbackContext(ctx)                  adds, re-attaches the subject
A: DetachSubjectFromContext                 the subject is attached again, so this detaches for real
A: base.RemoveFallbackContext               removes the fallback C just added
```

C's attach is silently undone and a live fallback registration is removed. Idempotency does not save this, because the subject was legitimately re-attached between A's check and A's callback.

Reachable wherever the same subject is attached and detached concurrently. `SubjectUpdateApplier` and `SubjectItemsUpdateApplier` call `AddFallbackContext` per item, `OpcUaSubjectLoader`, `DynamicSubject` and `RootManager` call it directly, and `ContextInheritanceHandler` adds and removes on property changes.

## 2. Two threads can both run the detach callbacks

Both can pass `HasFallbackContext` and both run the callbacks. Benign only by accident: `LifecycleInterceptor` is the sole production `ILifecycleInterceptor` and its detach path happens to be idempotent under `lock (_attachedSubjects)`. Nothing in the `ILifecycleInterceptor` contract requires that of a consumer implementation, so this is a defect in its own right and not merely the mechanism behind defect 1.

## 3. A subject on a cyclic chain cannot be detached

`RemoveFallbackContext` resolves from `context` before removing the edge. Since #400 a pure delegation cycle raises `InvalidOperationException` instead of overflowing the stack, so when that chain is a cycle the resolve throws and `base.RemoveFallbackContext` is never reached.

Detaching is the natural way to recover from a cyclic chain, and it is exactly what does not work. Recovery otherwise requires registering a service on a member, which stops pure delegation.

## 4. The same input leaves a half-attached subject on the add path

`AddFallbackContext` commits the edge and then resolves, so when that resolve throws the edge is registered, no attach callback has run, and the exception leaves the method with the mutation applied.

## 5. Defect 3 also retains the subtree

`_usedByContexts` holds a reference from a context to every context that resolves through it, chaining down from the root. A subject is therefore collectable only once its fallback registration is actually removed. The failed resolve in defect 3 aborts before `base.RemoveFallbackContext`, so the edge stays, the reverse entry stays, and the detached subtree stays reachable.

## Fix

Record the interceptors the attach resolved, and make that record the ownership token for the edge. The record lives on the context, guarded by the same `_mutationLock` that publishes the topology, so the two move together. It carries a phase, so a removal cannot overtake an attach that is still running.

```
Add:    resolve (before any commit) -> publish edge + unattached record
        -> attach callbacks -> mark the record attached, from a finally

Remove: claim the record, but only if attached  (phase one, no publish)
        -> detach callbacks, best effort across the invoked prefix
        -> publish the removal, from a finally
```

A remover that arrives mid-attach does not refuse. It hands its removal to the attaching thread, which performs it once the attach completes. Refusing would strand the edge, and waiting would deadlock: the attaching thread is parked in callbacks that take `LifecycleInterceptor._attachedSubjects`, which the remover may already hold. That was measured, not assumed.

### Both callback orderings are forced

Attach callbacks must run **after** the edge is published, and detach callbacks **before** it is removed. Both resolve their handlers through the subject's own context:

```
interceptor.Attach|DetachSubjectFromContext(_subject)
  -> LifecycleInterceptor uses subject.Context, which IS the executor
    -> Invoke{Added,Removed}LifecycleHandlers -> this.GetServices<ILifecycleHandler>()
```

A generated subject's executor owns no services, so that resolve returns nothing unless the edge is in place.

### An earlier revision of this issue proposed the opposite, and it is wrong

That revision proposed making `base` the arbiter and resolving *after* the removal, and asserted in bold that the original "interceptors must be resolved before the removal" objection was wrong. **Both claims are false.** Applying that inversion fails exactly four tests in `Namotion.Interceptor.Tracking.Tests`:

- `WhenRemovingInterceptors_ThenAllChildrenAreDetached`
- `WhenRemovingInterceptors_ThenAllArrayChildrenAreDetached`
- `WhenAssigningSubject_ThenAllSubjectsAreAttached`
- `LifecycleEventsTests.SubjectAttached_FiresAfterHandler_And_SubjectDetaching_FiresBeforeHandler`

The first loses **every** detach event, because `SubjectRegistry` and `ContextInheritanceHandler` live on the parent context and become unreachable the moment the edge is gone. The earlier revision verified the wrong thing: that `context.GetServices<T>()` still succeeds after the removal. It does. What it missed is that the callback body resolves through `subject.Context`, not through `context`. Recorded here so the inversion is not proposed again.

## What this closes

| # | Defect | Outcome |
|---|---|---|
| 1 | Remove racing add | Closed. The claim is atomic, and the phase stops a removal overtaking an unfinished attach, so topology and lifecycle bookkeeping agree once writes settle. |
| 2 | Double detach | Closed by construction, no longer dependent on consumer idempotency. |
| 3 | Cyclic chain blocks detach | Closed for removability: the edge always comes out. The call can still raise, see below. |
| 4 | Half-attached subject | Closed. The resolve precedes the commit, so a throw leaves no edge and no attach. |
| 5 | Retained subtree | Closed for the fallback edge and its reverse entry. Not for the lifecycle bookkeeping, see below. |

## Declared behaviour changes

- **Detach notifies exactly the interceptors the attach resolved.** Previously the set was re-resolved at detach time, so an `ILifecycleInterceptor` registered on the parent after the attach received a detach it never saw an attach for, and one unregistered in between missed its detach.
- **An add that closes a delegation cycle now succeeds instead of raising.** The resolve moved ahead of the publish, and a pure cycle requires both ends to be empty, so there is nothing to resolve and nothing to raise. The cycle is still detected on the next resolution, and is now removable.
- **`RemoveFallbackContext` returning `true` means the removal is committed, not that the edge is already gone.** When an add is mid-callback the removal completes on that thread.

## Known gaps, tracked elsewhere

None of these is closed by this work. They are recorded so they are not lost when this issue closes.

- **Removing an edge on a cyclic chain still raises, and leaves the lifecycle bookkeeping stale.** The recorded callback resolves its handlers through a chain that no longer resolves, so `LifecycleInterceptor._attachedSubjects` and `SubjectRegistry._knownSubjects` both retain the subtree. The edge comes out either way. This is the detach half of #384 and is tracked there with a measurement.
- **A detach failure on the handed-over path is swallowed.** When a removal is completed by the attaching thread, a throwing detach handler is not reported, because the remover has already returned and the adder only asked to add. The edge still comes out. Tracked in #384.
- **An add during the claimed-record window is refused and the edge then disappears.** Between removal phase one and phase two the edge exists with no record, so a concurrent add returns `false` meaning "already registered" and the in-flight removal then drops the edge. Pre-existing in shape, but two-phase removal widens the window to the duration of the detach callbacks. Tracked separately.
- **An atomic clear-all-registrations primitive** for a final subject detach is #207. The removal primitive it needs now exists.

## Tests

- concurrent removal of the same fallback runs the detach callbacks exactly once
- a remove racing an add leaves the topology and the lifecycle bookkeeping in agreement, and no thread removes an edge whose record it does not own
- a removal that runs during the attach callbacks still removes the edge, verified by mutation: reverting the handoff to a refusal must fail this test
- detaching a subject whose chain turned cyclic after the record was captured removes the edge and raises
- an add whose pre-commit resolve raises registers no edge
- an add that closes a delegation cycle registers the edge and does not raise
- an attach that throws part way detaches only the prefix it invoked
- a throwing detach interceptor does not prevent the ones behind it from being detached
- an acyclic detached subtree becomes collectable, by weak reference probe
- both callback orderings are observable from inside the callbacks

Related: #401 (the cyclic shape defects 3 and 4 need), #400 (turned the crash into the catchable exception that made 3 and 4 observable), #384 and #207 (see the gaps above).

RicoSuter:
Closes #402.

`InterceptorExecutor` re-derived at detach time what the matching attach already knew: it resolved `ILifecycleInterceptor` from the fallback context again, against a chain that may have changed or broken in between. All five defects in #402 come from that.

Each edge added through the executor now carries a record of the interceptors its attach resolved, and that record is the ownership token for the edge. It lives on the context guarded by the same `_mutationLock` that publishes the topology, so the two move together.

```
Add:    resolve (before any commit) -> publish edge + unattached record
        -> attach callbacks -> mark the record attached, from a finally

Remove: claim the record, but only if attached  (phase one, no publish)
        -> detach callbacks, best effort across the invoked prefix
        -> publish the removal, from a finally
```

## Two orderings that turned out to be forced

Attach callbacks must run **after** the edge is published, detach callbacks **before** it is removed. Both resolve their handlers through the subject's own context:

```
interceptor.Attach|DetachSubjectFromContext(_subject)
  -> LifecycleInterceptor uses subject.Context, which IS the executor
    -> Invoke{Added,Removed}LifecycleHandlers -> this.GetServices<ILifecycleHandler>()
```

A generated subject's executor owns no services, so that resolve returns nothing unless the edge is in place.

This matters because **#402's own prescribed fix was the opposite and would have silently lost every detach event.** Inverting the remove path fails exactly four tests in `Namotion.Interceptor.Tracking.Tests`, and `WhenRemovingInterceptors_ThenAllChildrenAreDetached` loses every event: `SubjectRegistry` and `ContextInheritanceHandler` live on the parent context and become unreachable the moment the edge is gone. The issue has been amended to record this so the inversion is not proposed again. Both orderings are now pinned by a test that observes them from inside the callbacks.

## The phase, and why a removal hands over instead of refusing

A record carries a phase, so a removal cannot overtake an attach that is still running. Without it this diverges:

```
C: publish edge and record
A: claim record
A: detach callbacks run, no-op because C has not attached yet
A: remove edge
C: attach callbacks run
```

Edge absent, subject attached. AGENTS.md ranks quiescent consistency under correctness, so that is a defect rather than an accepted anomaly.

A remover arriving mid-attach does **not** refuse, because refusing strands the edge, and the window is wide: the attaching thread is parked on `LifecycleInterceptor._attachedSubjects`, which the remover typically holds, so it stays parked for the whole lock hold. It hands the removal to the attaching thread instead, which performs it once the attach completes. Waiting would deadlock against the same lock.

This is verified by mutation: reverting the handoff to a refusal fails `WhenRemoveRunsDuringAttachCallbacks_ThenTheEdgeIsStillRemoved` on `Assert.True(removed)`.

## What this closes

| # | Defect | Outcome |
|---|---|---|
| 1 | Remove racing add undoes a live registration | Closed. Atomic claim plus the phase, so topology and bookkeeping agree once writes settle. |
| 2 | Double detach | Closed by construction, no longer dependent on consumer idempotency. |
| 3 | Cyclic chain blocks detach | Closed for removability: the edge always comes out. The executor resolves nothing on the detach path, but a recorded interceptor still resolves through the subject's own context and can raise, so the call can still throw. |
| 4 | Half-attached subject on the add path | Closed for a failing resolve: it precedes the commit, so a throw leaves no edge and no attach. A throwing attach *callback* still leaves the edge committed and the subject partly attached, which is #384's rollback problem. |
| 5 | Retained subtree | Closed for the fallback edge and its reverse `_usedByContexts` entry. |

## Behaviour changes

- **Detach notifies exactly the interceptors the attach resolved.** Previously the set was re-resolved at detach time, so an `ILifecycleInterceptor` registered after the attach received an unpaired detach, and one unregistered in between missed its detach.
- **An add that closes a delegation cycle now succeeds instead of raising.** The resolve moved ahead of the publish, and a pure cycle needs both ends empty, so there is nothing to resolve and nothing to raise. The cycle is still detected on the next resolution and is now removable.
- **`RemoveFallbackContext` returning `true` means the removal is committed, not that the edge is already gone.** When an add is mid-callback it completes on that thread.
- **A throwing detach handler no longer blocks the removal.** Previously the first exception propagated before the edge was removed and the edge stayed. Now every recorded interceptor is notified, the edge drops from a `finally`, and the first exception is rethrown afterwards.

## Known gaps, tracked and not edited away

- **#411**: an add during the two-phase removal window is refused and the edge then disappears. Pre-existing in shape, widened here to the duration of the detach callbacks.
- **#384**: removing an edge on a cyclic chain still raises and leaves `_attachedSubjects` and `SubjectRegistry._knownSubjects` retaining the subtree, because the handlers are unreachable from every context on the loop. Measured there. A detach failure on the handed-over path is also swallowed.
- **#207**: the atomic clear-all primitive. The removal primitive it needs now exists.

## Performance

Measured with `scripts/benchmark.ps1 -Filter "*RegistryBenchmark*" -Stash -LaunchCount 3` against master.

| | branch | master | delta |
|---|---:|---:|---|
| AddLotsOfPreviousCars allocated | 19,808,842 B | 19,528,737 B | +280 KB, +1.4% |
| ChangeAllTires allocated | 14,336 B | 14,112 B | +224 B, +1.6% |

The allocation figures reproduce to the byte across two independent runs, and `ChangeAllTires` works out to roughly 56 B per edge against 48 B for the `FallbackAttachment` node, so the record costs what it should with nothing unexplained on top.

Timing showed no reliable difference in either direction. An earlier single-launch run on a busy machine suggested regressions of 8 percent and 6 percent; both inverted under three launches on a quiet one, and untouched code paths such as `GetOrAddSubjectId` moved by up to 15 percent in the same run, which sets the noise floor. Neither figure should be read as a result.

A follow-up comment describes a memory optimisation that would remove the per-edge *allocation* for the single-edge case, which is roughly 40 percent of the byte delta rather than all of it, kept separate so it can be measured on its own.

## Tests

Eleven tests in `FallbackAttachmentOwnershipTests` and one in `FallbackContextLifecycleConcurrencyTests`, plus two fuzz model corrections. The last of those runs a concurrent add/remove toggle against the real `LifecycleInterceptor` and `ContextInheritanceHandler` over 400 rounds and asserts the attach and detach balance matches the final topology, which is the only coverage here that exercises the production stack rather than test doubles. Non-integration suite across 26 assemblies: **2664 passed, 0 failed, 0 warnings**. Public API snapshot unchanged.

`ContextConcurrencyFuzzTests` needed both of its model sites updated, since a failed add now leaves no edge where it previously left one. With only the runtime site changed, 5 of 8 seeds still fail.


RicoSuter:
## Memory optimisation, deliberately not in this PR

Kept separate so it can be measured on its own. Correctness is already pinned by the suite, so the optimised variant has exactly one question to answer.

*Corrected after review: an earlier version of this comment claimed a variant that saves nothing, and overstated the other. Numbers below are arithmetic you can check, not estimates.*

### What this branch costs today

```
FallbackAttachment field region (x64)
  Context                    ref    8
  Interceptors               ImmutableArray<T> = 1 ref    8
  InvokedInterceptorCount    int    4
  IsAttachCompleted          bool   1
  IsPendingRemoval           bool   1
  Next                       ref    8
                                   30 -> aligned 32, + 16 header = 48 bytes
```

Plus one 8 byte head field per context, paid whether or not the context ever holds a record.

The measured `ChangeAllTires` delta of +224 B decomposes exactly: four new subject executors, one edge each, `4 x 48 + 4 x 8 = 224`. Nothing unexplained.

### Variant A: inline the first attachment

The dominant shape is exactly one fallback per subject, so put the first attachment in fields on the context and allocate nodes only for the rare second and beyond.

**Per single-edge subject: 56 B and one allocation, down to about 32 B and no allocation.** The allocation going away is the real prize, since AGENTS.md ranks allocations above CPU. On `ChangeAllTires` that is roughly +224 B down to about +128 B, so **it removes something like 40 percent of the delta, not all of it.** An earlier version of this comment said "+0 B", which was wrong: it ignored that the inline fields are paid by every context, including the many that never hold a record.

It forces an API change, and that is the real cost. Today `TryBeginFallbackAttachment` returns the record and the executor holds that reference across the callbacks. With the first slot inlined there is no object to hand back, so the four operations have to become identity based, looking the record up by the fallback context the executor already has:

```csharp
private protected bool TryBeginFallbackAttachment(
    InterceptorSubjectContext contextImpl, ImmutableArray<ILifecycleInterceptor> interceptors);

private protected bool CompleteFallbackAttachment(
    InterceptorSubjectContext contextImpl, int invokedInterceptorCount,
    out ImmutableArray<ILifecycleInterceptor> interceptors, out int detachCount);

private protected FallbackRemovalOutcome TryTakeFallbackAttachment(
    InterceptorSubjectContext contextImpl,
    out ImmutableArray<ILifecycleInterceptor> interceptors, out int invokedInterceptorCount);

private protected void CompleteFallbackContextRemoval(InterceptorSubjectContext contextImpl);
```

Every operation then gains an inline-slot-or-node branch, roughly doubling the ownership code. That is why it is not in this PR: this is the code that took five review rounds and a prototype to get right, and the defects found along the way were a field pairing race, a stranding, and a guard whose own failure path was unrecoverable.

Note also that the "two edges" case is rarer than it looks. A subject constructed with an explicit context and then placed under a parent ends up with **one** edge, not two: it is already in `_attachedSubjects` from the constructor, so `IsContextAttach` is false and `ContextInheritanceHandler` adds nothing. That is the #207 leak. So the inline slot covers even more of the population than the single-edge argument suggests.

### Variant B: packing the node saves nothing, do not attempt it

An earlier version of this comment proposed packing `InvokedInterceptorCount` and the two flags into a single `int` to take the node from 48 to 40 bytes. **That saves zero bytes.**

```
8 + 8 + 4 + 8 = 28 -> aligned 32, + 16 header = 48
```

The field region is already padded to 32, so removing 2 bytes of flags changes nothing. Reaching 40 requires eliminating a whole reference field, not packing the scalars. The only candidate is `Next`, which means replacing the linked list with an array and paying a resize instead. Variant A is the better use of the same effort.

### How to measure

Branch from this one so the diff is the optimisation alone, then:

```
pwsh scripts/benchmark.ps1 -Filter "*RegistryBenchmark*" -Stash -LaunchCount 3
```

Read **Allocated** on `AddLotsOfPreviousCars` and `ChangeAllTires` and treat those as the result: allocations here are deterministic and reproduced to the byte across runs. Treat the timing columns with suspicion unless the machine is quiet. On a busy machine this PR appeared to regress attach by 8 percent and `Write` by 6 percent; under three launches on a quiet one both inverted, while untouched paths such as `GetOrAddSubjectId` moved by up to 15 percent in the same run.

The suite must stay green with the public API snapshot unchanged. If any of `FallbackAttachmentOwnershipTests` needs adjusting to accommodate the variant, that is a signal the behaviour moved, not the storage.

PR #414, opened 2026-08-04
perf: Batch OPC UA browse operations during subject loading

### LINKED ISSUE #320: OPC UA loader: defer root-subject value mutations for atomic observer view

RicoSuter:
The OPC UA client loader browsed one node per round-trip. It now discovers the address space breadth-first and browses each level as one batched call, so the number of round-trips grows with the depth of the address space rather than with the number of nodes. It also gives a load a defined outcome: every ownership claim and every property binding is queued during discovery and applied by one commit at the end, so the model changes in one burst and a load that fails before that commit leaves it as it found it. That burst at the end of the load is what #320 asked for. On the way it fixes a live reconnect-after-dispose bug in the subscription manager and closes the window in which a notification could write into a subject that detached mid-setup.

Consolidates #397, #398, #357 and #313, all closed in its favour.

- Closes #320
- Related to #416, the pre-existing integration flakes, not caused here
- Follow-up: the staging-link case under Known limitations, which needs a fix in the core rather than in this connector

## Why

Round-trips, not bytes, dominate load time on a real link. At 5 ms RTT the old loader spent about 5.4 s browsing that address space, at 20 ms about 22 s. Batching brings both under 200 ms.

The four branches could not be judged apart. The browse-completion contract in #397 has no caller without the loader, its continuation cap needed a behaviour change motivated by the loader's measurements, and #313 and #357 shared roughly 90 percent of their tests, so a split review would have read the same lines two or three times.

### Review order

Thirty-five commits including three merges of master, each building and passing on its own. The first five are the change, the next batch fixes defects an independent review found afterwards, and the rest are follow-ups from the reviews of this pull request.

1. `9af09f53` The subscription manager no longer resets its terminal shutdown flag on reconnect. A reconnect racing disposal used to resume notifications on a disposed manager.
2. `e6e2e1b0` Batched browse and read primitives, plus the split between session-permanent status codes and load-skip codes.
3. `749ff8c2` Data-change callbacks gated until subscription setup completes, with a detached-subject sweep before read-after-write registration.
4. `b7a66b4f` The two traversal bounds, since renamed to `MaxBrowseContinuationRounds` and `MaxAttributeTraversalDepth`.
5. `77f74f4d` The loader itself: breadth-first discovery, a per-load staging context, deferred root mutations.
6. `6a55c7e2` onwards: the rollback deadlock and zombie-subject fixes, a production-shaped failure fixture that can actually reach them, the mid-setup detach set, and the rollback semantics stated under Contract. `ddeb4ce7` pins the deadlock at integration level.
7. `4f012f85` defers every binding to the commit and completes the rollback, `2909647a` recurses once per level, `8fcc3446` tightens and deduplicates the loader tests, `38ea31c3` names the options and the status predicates by what they bound, `b1228668` drops the subscription manager's test seams and its registrar interface, `92cb464b` documents the loader's contract and moves the design rationale into `docs/design/opcua-client-loader.md`, `569cfe8a` applies the final review's assertion strengthening and comment trims, and `fdc322ee` cuts the design document to what the code cannot say.

### Design notes

- Subjects stay attached while being discovered. Property mapping, dynamic property creation and attribute matching all go through the registry's `RegisteredSubject`, which exists only for an attached subject, and `RegisteredSubjectProperty.Children` is what lets the loader reuse an existing child instead of rebuilding the subtree. Discovering off-graph would lose both.
- Nothing the load changes is applied while browsing. Claims and bindings are queued and applied by `OpcUaLoadContext.Commit`, claims first so an observer that sees a new child finds its leaves already source-owned, then the bindings deepest level first so a subject is unreachable from the root until its own subtree is bound. Only dynamic properties and attributes are created eagerly, because the registry has to hand back the property that is then browsed and monitored.
- Each level recurses once, over the union of the children its phases discovered. The container and reference phases no longer recurse on their own, so browse calls grow with depth and attribute depth rather than with the number of branch kinds along a path.
- Removing a detached subject's monitored items on the client source is lock-free. It runs from the synchronous detach callback under the lifecycle interceptor's lock, and taking the structure lock there deadlocked a failed load's own rollback and inverted against external detaches. Both dictionaries are concurrent and the removals are idempotent.
- The continuation-point cap applies only when `MaxReferencesPerNode` is non-zero, the only case a browse can leave continuation points open. `BadNoContinuationPoints` releases the points the server issued, halves the batch and retries.
- Re-keying a staged subject's fallback context to the parent that binds it was implemented and rejected. Staging-parent keying is what keeps the fallback chain acyclic; the re-key makes mutually referencing subjects point at each other and fails the address-space-cycle test with a delegation-cycle error.
- The reasoning behind all of the above lives in `docs/design/opcua-client-loader.md`. The code comments state the rule, name the one fact that guards against the wrong edit, and point to the design document for the rest.

## Contract

- A node present in a browse result was browsed to completion. A node whose pagination did not finish, or that exceeded `MaxBrowseContinuationRounds`, is omitted, and the loader keeps that property's current items and retries on the next load. A subject whose own browse did not complete is skipped for the load for the same reason.
- A load applies all of its ownership claims and property bindings at the end, claims first, then the bindings deepest level first, so the model changes in one burst rather than progressively during browse IO.
- A load that fails before the commit leaves the model at its pre-load state, except for the dynamic properties and attributes discovery adds eagerly. Those are transient rather than torn: each carries its exact node ID, so the next load re-matches and re-monitors it.
- A failure inside the commit releases the claims that commit established and restores the bindings it applied. Ownership from a previous successful load is never released, and a property the application has re-set since keeps the newer value.
- No data-change notification reaches a subject before subscription setup completes. The initial state read that follows supplies any value a dropped notification carried; notifications after the gate opens are buffered and replayed.
- Subscription manager disposal is terminal. A reconnect that races disposal never resumes callbacks.
- A transient browse or read status aborts the load with `OpcUaTransientServiceException`. The source catches it, rolls back and retries, so consumers only see it through `Diagnostics.LastError` and the source event stream.

## Breaking changes

**API.** `OpcUaTypeResolver` is reshaped, because resolving one node per call cannot be batched and its members had disagreed on where the session went and what the node argument was called. Every seam now takes a context, and each per-node context carries the session, the node and its resolved `NodeId`:

| Was | Is |
|---|---|
| `TryGetTypeForNodeAsync(session, reference, ct)` | `ResolveVariableNodeTypeAsync(OpcUaVariableNodeContext, ct)` for one variable, `ResolveObjectNodeTypeAsync(OpcUaObjectNodeContext, ct)` for one object, `ResolveVariableNodeTypesAsync(session, variables, ct)` to replace the batched read itself, `TryMapBuiltInType(builtInType)` for a per-built-in-type rule |
| `GetDynamicPropertyAttributes(reference, session)` | `GetAttributesForDynamicProperty(OpcUaDynamicPropertyContext)` and `GetAttributesForDynamicAttribute(OpcUaDynamicAttributeContext)` |

In-repo impact: `HomeBlazeOpcUaTypeResolver` stamped its attribute through the single decoration method, which served both dynamic properties and dynamic attributes, so it now overrides both to keep doing that. The updated sample is in `docs/connectors-opcua-client.md`.

**Behavior.**

- Collection, dictionary and subject-reference assignments now appear to observers as one burst at the end of the load, at every level, instead of progressively during browse IO.
- A dictionary entry's key is now the bracket content of the child's browse name, so `Items[1]` yields the key `1`; a child whose browse name has no brackets uses that full browse name as its key.
- A transient browse or read status now aborts the load and the source retries it. Master's loader did not classify browse statuses.
- The initial state read now batches at 256 nodes when the server reports no read limit, and splits an oversized batch on `BadTooManyOperations` instead of failing the load.
- Browse log templates now start with the operation name, so a log filter matching the old prefix needs updating.
- A data change notification is dropped until subscription setup completes. Code that drives a `SubscriptionManager` directly, which in practice means tests, has to complete setup before delivering one.

**Contract.** None.

Additive public API, all in the snapshot diff: `OpcUaTransientServiceException`, `OpcUaClientConfiguration.MaxBrowseContinuationRounds` and `MaxAttributeTraversalDepth` (both default 100, validated positive), and the four context types the resolver seams take. `ResolveObjectNodeTypeAsync` is asynchronous so an override can read the server to classify a node, though the base never does.

## Performance

No BenchmarkDotNet benchmark reaches the OPC UA loader, so the measurement is browse round-trips, taken with a temporary counter around `BrowseAsync` and `BrowseNextAsync` against the in-process test server (`MaxNodesPerBrowse = 4000`, `MaxBrowseContinuationPoints = 100`) on an address space of 241 subjects and roughly 1,000 value properties. Round-trip counts are deterministic, so no noise reference applies, and wall clock on loopback is noise; the latency rows multiply the count by a link RTT. The cost is one extra node submitted per load and the per-load staging context.

The table below was measured before the level structure changed. Recursing once per level rather than once per branch kind removes calls that the table's tree shape never charged for: on the unit fixture of three container levels that each carry both a collection and a subject reference, the loader makes 8 browse calls where the earlier structure of this pull request made 23. Three batching tests pin exact call counts on their fixtures: `WhenNodeCountExceedsMaxNodesPerBrowse_ThenBrowseIsChunked` asserts three calls with none above the configured two nodes, `WhenLoadingMultiLevelTreeWithMixedNodeTypes_ThenAllMonitoredItemsAndStructureAreCorrect` asserts five calls, and `WhenEveryLevelCarriesACollectionAndASubjectReference_ThenBrowseCallCountGrowsWithDepthNotWithBranches` asserts the node count of every one of its eight calls, so the improvement cannot silently regress to one node per call.

| Measurement | Before | After | Delta |
|---|---:|---:|---:|
| Browse round-trips | 1,083 | 9 | -99% |
| Nodes submitted | 1,083 | 1,084 | +1 |
| Browse latency at 5 ms RTT | ~5.4 s | ~45 ms | |
| Browse latency at 20 ms RTT | ~22 s | ~180 ms | |

## Diff composition

Against current master.

|---|---:|---:|---:|---:|
| Production code | 19 | 2'416 | 626 | +1'790 |
| Tests and benchmarks | 21 | 6'139 | 832 | +5'307 |
| Documentation | 3 | 139 | 13 | +126 |
| **Total** | **43** | **8'694** | **1'471** | **+7'223** |

## Verification

- [x] Documentation updated: `docs/connectors-opcua-client.md` (custom type resolver, the two new browsing options, load failure and rollback, deduplication collisions, lifecycle and the components table), `docs/connectors-opcua.md`, and the new `docs/design/opcua-client-loader.md`.
- [x] Unit tests: 4,069 passing across 34 assemblies, zero failures.
- [x] Integration tests: `Namotion.Interceptor.OpcUa.Tests`, 462 passing including integration, zero failures.
- [ ] ~~Benchmarks~~ no BenchmarkDotNet benchmark reaches the loader; see Performance.
- [ ] Load tests: Connector Tester not run.
- [ ] Chaos tests: Connector Tester not run.

## Known limitations

- The staging link can survive a successful load in a graph-shaped address space. `ContextInheritanceHandler` adds a subject's inherited fallback context keyed to the parent that gave it its first reference and removes it keyed to the parent that held the last one, so for a subject reachable under two parents the removal does not match the addition and a link to the first parent's context remains. This is the core's existing asymmetry, not something the loader introduces, and it costs a retained context reference rather than a wrong model: the subject stays registered, monitored and correct. The fix belongs in the core, by recording the inherited context per subject instead of re-deriving it from the last parent, and is a follow-up.

## Follow-ups

None of these is a regression against master; each needs its own design decision.

- A staged child that the commit binds under a subject the application detached during the load stays registered and monitored as an orphan, because the parent's detach cascade ran before the binding existed.
- After a re-listen, properties that keep ownership from an earlier load but whose subject was skipped or whose container browse failed have no node and swallow application writes as success instead of parking them.
- A reference back to the root's own node creates and stages a second subject for it instead of reusing the root, because the root is not seeded into the per-load NodeId cache.
- The subscription manager test harness never disposes the read-after-write and polling managers it builds; nothing runs in them today.
- The core `ContextInheritanceHandler` keying asymmetry under Known limitations.


src/Namotion.Interceptor.OpcUa/Client/OpcUaClientConfiguration.cs
```diff
@@ -207,6 +207,17 @@ public class OpcUaClientConfiguration
...
+    /// <summary>
+    /// Gets or sets the maximum number of BrowseNext continuation rounds per browse. Bounds pagination depth
+    /// so a server that returns a fresh continuation point forever cannot loop the loader. Default is 100.
+    /// </summary>
+    public int MaxBrowseContinuations { get; set; } = 100;
+
+    /// <summary>
+    /// Gets or sets the maximum attribute-traversal depth (attributes of attributes) during loading. Default is 100.
+    /// </summary>
+    public int MaxAttributeTraversals { get; set; } = 100;
```
RicoSuter: we should probably document what happens after this max? same for MaxBrowseContinuations

src/Namotion.Interceptor.OpcUa/Client/OpcUaSubjectClientSource.cs
```diff
@@ -788,16 +812,22 @@ private void Reset()
...
-        try
-        {
-            _sessionManager?.SubscriptionManager.RemoveItemsForSubject(subject);
-            _sessionManager?.PollingManager?.RemoveItemsForSubject(subject);
-        }
-        finally
-        {
-            _structureLock.Release();
-        }
+        // Lock-free by design. This runs from the synchronous subject-detach callback, which the
```
RicoSuter: compact docs to 2-3 lines or so, just keep what is essentiall and not clear from code

src/Namotion.Interceptor.OpcUa/Client/OpcUaTypeResolver.cs
```diff
@@ -30,110 +28,107 @@ public virtual Attribute[] GetDynamicPropertyAttributes(ReferenceDescription ref
...
                 }
             }
-        }
-        catch (Exception ex)
-        {
-            _logger.LogDebug(ex, "Failed to infer CLR type for node {BrowseName}", reference.BrowseName.Name);
+            catch (Exception ex)
+            {
+                _logger.LogWarning(ex, "Failed to infer CLR type for node {BrowseName}.", reference.BrowseName.Name);
+            }
```
RicoSuter, line 167: should we maybe extract this as virtual ResolveVariableTypeAsync (with all params for max extensibility) and keep ResolveVariableTypesAsync non-virtual? better extensibility point or not?

src/Namotion.Interceptor.OpcUa/Client/OpcUaClientConfiguration.cs
RicoSuter: Both options now document what happens at the limit. A node still paginating when MaxBrowseContinuationRounds is reached is omitted from the browse result, so the loader keeps that property's current value and reloads the node on the next load. At MaxAttributeTraversalDepth the remaining pending attributes are skipped for that load and everything matched in earlier rounds stays monitored. The same is in the browsing options table in docs/connectors-opcua-client.md. Both were also renamed, since they bound BrowseNext rounds and attribute depth rather than continuation points and traversals.

src/Namotion.Interceptor.OpcUa/Client/OpcUaSubjectClientSource.cs
RicoSuter: Trimmed to three lines: it states that the method is lock-free because it runs from the synchronous detach callback under the lifecycle interceptor's lock, so taking the structure lock there deadlocks, and points at docs/design/opcua-client-loader.md for the two deadlock shapes.

src/Namotion.Interceptor.OpcUa/Client/OpcUaTypeResolver.cs
RicoSuter: Done in 24579e7c, with one deviation from the suggestion. Extracted a protected virtual ResolveVariableTypeAsync that decides one node from the attributes the batch already read, so an override no longer has to reimplement the batching, the positional alignment of the two attributes per node or the transient status handling.

It takes an OpcUaVariableTypeContext (session, reference, resolved node id, DataType, ValueRank) rather than a positional parameter list, so adding to what a resolver is given later does not break existing overrides.

I kept ResolveVariableTypesAsync virtual instead of sealing it. A resolver whose types come from a model file rather than the server wants to replace the batched read entirely, and sealing that buys nothing; the XML docs on both now say which seam to use. A test covers that a subclass overriding only the per-node hook still gets one batched read of all attributes and the base mapping for the nodes it does not claim.

RicoSuter: Renamed for symmetry in cd85d906. My earlier objection was to ResolveNodeTypeAsync being too generic, since Object nodes go through a different method with different inputs, but ResolveVariableNodeTypeAsync is precise and symmetric, so you were right.

The batch method moved with it, otherwise the singular and plural stopped reading as a pair. The three seams are now ResolveObjectNodeType, ResolveVariableNodeTypesAsync and ResolveVariableNodeTypeAsync, and the context struct is OpcUaVariableNodeContext, since it describes the node rather than a type.

Issue #410, opened 2026-08-03
ContextInheritanceHandler strands fallback edges on a partial detach


RicoSuter:
Split out of #401, which is closed by #400. #400 fixed the symptom named in that issue's title, the uncatchable `StackOverflowException` on a pure delegation cycle, which now raises a catchable `InvalidOperationException` instead. It did not touch the mechanism that builds those cycles, and it did not address the second symptom #401 recorded. This issue carries both forward so they are not lost with the close.

## Root cause

`ContextInheritanceHandler.HandleLifecycleChange` removes an inherited fallback only on `{ ReferenceCount: 0, IsPropertyReferenceRemoved: true }`:

```csharp
else if (change is { ReferenceCount: 0, IsPropertyReferenceRemoved: true })
{
    change.Subject.Context.RemoveFallbackContext(change.Property.Value.Subject.Context);
}
```

A detach that leaves property values set therefore strands the fallback edge. Orderings exist where a mutually referencing subject pair is orphaned from the root and then partially detached, leaving `A.Context` with `[B.Context]` and `B.Context` with `[A.Context]` and no services on either.

## Symptom 1: the graph builds a cycle nobody asked for

Once both contexts are pure delegators pointing at each other, every service resolution and every intercepted property access on either subject raises. Before #400 the same shape killed the process, so this is strictly better, but the graph still ends up in a state the consumer never constructed and cannot easily diagnose.

A cycle with at least one service on either side is fine and must stay fine: the delegation fast path does not apply and the service walk's visited set terminates it. That behaviour is pinned by `ContextDelegationCycleTests.WhenCycleContainsContextWithService_ThenResolvingSucceeds`.

## Symptom 2: stale resolution through a detached context

The same stranded edge lets a still-attached subject keep a fallback to a fully detached context and keep resolving services through it. This one produces no exception at all: the subject silently resolves against a context that is no longer part of the graph. Nothing in #400 changes it, and no other open issue covers it.

## Relationship to the other fallback issues

- #207 is the memory consequence of the same class of stranded edge: the reverse `_usedByContexts` entry retains the detached child.
- #402 hardens the removal path itself (callback ordering, exactly-once detach, and behaviour when the context being removed is on a cycle). Its "related fix scope" already calls for a shared removal primitive and names #207.

[...]

Issue #210, opened 2026-03-05
LifecycleInterceptor: Minor memory leak in dynamic property removal


RicoSuter:
## Summary

`LifecycleInterceptor._lastProcessedValues` entries are not cleaned up when a dynamic property is removed from a subject that remains attached.

## Details

`_lastProcessedValues` tracks the last reconciled value per structural property for thread-safe lifecycle reconciliation. Entries are cleaned up when the owning subject is fully detached (ref count → 0), via both `DetachFromContext` and `DetachFromProperty`.

However, if a dynamic property (e.g. added via `AddDerivedProperty`) is **removed** while the subject is still attached, the corresponding `_lastProcessedValues` entry remains in the dictionary. The entry is inert — no `WriteProperty` call will ever target a removed property — so it has no correctness impact. It just occupies a small amount of memory until the subject is eventually detached.

## Impact

Negligible unless a long-lived subject has many dynamic properties added and removed over time. Each orphaned entry holds a `PropertyReference` key and an `object?` value reference.

## Fix

Hook into the property removal path and call `_lastProcessedValues.Remove(property)` when a dynamic property is removed from an attached subject.

## Related

Introduced in PR #208 (`fix/lifecycle-thread-safety`).

PR #419, opened 2026-08-04
Redesign subject ownership around explicit roots

RicoSuter:
## Summary

This PR is the second change in the context and lifecycle simplification stack. It is stacked on #474 and redesigns lifecycle ownership around explicit roots, parent membership, one ownership domain, and one effective ownership route.

The current branch contains the approved design, roadmap refinement, and independently reviewed implementation plan. Production implementation follows on this same PR.

## Contract

- `AddFallbackContext` and `RemoveFallbackContext` become service composition only.
- `AttachToContext` and `DetachFromContext` are the strict explicit-root lifetime API.
- A generated or Dynamic `new Subject(context)` is an explicit attachment and requires an exact detach.
- A parameterless property child is owned automatically through parent membership and releases after its final parent leaves.
- Registry, parent tracking, hosting, and lifecycle callbacks are projections and reactions. They are not the source of ownership.
- One owned subject has one exact configured-context domain and one effective route. Incompatible property assignment fails before the backing property changes.
- Multiple references, shared DAGs, and cycles remain supported.
- Public fallback composition and late nonunique service additions remain supported.
- Optional context inheritance is removed. `WithLifecycle()` provides the complete recursive lifecycle capability.

## Compatibility and performance

The ordinary one-global-context callback order and callback-time service visibility remain compatibility contracts. The scalar generated setter and stable read, invoke, and service-resolution paths must keep their existing fast shapes. Warm route-free and owned structural writes must not add managed allocations beyond configured custom interceptor work.

Performance must be the same as or better than master for the normal one-global-context use case. Local numbers are diagnostic only. Final comparison, solution build/pack if needed, and Connector Tester verification are handed to the stable benchmark machine before this PR is declared ready.

## Implementation plan

The approved plan has seven review gates:

1. Centralize structural property classification in Core.
2. Atomically introduce the Core ownership protocol, prospective structural admission, generated/Dynamic constructor cutover, and Tracking lifecycle integration.
3. Complete deterministic parent routes, cycles, and component release.
4. Harden concurrency, reentrancy, callback failure, and lifetime behavior.
5. Migrate first-party roots, children, samples, benchmarks, Connector Tester, and HomeBlaze.
6. Make connector factories route-free and publish children through parents before recursive population.
7. Complete permanent documentation, full verification, independent stack review, cleanup, and stable-machine handoff.

Tasks with implementation uncertainty have explicit stop, amend, and re-review checkpoints. The Core provider interface, write terminal, generated/Dynamic constructors, fallback semantics, and Tracking coordinator are one atomic task so the branch never needs a temporary compatibility adapter.

## Current verification

- The design passed three independent written-spec review rounds.
- The implementation plan passed an independent file-scope audit and a separate strict execution review with no remaining Critical or Important findings.
- The branch is based on exact #474 head `a88b456ef681dc4505f1edce040b56fb83a6a034`.
- Current design/plan head: `a7d0c94893d143557c9b897b831120eddbec5516`.
- No production code is changed yet, so performance and connector gates are intentionally pending.

The temporary `docs/superpowers/` roadmap, spec, plan, and task reports remain only while the stack is implemented and reviewed. They will be removed from both #474 and this PR before merge. Permanent user and design documentation will live in the normal `docs/` pages.


RicoSuter:
## Follow-up work

Rewritten to match the branch as it stands. Originally 21 items from six review passes; ten have since closed and are listed at the bottom so nobody re-does them. The companion comment holds the open *questions*, as opposed to the work items here.

### Before merge

1. **Run `RegistryBenchmark` on master against this branch.** Every performance statement in the description is currently a prediction. Read `ChangeAllTires` and `AddLotsOfPreviousCars` for time and allocated bytes, with `Write` and `Read` as the hot-path control. Expected signature: measurably fewer allocated bytes at flat or slightly better time, because every attach and detach previously boxed an int into a `ConcurrentDictionary` with a tuple key.

### Issue updates (the plan deliberately stopped short of these)

| Issue | Action |
|---|---|
| #207 | Close. Both reproductions verified: they fail against pre-change production code at `dbfb8984` and pass at HEAD. |
| #402 | Update, keep open on defect 1. Defects 2, 3, 4 and 5 close. Defect 1 is narrowed from every child detach to an explicit concurrent root operation. |
| #210 | Close as not reachable. Record the constraint that a future property-removal API must clear the reconciliation ledger entry itself. |
| #411 | Update, keep open. The loud form now throws instead of returning `false` and losing the edge, but one shape stays silent: if the attach context has had its lifecycle services removed since the attach, the guard passes on the empty set, the dedup returns `false`, and the detach then removes the edge. |
| #410 | Update, keep open. Symptom 2 is not closed. Post all three measurements, including the correction that the subject resolves nothing rather than continuing to resolve through the departed context. |
| #384 | Update, narrowed. The stated blocker is removed. Note that "inheritance is no longer a handler" is imprecise: `ContextInheritanceHandler` is still an `ILifecycleHandler`; what changed is that no edge mutation runs user code. |
| #412 | Close unmerged, referencing this work. |

### Benchmarks

2. No benchmark covers the constructor-attached-then-parented shape, which is the one shape that regresses: it holds two edges, stops being a pure delegator, and keeps its own service cache and compiled chains instead of sharing the terminal's. Roughly 1.5 to 3 KB per affected subject. Nothing existing will ever see that cost.
3. Unrelated but noticed: every `[Benchmark]` attribute in `DynamicSubjectBenchmark` is commented out, so it measures nothing.

### Test coverage

4. `WhenTwoThreadsDetachTheSameRoot_...` is single-shot, so it only catches its regression when the two racers actually overlap. Every other race test on this branch loops fifty rounds.
5. The "already attached through a different context" throw is unprotected for core-only consumers: for Tracking graphs the ownership throw shadows it, but with a custom core-only `ILifecycleInterceptor` no owner is claimed, so deleting it silently overwrites the record and leaks the first edge.
6. No direct test that a two-parent child keeps resolving after losing one reference while the linked parent stays.
7. The documented root allowance in `ThrowIfDetachIsUnwinding` (a root holds no link, so re-attaching one during a root detach does not throw) is untested.
8. No public-seam race test for a property write racing a root attach or detach.
9. **New.** The dark-subject gap is reachable by a plainer route than the pinned test uses. That test detaches a root through `DetachFromContext`; simply removing a child from a collection produces the same dark state, and that is the shape the documentation diagram shows. Same gap, unpinned route.

### Closed since this comment was first written

- Explicit composition resolving before the inherited parent link, which nothing pinned (`9c882af8`).
- The wrong-context detach throw, the attach rollback's edge removal, and the connector-shape test that missed its adjacent mutant, all three now verified against their own mutants (`3f34fc74`).
- Three factually wrong statements in the design document, plus six comments citing sections of a specification this branch removed (`3f34fc74`).
- The lock-ordering section's two false claims, and the deadlock now recorded under Known Gaps with its direction table (`ca3e7e8b`).
- The DAG section contradiction, the understated executor requirement, the undocumented recorded interceptor set, and the root that leaves the graph without a detach call (`797533ef`).
- `AttachToContext`'s XML remarks, and both cross-graph rejection messages now naming the recovery (`797533ef`).
- `DetachFromContext`'s XML no longer cites an internal method either; the only remaining occurrence of that name is the implementation call (`797533ef`).
- The generator and runtime version-pairing warning, and the migration guide, both now in the pull request description rather than pending.

RicoSuter:
## Open questions, for a fresh pair of eyes

Rewritten to match the branch as it stands. Things that are unresolved rather than merely unfinished, each with what is known and what is not, so nobody has to re-derive the reasoning. The companion comment lists the work items.

### 1. The co-resolved interceptor deadlock

Two `LifecycleInterceptor` instances resolving from one context means two `_attachedSubjects` monitors with no order between them, and the subtree descent enters one while holding the other. The directions are opposed by construction: attach descends at the first attach, which the innermost interceptor body sees, and detach descends at reference count zero, which only the outermost body reaches. Two concurrent property writes are enough, one attaching and one detaching, with no root operation involved. **No resolution order removes it.**

Verified pre-existing: master has the same lock topology, produced through the old executor override rather than the explicit descent.

Now documented, with the direction table, under Known Gaps in `docs/design/tracking-lifecycle.md`, and the two false sentences in the lock-ordering section are corrected. **Not fixed.**

Deliberately not fixed here, for a reason worth recording: it looks like a small change and is not. Canonical lock ordering has three unresolved problems. The pre-acquired interceptor set can be stale, because a concurrent `AddFallbackContext` anywhere in the chain changes what resolves, and that is reachable from a handler mid-operation. The descent may resolve from a different context than the acquisition did, and there is already a documented shape where those two sets differ. And a deadlock fix cannot be pinned by a test, so it would be permanent locking machinery in the hottest lifecycle code guarded by an argument.

Disallowing the configuration instead was considered and rejected: `WhenContextsAreAggregated_ThenEveryChangeInterceptorPrecedesEveryLifecycleInterceptor` predates this work and says in its own comment that this is *the only configuration where the derived-before-change ordering edge is load-bearing*. Removing the configuration removes the only coverage that edge has.

**Belongs on master as its own piece of work with its own design pass.**

### 2. Is the aggregated two-interceptor configuration actually supported?

The branch treats it as supported: ownership claim and release use graph membership rather than interceptor identity specifically to keep it working, and twelve tests now cover it, five as guarantees. Against that: `TryGetLifecycleInterceptor()` throws in it and three production call sites use it; the claim predicate is asymmetric and the root route and property route claim in opposite directions; it has the deadlock above; and nothing in this repository builds it outside single-threaded tests.

The graph-membership predicate is about four lines of production code but bought a 400-line test file and the most convoluted section of the design document. Worth revisiting deliberately rather than by default.

### 3. Unexplained CI flakiness in `test-homeblaze-integration`

Failed three times in eleven runs on this branch, zero in eight on master. Every failure identical: the application's nav never renders, so all 23 E2E tests time out at 30 seconds. A rerun of the exact failing commit passed with no code change, so it is flaky, not deterministic. If the branch's true rate matched master's, a clean 8 for 8 there has roughly a 7% probability. Suggestive, not conclusive, small sample.

The fixture now polls the server until it answers before the browser launches, so the next occurrence distinguishes "the server never came up" from "the UI never rendered", which was impossible before. **The cause is not established.** If it recurs, the candidate worth testing is whether startup got slower: a constructor-attached subject placed under a parent now keeps its own compiled chains instead of sharing the terminal's, and `CanvasNode`, `GridLayout` and `Motor` are exactly the types that get the generated context constructor.

### 4. Three claims protected by argument rather than by test

- **`ReleaseAttachEdge`'s capture-and-compare.** With the attach-side reference-count guard in place, a record can no longer be written between the decrement and the release at all, which makes the comparison unfalsifiable by any test. It is correct, and it is also unprotected against someone removing it.
- **Three mutants have no killer.** Two are unreachable given the guards above them; one is a missing `Volatile.Read`, a visibility weakening with no deterministic observation.
- **The deadlock in item 1** cannot be pinned without flakiness, which is why the entry says so explicitly.

### 5. A same-context ABA in the attach-edge release

Three threads: a connector item with record R at reference count 1; T1 decrements to zero and captures R; T2 detaches through R, clearing the record, then blocks; T3 re-attaches through R, re-recording the same context object and dedupping against the still-published edge, then blocks; T1's `finally` then finds a record reference-equal to its capture and clears T3's record and edge.

Inside the documented "root attach racing root detach" family, and the pinned recovery property holds, so a fresh attach and detach round un-wedges the subject. **Worth one sentence in that Known Gaps entry; not yet added.**

### 6. Calibration note

Every defect found in this work was found by reading code, by mutation testing, or by a reviewer constructing an interleaving. **Not one was found by a test.** The suite has been a reliable regression net and has never once been the discovery mechanism, including for two tests that were themselves passing under the very mutants they were named for, and for two guards whose deletion left the entire suite green.

Worth knowing when judging how much the green suite is evidence of. The areas where inspection has already found something and testing has not are the aggregated configuration and the concurrent root operations.

PR #420, opened 2026-08-04
Deliver connector changes by commit order, and stop echoes discarding committed local writes

RicoSuter:
Supersedes #355, folded in here so only one final state needs reviewing, and closes the connect-window write loss #355 was opened for.

## Summary

A connector drops an outbound change only when a later commit will carry the settled value in its place. That is decided by **commit order**, not by comparing values: every committed write stamps its revision on the property it wrote, and a change whose revision the property has moved past was superseded by a commit that is still to come or has already gone out.

Deciding by commit order rather than by value also means it works for derived and runtime-registered properties, whose getters need not return what a write stored. A value comparison could not judge those at all and silently delivered everything, which is every property the OPC UA client loader creates.

Which commits count as superseding is not the same question for every connector, and getting it wrong loses data in both directions. That is the second half of this PR, and the part that took the longest to find.

## The qualifier that makes dropping safe

"A later commit will carry the settled value in its place" is exact, and one case fails it: a commit that came **from the source** is skipped as an echo when that source's queue is drained, so nothing would be left to deliver. Source-originated commits therefore never advance the marker.

Without that exclusion, a source echoing back a value we had just written suppresses the *next* write to the same property, and both ends settle on the value the source already had, with the local write lost permanently and no diagnostic. Issue #373 covers why an echo cannot be ranked against local writes at all: its revision is stamped when it is applied here, not when the source produced it.

## The exception: a server is the other way round

The argument above rests on the source having produced its value before it saw our write. That is true of anything we talk to over a wire. It is false for a server, where the applied value is a **client's own write**, already sitting in the store the server serves.

Ranking against the wrong marker there fails the same invariant from the other side. A local commit that predated the client's write and reached the write loop late is not treated as superseded, so it goes out after the client's value and leaves every client on a value the model has moved past. On the OPC UA server there is no way back from that state, because it applies inbound writes under its own source, so the apply is skipped as an echo and nothing carries the client's value out again. The node and the subject disagree permanently.

So the rule is a property of the sink, and specifically **not** the client/server distinction. What decides it is whether an applied value has already reached the destination when we apply it, which is what `ChangeDeliveryRule` asks. The three servers satisfy that condition for two different reasons, and the difference is load-bearing:

| Connector | Inbound applied under | Echo-skipped | Why the rule holds |
|---|---|---|---|
| OPC UA server | its own source | yes | the SDK wrote the node before the change reaches the subject |
| MQTT server | `_mqttClientSource` | no | applied under a foreign source and relayed on to all clients |
[...]

Issue #416, opened 2026-08-04
OPC UA integration tests are flaky under load: two distinct root causes


RicoSuter:
The OPC UA integration tests fail intermittently under machine load. Investigation showed these are **two unrelated defects**, not one. They were initially grouped together under a single certificate error signature, which was wrong and sent the first investigation down the wrong path.

Both are pre-existing on `master`. Measured flake rate on a full `dotnet test src/Namotion.Interceptor.OpcUa.Tests` run:

| branch | runs | failures |
|---|---|---|
| `master` @ e616c769 | 3 | 1 (`NullableGuidType_ShouldSyncBothDirections`) |

Affected tests observed so far, across both cases:

- `OpcUaDataTypesTests.NullableGuidType_ShouldSyncBothDirections`
- `OpcUaTransactionTests.Transaction_DisposedWithoutCommit_ServerShouldNotReceiveChanges`
- `OpcUaMultiClientTests.SequentialWrites_SameProperty_LastWriteWins`

---

# Case 1: transaction conflict from an in-flight subscription echo

**Status: reproduced and root-caused. Ready to fix.**

## Symptom

```
Namotion.Interceptor.Tracking.Transactions.SubjectTransactionConflictException
  : Transaction conflict detected on 1 property(ies): Name
   at SubjectTransaction.ThrowIfConflictsDetected(ReadOnlySpan`1 changes)
        in src/Namotion.Interceptor.Tracking/Transactions/SubjectTransaction.cs:line 670
   at SubjectTransaction.CommitWithWriterAsync(ITransactionWriter writer, CancellationToken cancellationToken)
        in src/Namotion.Interceptor.Tracking/Transactions/SubjectTransaction.cs:line 372
   at OpcUaTransactionTests.Transaction_DisposedWithoutCommit_ServerShouldNotReceiveChanges()
        in src/Namotion.Interceptor.OpcUa.Tests/Integration/OpcUaTransactionTests.cs:line 79
```

Reproduced once in four consecutive full-suite runs on an otherwise idle machine. The captured log shows a clean connection with no reconnect and no certificate error, which is what established that this is a separate defect from Case 2.

## Root cause

Two tests in `OpcUaTransactionTests` write the **same** property, `ServerFixture.ServerRoot.Transactions.SingleProperty.Name`:

- `Transaction_CommitSingleProperty_ServerReceivesChangeOnlyAfterCommit` (line 18) sets it to `"Transaction Value"`
[...]

PR #423, opened 2026-08-05
Capture outbound changes before a source signals readiness

RicoSuter:
Fixes the first of the two defects in #416, and adds the CI path filters that make connector integration suites actually run for this kind of change.

## The defect

`SubjectSourceBase` created its `ChangeQueueProcessor`, and with it the subscription that captures outgoing property changes, only after `StartListeningAsync` and `LoadInitialStateAndResumeAsync` had run. Both of those are what the outside world uses to conclude the source is ready: `StartListeningAsync` claims the properties and creates the OPC UA monitored items, and loading the initial state applies the server's values to the subject, including the `Connected` flag the tests wait on.

A local write landing between either of those and the processor's construction had no subscriber. It was not queued, not retried and not logged, and it never recovered, while the next write to the same property went through in milliseconds.

This is not test only. Any application that writes a property right after host start loses those writes silently.

## Why it correlates with load

The window is one continuation. Under thread pool starvation it widens past the 100 ms readiness poll in the tests, so the test writes into the hole. That is the load correlation recorded in the issue.

## Evidence

Injecting a 3 second stall into that window makes it deterministic. Same harness, same test, both ways:

| | result |
|---|---|
| before | fails every time, server value never changes |
| after | passes in 3.7 s |

Instrumented run at the moment of failure, on master's code:

```
[0.3s] Client connected in 270ms total          <- readiness gate passes
[  7ms] CLIENT ...0000 -> b70ff4d7 origin=Local  <- the write, change IS raised
[3.2s] creating change queue processor           <- capture starts, too late
[20.3s] server subject value : <null>            <- never arrived
        server NODE value    : 00000000-...      <- never reached the node either
        client prop claimed  : True by OpcUaSubjectClientSource
        client connected     : True, reconnecting=False
        client pending writes: 0                 <- not queued, not retried, not logged
[30.3s] dequeued NullableGuidValue ... -> QUEUED  <- a later write goes through instantly
```

## The fix

Create the processor first, then buffer, listen and load. Its property filter is evaluated when a change is dequeued, which is after the claims exist, so capturing earlier is safe. Initial state is applied through `SetValueFromSource`, so those applies carry the source origin and the processor skips them as echoes rather than writing them back (verified: all 55 initial applies show up as `origin=FromSource -> SKIPPED`).

### What it does not cover

The retry backoff window. If `StartListeningAsync` throws, the processor is disposed and rebuilt only after the retry delay. Keeping one processor across attempts would let a source that can never connect accumulate changes without bound, since the queue is unbounded. The normal reconnect path is unaffected either way: it runs inside the health loop with the processor alive, so writes are queued and retried through `WriteRetryQueue`.

## Tests

`WhenPropertyIsWrittenWhileSourceIsStillStarting_ThenChangeIsStillSentToSource` holds the source inside `StartListeningAsync` after it has claimed the property, writes during that window, then releases. Deterministic and protocol free. It fails on master (5 s timeout) and passes with the fix (137 ms).

Three existing `SubjectSourceBaseTests` needed correcting:

- Two stub contexts never provided the `PropertyChangeInterceptor` the subscription is taken from. As a side effect of fixing that, those two stopped burning 30 s each on timeouts and the project went from 1m01s to 0.7 s.
- One simulated "the server changed this value" with a plain local write, which the now live subscription correctly forwards. Real sources apply loaded state through `SetValueFromSource`, so the stubs now do the same, at all eight sites. One other site had the same latent problem and was only passing because it stopped the source before the buffer flushed.

Full unit suite passes, 0 failures.

## CI path filters

The filters matched only each connector's own project, so a pull request touching `Namotion.Interceptor`, `Tracking`, `Registry` or `Connectors` ran no connector integration suite at all. This pull request is an instance: it modifies `Connectors`, which every connector consumes, and without this would not run the OPC UA job that is supposed to verify it. Taken from the same change in #420, so it is reviewed there too. If #420 lands first this hunk becomes a no-op.

## Still open in #416

The second defect is unaffected by this and stays open. Under a loaded harness (2 visible cores, saturated CPU) the first server to client update after a client connects still sometimes never arrives, permanently and with no log:

- `WriteAndReadArraysOnServer_ShouldUpdateClient`
- `ArrayResize_GrowAndShrink_ServerToClient_ShouldSync`

In every observed case the failing test is the one that created its class's shared client. The server side processor is built before the server starts, so the ordering bug fixed here does not explain it. Being investigated separately.

Case 2 in the issue (the certificate mismatch on reconnect) did not reproduce at all in roughly 25 loaded runs and remains unexplained.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -40,6 +40,18 @@ jobs:
         id: filter
         with:
           filters: |
+            shared: &shared
+              - 'src/Namotion.Interceptor/**'
+              - 'src/Namotion.Interceptor.Generator/**'
+              - 'src/Namotion.Interceptor.Tracking/**'
+              - 'src/Namotion.Interceptor.Registry/**'
+              - 'src/Namotion.Interceptor.Connectors/**'
+              - 'src/Namotion.Interceptor.Dynamic/**'
+              # Test-only references: not used by the connectors, only by their test projects.
+              - 'src/Namotion.Interceptor.Testing/**'
+              - 'src/Namotion.Interceptor.Hosting/**'
+              - 'src/Namotion.Interceptor.Validation/**'
+              - 'src/Namotion.Interceptor.AspNetCore/**'
             homeblaze:
               - 'src/HomeBlaze/**'
             nugetplugins:
@@ -50,11 +62,14 @@ jobs:
               - 'src/HomeBlaze/HomeBlaze.Plugins/**'
               - 'src/HomeBlaze/HomeBlaze.Plugins.Tests/**'
             opcua:
+              - *shared
               - 'src/Namotion.Interceptor.OpcUa*/**'
               - 'src/HomeBlaze/*OpcUa*/**'
             websocket:
+              - *shared
               - 'src/Namotion.Interceptor.WebSocket*/**'
             mqtt:
+              - *shared
               - 'src/Namotion.Interceptor.Mqtt*/**'
 
   test:
--- a/src/Namotion.Interceptor.Connectors.Tests/SubjectSourceBaseTests.cs
+++ b/src/Namotion.Interceptor.Connectors.Tests/SubjectSourceBaseTests.cs
@@ -22,6 +22,12 @@ public class SubjectSourceBaseTests
             .Setup(s => s.TryGetService<ISubjectRegistry>())
             .Returns(new SubjectRegistry());
 
+        // The source creates its change-queue subscription before it starts listening, so the
+        // stub context has to provide the change interceptor the subscription is taken from.
+        subjectContextMock
+            .Setup(s => s.TryGetService<PropertyChangeInterceptor>())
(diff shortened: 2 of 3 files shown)


RicoSuter:
Closing in favour of #355, which already contains this change as its first commit and goes further.

This PR implemented only part 1 of that design (create the change-queue subscription before the source signals readiness). It is missing the supersede check, so a write captured during the connect window and then overwritten by the initial-state snapshot would still be sent to the source, pushing a value the model no longer holds. #355 has that, plus the 3-way reconcile, plus capture across the reconnect-delay window that this PR explicitly did not cover, and it keeps memory bounded by draining into the retry queue.

The CI path filter change here is worth keeping: without it a pull request touching Namotion.Interceptor, Tracking, Registry or Connectors runs no connector integration suite at all, so #355 would not be verified by CI either. Carrying it over there. It also exists in #420.

PR #429, opened 2026-08-06
Initialize subjects once instead of once per attach [Design]

### LINKED ISSUE #430: Subject initializers re-run on re-attach and re-add properties that already exist
Anything that adds properties to a subject while it attaches runs again every time that subject re-attaches, but its effect does not re-run with it. The properties it added live on `subject.Properties` and survive detach, so the second run re-adds what is already there and `RegisteredSubject.AddProperty` throws.

An ordinary move between parents is enough to trigger it, whenever the old reference is removed before the new one is added and the reference count passes through zero.

## Scope

Three extension points are affected, not one:

| Extension point | Interface |
|---|---|
| property initializer | `ISubjectPropertyInitializer` |
| property lifecycle handler | `IPropertyLifecycleHandler` |
| subject lifecycle handler | `ILifecycleHandler` |

## Guarding is not a fix

Having each implementation check before adding, which is what the docs currently show, replaces the crash with a silent wrong answer. Measured:

```
initializer invocations:                    2
capture#1 Parent == live registration:      False
capture#1 parent still in KnownSubjects:    False
capture#1 sees marker:                      True
```

An initializer may capture the `RegisteredSubjectProperty` it was handed, and attribute lookups resolve through the `RegisteredSubject` captured with it. The registry builds a new one on every re-attach, so a guarded initializer keeps its first-run attribute alive with a getter that answers from a discarded registration instead of throwing.

## Status

Design agreed and written up in #429, which supersedes the earlier patch-style approach on that branch. Implementation has not started.

Design document: `docs/superpowers/specs/2026-08-07-registration-scoped-initializers-design.md`

The design is deliberately not duplicated here, so the two cannot drift apart. #429 and the spec are the source of truth.

## Superseded analysis

This issue previously argued that re-invocation was load-bearing for a subject moving between contexts, and that keying on the context or on the initializer instance both fail. The design keys on the subject instead, an option not considered at the time, and accepts that a subject moved into a context with different initializers will not receive them, on the grounds that it already carries the previous context's added properties either way. The `TryAddProperty` and `TryAddAttribute` direction previously recommended here was not taken: `AddProperty` keeps throwing, because under once-per-subject semantics a duplicate is a genuine error again rather than an artifact of re-attach.

RicoSuter:
Draft. The design is agreed and committed as a spec; the implementation has not started. The branch still contains the earlier patch-style fix, which this design supersedes and which will come out.

Design document: `docs/superpowers/specs/2026-08-07-registration-scoped-initializers-design.md`

Closes #430.

## Problem

Three extension points add properties to a subject while it attaches:

| Extension point | Interface | Example |
|---|---|---|
| property initializer | `ISubjectPropertyInitializer` | adds a `Unit` attribute |
| property lifecycle handler | `IPropertyLifecycleHandler` | adds `State` and `Configuration` |
| subject lifecycle handler | `ILifecycleHandler` | adds a property per `[Operation]` method |

All three re-run whenever a subject re-attaches, which an ordinary move between parents triggers as soon as the old reference is removed before the new one is added. Their effect does not re-run with them: the properties they added live on `subject.Properties` and survive detach. The second run re-adds what is already there and `AddProperty` throws.

The workaround the docs currently teach, guarding each implementation with a check before adding, makes it worse rather than better. Measured:

```
initializer invocations:                    2
capture#1 Parent == live registration:      False
capture#1 parent still in KnownSubjects:    False
capture#1 sees marker:                      True
```

An initializer may capture the `RegisteredSubjectProperty` it was handed, and attribute lookups resolve through the `RegisteredSubject` captured with it. The registry builds a new one on every re-attach, so a guarded initializer keeps its first-run attribute alive with a getter that silently answers from a discarded registration instead of throwing.

The root cause is that initialization is driven by attach, which is episodic, while its effect is structural and persists on the subject.

## Design

Initializers run once per subject, driven by registration and property creation rather than by attach.

Two flags:

| Flag | Lives on | Set | Read |
|---|---|---|---|
| subject | `subject.Data` | read-and-set once at first registration | decides whether this registration initializes |
| registration | `RegisteredSubject` | from that read | consulted per property in `AttachProperty` |

Both are needed. The subject flag alone flips to "done" while the first pass is still running and would skip every property after the first.

Three events drive initialization:

| Event | Initializes |
|---|---|
| a subject is registered for the first time | all of its properties |
| `AddProperty` creates a property | that property only |
| a subject re-attaches | nothing |

`RegisteredSubject` ownership does not change. It stays owned by its registry and is rebuilt per attach exactly as today.

### New API

```csharp
public interface ISubjectInitializer
{
    void InitializeSubject(RegisteredSubject subject);
}
```

Called once per subject. This is the missing grain size: `ISubjectPropertyInitializer` is per property, and the only per-subject hook is `ILifecycleHandler`, which is episodic by design. Work that is genuinely per-subject-once had nowhere correct to live.

The only new public type. No existing signature changes, so no consumer implementation must be updated.

### Concurrency

`AddProperty` locks on `subject.SyncRoot`, and the separate add lock is removed. `SyncRoot` already guards the subject's own copy of the property set in both the generated and dynamic implementations, so a separate lock leaves that copy free to be mutated concurrently. The check and the add must be atomic across both views.

The lock graph collapses to `SyncRoot -> RegisteredSubject._lock`, acyclic, with `Monitor` re-entrancy resolving the cycle a getter could otherwise create.

`AddProperty` also rejects being called from a getter or setter of the same subject, via `Monitor.IsEntered` before the lock is taken. That check is thread-local rather than a race, and take-or-throw cannot substitute for it because `Monitor` is re-entrant and would succeed in exactly the case being rejected.

## Consumer impact

No public API breaks. `ISubjectPropertyInitializer` keeps its signature. Existing guards in consumer initializers become redundant and stay harmless.

Two HomeBlaze classes change interface, one line each plus their registrations, and both get shorter:

| Class | From | To |
|---|---|---|
| `MethodPropertyInitializer` | `ILifecycleHandler` | `ISubjectInitializer` |
| `PropertyAttributeInitializer` | `IPropertyLifecycleHandler` | `ISubjectPropertyInitializer` |

## Measurements

| Question | Result |
|---|---|
| `Monitor.IsEntered` cost | 1.56 ns, about 0.03% of one `ToFrozenDictionary` already on the same path |
| Should `Parent` resolve the live registration? | No. 19.5x on attribute lookup (2.47 ns to 48.02 ns), 27 call sites, and `Children` stays stale anyway |
| Is a two-registry configuration usable today? | No. Throws when initializers exist, double-counts parent and child edges when they do not |
| Does detach leave graph state behind? | No. Parents, children and reference count are all empty before the registry drops a subject |

## Rejected alternatives

**Retaining `RegisteredSubject` per subject.** Would make initialization once-per-registration by construction and would fix stale captures as a side effect. Rejected because `RegisteredSubject` holds both the property set, which belongs to the subject, and parent and child edges, which belong to the graph. Sharing the object shares both, and two co-resolved registries would then apply every edge twice by construction. Sound only under a guarantee that a subject belongs to at most one graph, which #419 introduces and `master` does not yet make.

**Richer lifecycle context.** Passing a first-attach flag so handlers can defend themselves. Rejected: it preserves the shape of the problem after the problem has been deleted, and puts the burden back on every implementer to check correctly.

## Follow-ups, to be filed separately

- `AddParent` and `AddChild` are not idempotent. Aggregating a second lifecycle-bearing context makes a registry record the same parent edge twice, and detach clears only one of the pair. A pre-existing bug on `master`, the same root cause as the initializer double-add, independent of this change.
- Stale captures. This design stops initializers re-running but still rebuilds `RegisteredSubject` per attach. The correct fix is on the capturing side, capturing `property.Reference` rather than the property, which is free and fully correct. Needs documenting on `ISubjectPropertyInitializer`.
- Benchmark coverage. No benchmark registers an initializer and none re-attaches the same subjects, so every path this design changes is currently unmeasured.

## Relationship to #419

Independent, and this targets `master`. #419 strengthens the design by guaranteeing that a subject belongs to at most one graph, which is what would make retaining the registration safe. Nothing here depends on it. When it lands, the subject flag should move from `subject.Data` onto `InterceptorExecutor`, where that PR already moves other per-subject state.

### DIFF
--- a/docs/registry.md
+++ b/docs/registry.md
@@ -167,6 +167,13 @@ public class UnitAttribute : Attribute, ISubjectPropertyInitializer
 
     public void InitializeProperty(RegisteredSubjectProperty property)
     {
+        // Initializers run again whenever a subject is re-attached, over properties that are
+        // still on the subject, so check before adding.
+        if (property.TryGetAttribute("Unit") is not null)
+        {
+            return;
+        }
+
         property.AddAttribute("Unit", typeof(string), _ => _unit, null);
     }
 }
@@ -187,7 +194,7 @@ public class DefaultValueInitializer : ISubjectPropertyInitializer
             .OfType<DefaultValueAttribute>()
             .FirstOrDefault();
 
-        if (attribute is not null)
+        if (attribute is not null && property.TryGetAttribute("DefaultValue") is null)
         {
             property.AddAttribute("DefaultValue", property.Type,
                 _ => attribute.Value, null);
--- a/docs/superpowers/specs/2026-08-07-registration-scoped-initializers-design.md
+++ b/docs/superpowers/specs/2026-08-07-registration-scoped-initializers-design.md
@@ -1,12 +1,13 @@
-# Registration-scoped subject initializers
+# Initialize subjects once, not once per attach
 
 Status: approved, not yet implemented
 Issue: #430
 Supersedes the approach in PR #429
+Targets `master`. PR #419 strengthens this design but is not a prerequisite.
 
 ## Problem
 
-Three separate extension points add properties to a subject while it attaches:
+Three extension points add properties to a subject while it attaches:
 
 | Extension point | Interface | Example |
 |---|---|---|
@@ -14,27 +15,26 @@ Three separate extension points add properties to a subject while it attaches:
 | property lifecycle handler | `IPropertyLifecycleHandler` | `PropertyAttributeInitializer` adds `State` and `Configuration` |
 | subject lifecycle handler | `ILifecycleHandler` | `MethodPropertyInitializer` adds a property per `[Operation]` method |
 
-All three re-run whenever a subject re-attaches, which an ordinary move between parents
-triggers as soon as the old reference is removed before the new one is added. Their effect
-does not re-run with them: the properties they added live on `subject.Properties` and
-survive detach. The second run therefore adds a property that is already there, and
-`RegisteredSubject.AddProperty` throws.
+All three re-run whenever a subject re-attaches, which an ordinary move between parents triggers
+as soon as the old reference is removed before the new one is added. Their effect does not re-run
+with them: the properties they added live on `subject.Properties` and survive detach. The second
+run therefore adds a property that is already there, and `RegisteredSubject.AddProperty` throws.
 
 ### The re-run is redundant, not merely unsafe
 
 `RegisteredSubject`'s constructor rebuilds its property set from `subject.Properties`
-(`RegisteredSubject.cs:143`), which still holds everything the previous run added. On
-re-attach the entire prior effect is already present before any initializer runs.
+(`RegisteredSubject.cs:143`), which still holds everything the previous run added. On re-attach
+the entire prior effect is already present before any initializer runs.
 
 `LifecycleInterceptor.AttachToContext` snapshots `subject.Properties.Keys` before invoking
-handlers, and on re-attach that snapshot already contains the dynamic properties. So the
-second run does not repeat the first, it runs every initializer over a strictly larger
-property set, including the properties the initializers themselves created.
+handlers, and on re-attach that snapshot already contains the dynamic properties. So the second
+run does not repeat the first. It runs every initializer over a strictly larger property set,
+including the properties the initializers themselves created.
 
-### Guarding makes it worse
+### The documented workaround makes it worse
 
-The documented workaround is for each implementation to check before adding. Measured, that
-converts a loud failure into a silent one:
+Guarding each implementation with a check before adding is what the docs currently teach.
+Measured, it converts a loud failure into a silent one:
 
 ```
 initializer invocations:                    2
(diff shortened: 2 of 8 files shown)


RicoSuter:
Design revision from working through the neighbouring code in #494. The spec's single subject flag does not survive a subject moving between contexts, and there is a cleaner semantic that does.

## The semantic

Attach and detach are episodic and repeat. Initialization is structural and should happen **once per handler that cares**, not once per subject.

Record `(subject, handler)` rather than `(subject)`:

```
context A = {H1, H2}, subject attaches
  H1 initializes  -> record (subject, H1)
  H2 initializes  -> record (subject, H2)

detach, attach to context B = {H2, H3}
  H2 already recorded -> skip
  H3 not recorded     -> run
```

The cross-context subset falls out with no special case. Any model that needs an explicit "did the context change" rule is the wrong shape.

## Why per-context recording is not enough

There is no guard on `InterceptorSubjectContext.AddService` or `TryAddService`, so a handler can be registered after subjects have already attached. Recording "initialized in context X" would skip that handler forever. Per-handler records are required for correctness, not just tidier.

## Granularity, so this does not cost N x H

Handlers do one pass over all properties, so the record belongs at `(subject, handler)`, giving H entries per subject rather than one per property per handler. The only per-property case left is a property created later, which gets initialized at creation time by every handler already recorded for that subject.

The spec's existing note still applies and is the reason two flags are needed: the subject-level record cannot be written before the pass, or the first property flips it and the rest are skipped.

## The open question is handler identity

This is the part worth settling before any code:

- **Instance reference** is semantically right but retains a discarded context's handler if stored in `subject.Data`. A `ConditionalWeakTable` handing out a stable id per instance fixes that: the subject stores an int and nothing retains the handler.
- **Type** is cheap and non-retaining but conflates two instances of the same initializer type.

It also forces a judgement the spec currently avoids: if a context is torn down and rebuilt with fresh instances of the same initializers, should they re-run? Per-instance says yes. I think yes, which argues for instance identity.

## What this does not fix, so #430 should not close on it

Capture staleness is orthogonal. `SubjectRegistry` rebuilds `RegisteredSubject` on every attach regardless of whether initializers run, so anything holding a `RegisteredSubjectProperty` across detach is stale either way. Running exactly once actually removes the initializer's chance to refresh its own capture. "Do not capture projections" is needed guidance under any of these models.

## Relationship to #494

#494 makes admission reject a name the subject already carries, and adds an idempotent guard in `RegisteredSubject.AddProperty` so a re-running initializer does not throw. That is deliberately a symptom fix: it makes the re-run harmless rather than preventing it. All three extension points listed above funnel through `RegisteredSubject.AddProperty`, so they are all covered by it.

That guard is what this issue would let us delete, which is the main reason to build this on top of #494 rather than on master.

RicoSuter:
Scope decision: this stacks on #494 rather than on master, and it also **deletes** the idempotency guard #494 adds to `RegisteredSubject.AddProperty`.

That makes the end state clean, and it has two consequences worth writing down now rather than rediscovering later.

## The duplicate semantics change twice

| | duplicate `AddProperty` |
|---|---|
| master | last-wins, silently |
| #494 | first-wins, silently |
| this issue | throws |

Two migrations across two releases for the same call. Deliberate, and worth a single combined note in the release notes rather than two separate ones.

## Removing the guard removes a safety net that we know catches something real

Measured on master: an initializer adding an attribute property, then an ordinary move between parents, throws `ArgumentException: An item with the same key has already been added. Key: LastName@Unit` from the projection insert in `AddPropertyInternal`. That is exactly the crash the guard absorbs.

So deleting the guard is only safe once the once-per-handler skip provably covers **every** path that currently re-registers a name the subject already carries. If one is missed, the crash comes back in a new guise rather than being caught. Before the guard comes out, that list needs to be enumerated rather than assumed:

- context-registered `ISubjectPropertyInitializer` reruns
- attribute-sourced initializers from `property.ReflectionAttributes`
- `ILifecycleHandler` implementations that add properties, such as HomeBlaze's `MethodPropertyInitializer`
- `IPropertyLifecycleHandler` implementations that do the same
- a consumer calling `AddProperty` twice deliberately, which is the one case the skip cannot cover, and which becomes a throw by design

The last bullet is the intended new behaviour. The first four have to be shown to no longer re-register, not merely expected to.

## Suggested order

1. Add the once-per-handler skip, keeping the guard. Verify the reruns stop, including the reattach test `WhenASubjectCarryingAnInitializerAddedAttributeReattaches_ThenTheRerunIsAbsorbed`, which should then show the initializer not rerunning at all.
2. Only then delete the guard, and confirm the suite is still green with the throw as the new duplicate behaviour.

Splitting it that way means step 2 failing tells you a path was missed, rather than the two changes masking each other.

RicoSuter:
## Implementation notes

Parked for later. Everything below is against #494's head, which is the intended base.

### Where it goes

| What | Anchor |
|---|---|
| the two initializer loops to gate | `SubjectRegistry.cs:251` and the context-service loop directly below it |
| the guard to delete in step 2 | `RegisteredSubject.cs:331` |
| record storage | `subject.Data`, typed `ConcurrentDictionary<(string? property, string key), object?>` so a `(null, "initialized-handlers")` entry holds the set |
| test that must flip | `InitializerReattachTests.WhenASubjectCarryingAnInitializerAddedAttributeReattaches_ThenTheRerunIsAbsorbed` |

That last one is the cheapest progress check: it currently asserts the rerun happens and is absorbed. When the skip works it should assert the initializer does **not** rerun, and its name should change with it. If it still passes unmodified, the skip is not wired in.

### The record

`(subject, handler)` granularity, not `(subject, property, handler)`. Handlers do one pass over all properties, so H entries per subject rather than N x H. Write the record **after** the pass, not before, or the first property flips it and the rest are skipped: that failure mode is already noted in the original spec and is the reason two flags exist there.

A property created after a handler's pass is initialized at creation time by every handler already recorded for that subject. That is the only per-property case.

### Handler identity, the one open decision

Recommended: `ConditionalWeakTable<object, object>` handing out a stable boxed id per handler instance, so `subject.Data` stores the id and nothing retains a discarded context's handler.

The alternative is keying on handler **type**: cheaper, but it conflates two instances of the same initializer type, and a context torn down and rebuilt with fresh instances would silently skip them. I think a rebuilt context should re-run, which is what argues for instance identity, but this is a semantic call rather than a mechanical one and it is worth deciding deliberately because it is awkward to reverse after shipping.

### Order, and why it matters

1. Add the skip, **keeping** the guard. Verify reruns stop.
2. Delete the guard as a **separate** commit. Confirm green with throwing as the new duplicate behaviour.

If both land together and a re-registration path was missed, the guard hides it and the crash returns later looking like something unrelated. Split, step 2 going red tells you exactly that.

### Paths that must be shown to stop re-registering before step 2

- context-registered `ISubjectPropertyInitializer`
- attribute-sourced initializers from `property.ReflectionAttributes`
- `ILifecycleHandler` implementations adding properties, HomeBlaze's `MethodPropertyInitializer` being the live one
- `IPropertyLifecycleHandler` implementations doing the same
- a consumer calling `AddProperty` twice deliberately: the one case the skip cannot cover, and the one that becomes a throw by design

### Tests to add

- cross-context move where the two contexts have overlapping but different handler sets, asserting only the subset runs
- a handler registered on the context **after** subjects already attached, asserting it still runs (there is no guard on `AddService`, which is why per-context recording is insufficient)
- a property added after a handler's pass, asserting recorded handlers initialize it at creation
- a context rebuilt with fresh handler instances, asserting whatever the identity decision above settles on

### Out of scope

Capture staleness. `SubjectRegistry` rebuilds `RegisteredSubject` per attach regardless, so a captured `RegisteredSubjectProperty` is stale across detach either way, and running once actually removes the initializer's chance to refresh its own. #430 keeps that part and should not close on this.

Issue #403, opened 2026-07-31
Concurrent TryAddService across delegation levels can admit a duplicate


RicoSuter:
`TryAddService` keeps its exists-check and its add atomic against other mutators of the same context, because they all serialize on that context's mutation lock. It does not serialize against a mutation of a different context in the same fallback chain.

Two threads calling `TryAddService` for the same service type, one on a delegating context and one on the context it delegates to, can therefore both observe the service as absent and both add it. A later `TryGetService` for that type then throws:

```
There must be exactly one service of type ...
```

Pre-existing and unchanged by #400. The old design had the same window: its per-hop `_lock` was released between hops during the downward walk, so it did not serialize the two adds either.

## Practical impact

Low. Every `TryAddService` call site in the repository is a `WithXxx()` configuration extension that runs during setup, effectively single-threaded. The one lazy runtime registration is HomeBlaze's `SubjectPathResolver`, and it registers into a single context.

## Options

- Document the contract as "atomic per context, not per chain", which matches what the code has always done.
- Serialize chain-wide registration behind a lock owned by the root of the chain, which reintroduces cross-context lock ordering and would need care to stay deadlock free.

The first is probably right unless a concrete consumer needs the stronger guarantee.

Issue #380, opened 2026-07-18
Bug: ServiceOrderResolver binds RunsBefore/RunsAfter to only the last instance per type under aggregation


RicoSuter:
`ServiceOrderResolver.TopologicalSortInto` builds a `Dictionary<Type, int>` that maps each service type to a single index, the last-registered instance of that type:

```csharp
var typeToIndex = new Dictionary<Type, int>(count);
for (var i = 0; i < count; i++)
    typeToIndex[services[i].GetType()] = i;   // overwrites, keeps only the last index per type
```

Both edge loops then look up a single index:

- `RunsBefore(T)` on instance `i` resolves `T` to one target
- `RunsAfter(T)` on instance `i` resolves `T` to one source

Consequence: when the referenced type `T` has multiple aggregated instances (which happens when a subject's context aggregates several fallback contexts that each register the same interceptor type), an ordering edge to or from `T` binds to only one of them (the last). The other instances are left unordered relative to that edge, so a `RunsBefore` / `RunsAfter` constraint can be silently violated for all but the last instance.

Example: with two `SubjectTransactionInterceptor` instances, `[RunsAfter(SubjectTransactionInterceptor)]` on another interceptor orders it after only the last transaction interceptor, not both.

## Fix

Map each type to a list of indices and add the edge to every index:

```csharp
var typeToIndex = new Dictionary<Type, List<int>>(count);
for (var i = 0; i < count; i++)
{
    var type = services[i].GetType();
    if (!typeToIndex.TryGetValue(type, out var list))
        typeToIndex[type] = list = new List<int>();
    list.Add(i);
}
// ...
foreach (var beforeType in info.RunsBefore)
    if (typeToIndex.TryGetValue(beforeType, out var targets))
        foreach (var target in targets) { (adjacency[i] ??= []).Add(target); inDegree[target]++; }

foreach (var afterType in info.RunsAfter)
    if (typeToIndex.TryGetValue(afterType, out var sources))
        foreach (var source in sources) { (adjacency[source] ??= []).Add(i); inDegree[i]++; }
```

## Scope and testing

This changes ordering behavior for every interceptor type (validation, lifecycle, context inheritance, derived-change detection, property-change delivery), so it needs:

- New tests for multi-instance `RunsBefore` / `RunsAfter` under aggregation (an edge to a duplicated type must order against all instances).
- A regression pass over the existing ordering tests.
[...]

Issue #343, opened 2026-06-10
Transaction commit local apply is re-pushed to sources by the change queue (double write, revert race)


RicoSuter:
## Summary

A source-bound transaction commit writes each change to its external source twice: once through the transaction writer during commit (stage 1), and again through the background change queue a few milliseconds later, triggered by the stage-2 local apply. In failure flows the queue echo can also race the commit's own revert writes and leave the source at the wrong value. The fix is to mark stage-2 applies of source-bound changes with their owning source so the existing echo suppression treats them like inbound data: the transaction writer becomes the only thing that writes to sources during a commit.

Day-one behavior: present since source transactions were introduced in #114 and faithfully preserved by #337. Uncovered while analyzing #340.

## Current behavior

`ChangeQueueProcessor` has no transaction awareness. Its only outbound suppression is the echo filter (`change.Source == _source`) plus the ownership filter. The stage-2 apply (`SubjectPropertyChangeOperations.ApplyLocalChanges`) replays each change through the interceptor chain with the change's captured source, which is null for user writes inside a transaction. Every notification fired by the apply pass therefore looks like an ordinary local write and is enqueued for a background write to the owning source.

Per flow:

| Flow | Notifications fired by commit | Queue behavior |
|---|---|---|
| Successful commit | new values | Double write: stage 1 already wrote to the source, the queue writes the same values again about one buffer window (8 ms) later |
| Rollback, source write fails | none (nothing applied) | Nothing; divergence from a failed revert persists as documented |
| Rollback, local apply fails | new (apply), then old (local revert) | Usually merged by flush dedup into one old-value write: an accidental, unspecified second revert attempt backed by the `WriteRetryQueue`. It can quietly repair a failed source revert |
| Same, flush boundary between apply and revert | new in one flush, old in the next | The queue pushes new, then old, racing the transaction's revert writes. Depending on which write the source processes last, the source can end at new even though the revert succeeded: the echo can create divergence |
| BestEffort, local apply fails | new for kept properties | Duplicate new-value push (idempotent); failed-apply properties fired no notification, their reverts stand |

## Design

Re-mark written changes with their owning source, per change:

[...]

PR #438, opened 2026-08-07
generator: Intercept base-declared properties in subject hierarchies

### LINKED ISSUE #437: Generator: base-declared properties are never intercepted in a subject inheritance hierarchy
In a `[InterceptorSubject]` inheritance hierarchy, properties declared on the **base** class are never intercepted. The value is written and `PropertyChanged` still fires, so the failure is silent: change tracking never records the write, and connectors such as OPC UA and MQTT never see it.

Found while adding behaviour tests in #432. Reproduced identically at that PR's merge base, so it is pre-existing and neither caused nor fixed there.

## Repro

```csharp
[InterceptorSubject]
public partial class ProbeBaseSubject
{
    public partial string BaseProperty { get; set; }
    public ProbeBaseSubject() { BaseProperty = ""; }
}

[InterceptorSubject]
public partial class ProbeDerivedSubject : ProbeBaseSubject
{
    public partial string DerivedProperty { get; set; }
    public ProbeDerivedSubject() { DerivedProperty = ""; }
}

var context = InterceptorSubjectContext
    .Create()
    .WithFullPropertyTracking()
    .WithService(() => recorder);   // an IWriteInterceptor that records Property.Name

var subject = new ProbeDerivedSubject(context);
subject.DerivedProperty = "d";
subject.BaseProperty = "b";
```

Observed, at both the merge base and current head:

```
observed writes: [DerivedProperty]
DerivedProperty observed: True
BaseProperty observed:    False
values -> derived=d base=b
```

## Mechanism

`SubjectCodeGenerator.Generate` calls `EmitInterceptorSubjectImplementation(builder)` unconditionally, so **every** `[InterceptorSubject]` class in a hierarchy emits its own `private IInterceptorExecutor? _context` field and its own explicit `IInterceptorSubject.Context` implementation.

The most derived explicit interface implementation wins the interface map, so the generated constructor's `((IInterceptorSubject)this).Context.AddFallbackContext(context)` only ever populates the **most derived** class's `_context`. The base class's own `GetPropertyValue` and `SetPropertyValue` read the base's field, which stays null forever, and take the documented no-interception fast path:

```csharp
return _context is not null ? _context.GetPropertyValue(propertyName, readValue)! : readValue(this)!;
```

`PropertyChanged` still fires because the fast path returns `true` and the setter calls `RaisePropertyChanged` directly, which is what makes this look like it works.

The asymmetry is visible in one place. On the adjacent line, the `INotifyPropertyChanged` plumbing IS gated:

```csharp
EmitNotifyPropertyChangedImplementation(builder, metadata.BaseClassHasInpc);   // gated
EmitInterceptorSubjectImplementation(builder);                                 // not gated
```

`SubjectMetadata` already carries `BaseClassHasInterceptorSubject`, which is the flag this needs, and it is already used to decide how `RaisePropertyChanged` is called and whether `DefaultProperties` gets the `new` modifier.

## Scope

Affects any subject hierarchy where a base class is itself `[InterceptorSubject]` and declares partial properties. `docs/subject-guidelines.md` documents that pattern under "Virtual and Override" and `docs/generator.md` under "Inheritance", so it is a supported and advertised shape.

The repository's own `Employee : PersonBase` test pair exhibits it.

## Note on fixing: do not start with the one line gate

Gating `EmitInterceptorSubjectImplementation` on `BaseClassHasInterceptorSubject`, mirroring the line above it, is the obvious move and it is wrong on its own. The same emitted block also contains:

```csharp
IReadOnlyDictionary<string, SubjectPropertyMetadata> IInterceptorSubject.Properties
    => _properties ?? DefaultProperties;
```

`DefaultProperties` there is **statically bound**. Emitted inside the derived class it resolves to that class's `public new static DefaultProperties`, which already concatenates the base's, so `Properties` is correct today *precisely because* the block is duplicated. Emit it only in the base and every derived subject silently reports the base's property set instead of its own, trading a silent interception bug for a silent metadata bug.

So the block has to be split rather than gated:

- `_context` and the three helpers (`GetPropertyValue`, `SetPropertyValue`, `InvokeMethod`) become shared, which means `protected` rather than `private`, since the derived class's generated properties call them.
- `Properties`, `AddProperties` and `_properties` either stay per class, or move behind a virtual member so the most derived `DefaultProperties` still wins.
- `SyncRoot` and `Data` need the same call made. Each is currently re-implemented and shadowed the same way, and sharing them is probably correct, but it is a separate decision from the two above.

Worth a behaviour test pinning that a base-declared property is intercepted, since nothing currently covers it. Note that `PropertyChanged` and the property value both work today, so a test asserting either of those will pass while the bug is present; the assertion has to be that an `IWriteInterceptor` observed the write.

---

## Measured scope of the duplication

Three level hierarchy (`ScopeBase` -> `ScopeMiddle` -> `ScopeLeaf`), inspected by reflection:

| Plumbing member | Copies across 3 levels | Status |
|---|---|---|
| `_context` | 3 | **Broken.** Only the most derived is ever populated, so base declared properties are never intercepted |
| `_properties` | 3 | Dead. Only the most derived is read |
| `Data` backing field | 3 | **Dead allocation.** One `ConcurrentDictionary` per level |
| `SyncRoot` backing field | 3 | **Dead allocation.** One `object` per level |

`IInterceptorSubject.Properties` correctly reports all three levels, and `Data` and `SyncRoot` are stable when reached through the interface, because the most derived copy consistently wins. So the base copies are not wrong, they are unreachable.

That makes this an allocation fix as well as a correctness fix: every extra inheritance level costs a wasted `ConcurrentDictionary` plus a wasted `object` **per subject instance**. `AGENTS.md` ranks allocations first among performance concerns, and HomeBlaze instantiates many device subjects.

## Proposed split

Divide the emitted block by whether a member is per hierarchy state or per class metadata.

**Emit only in the root subject** (the one whose base is not itself an interceptor subject):

- `_context` and `IInterceptorSubject.Context`, which is the actual fix
- `Data` and `SyncRoot`, which removes the dead allocations
- `_properties` and `AddProperties`
- the three helpers `GetPropertyValue`, `SetPropertyValue` and `InvokeMethod`

The helpers and `_properties` become `protected` rather than `private`, because the derived class's generated properties call them. This matches `RaisePropertyChanged`, which the generator already emits as `protected`.

**Emit in every class:**

- `IInterceptorSubject.Properties => _properties ?? DefaultProperties`, the one line that must statically bind to its own `DefaultProperties`

**`AddProperties` needs one rewrite to be shareable.** It currently reads:

```csharp
_properties = (_properties ?? DefaultProperties).Concat(...).ToFrozenDictionary();
```

which carries the same static binding trap. Changing the left operand to `((IInterceptorSubject)this).Properties` makes it dispatch through the per class property, so it always sees the most derived `DefaultProperties`. Semantically identical, and it lets `AddProperties` live in the root.

## Open questions for the design pass

1. Are `protected` generated helpers acceptable as visible surface to a subclass author? They become part of the shape every subject exposes. The precedent is `RaisePropertyChanged`, which is already `protected`.
2. What happens when the base class implements `IInterceptorSubject` by hand rather than through the attribute? `BaseClassHasInterceptorSubject` is false in that case, so the derived class would keep emitting its own plumbing and shadow the hand written one, which is today's behaviour. Confirm that stays correct.
3. Does anything outside the generator read these members by reflection, expecting one per class?

## Verification the fix must pass

- A base declared partial property is observed by an `IWriteInterceptor`. Note that both the property value and `PropertyChanged` work today, so a test asserting either passes while the bug is present. The assertion has to be interceptor observation.
- The three level case, not just two.
- `IInterceptorSubject.Properties` still reports every level's properties.
- Generated output for a subject with no interceptor subject base is unchanged, since that is the overwhelming majority.

RicoSuter:
Fixes #437.

Properties declared on a base `[InterceptorSubject]` class are never intercepted. The value is written and `PropertyChanged` still fires, so the failure is silent: change tracking never records the write, and connectors such as OPC UA and MQTT never see it.

Every class in a hierarchy emits its own `private IInterceptorExecutor? _context` and its own explicit `IInterceptorSubject.Context`. The most derived explicit implementation wins the interface map, so the constructor's `AddFallbackContext` only ever populates the most derived field. Every base class keeps a permanently null `_context` and takes the no-interception fast path.

## Approach

The generator gains two emission modes:

- **Root mode**, for a subject with no subject ancestor: the full interception members as today, with the three accessor helpers `protected` instead of `private` and a new `protected GetInstanceProperties()`.
- **Derived mode**, for a subject whose ancestor provides those members: one line, its own explicit `IInterceptorSubject.Properties`, inheriting everything else.

`Properties` is the one member that cannot move. `DefaultProperties` is a `static` hidden by `new` at each level, so the expression binds at compile time to the class it was emitted into; emitted only in the root, every derived subject would report the root's property set.

Mode selection and all base class facts are resolved from the **nearest subject ancestor** rather than the immediate base, which also repairs two shapes that do not build today.

## Also fixed

Both are pre-existing, both reproduce identically at the commit before #432, and neither had any test:

- A plain class between two subjects does not build (three CS0108, which are build errors here), because the generator reads the immediate base and cannot see its own output.
- A sealed root subject does not build (CS0628 on `protected RaisePropertyChanged`).

## Allocations

Per instance interception members are allocated once per subject instead of once per inheritance level. A three level hierarchy drops from 3 `ConcurrentDictionary` plus 3 `object` plus 12 reference fields to 1, 1 and 4. This is a side effect rather than the motivation: subject over subject hierarchies are rare in this repository.

## Diagnostics

| Rule | Severity | Fires when |
|---|---|---|
| NI0011 | Error | The base implements `IInterceptorSubject` but does not satisfy the subject base contract. Replaces today's CS0117 inside generated code |
| NI0012 | Warning | The base cannot share its interception members, for example an older generator build. Falls back to today's behaviour so it still compiles |
| NI0013 | Error | A member captures an inherited generated member by name |
| NI0014 | Error | A member takes an `IInterceptorSubject` slot from the base under interface re-implementation, whether declared public or as an explicit implementation, anywhere on the ancestor chain |
| NI0006 | Warning | Extended: a `*WithoutInterceptor` method whose stripped name is one the generated half occupies. The wrapper is not emitted |

There are five breaking changes, all loud at compile time and all listed with their rationale in
`docs/design/generator-supported-shapes.md`. See "What changes for you" below.

## Follow-up

The hand-written base class contract this PR documents takes about 70 lines to satisfy, plus three
behavioural rules the compiler cannot check. #443 tracks shipping a base class that carries all of it,
which needs no generator change and would reduce the common case to deriving from one type. Kept out
of this PR because it is new public API in a released package.

## Change size

Measured against `origin/master`.

| Category | + | − | net | files |
|---|---|---|---|---|
| Generator (production) | 1,178 | 149 | **+1,029** | 12 |
| Tests | 3,419 | 83 | +3,336 | 11 |
| Documentation | 513 | 171 | +342 | 3 |
| Benchmarks | 95 | 0 | +95 | 1 |
| Test snapshots (generated) | 132 | 272 | **−140** | 16 |
| **Total** | **5,337** | **675** | **+4,662** | **43** |

Generated output got 140 lines shorter overall, because derived subjects stopped emitting a block they
now inherit.

The construction allocation win is structural rather than measured: `SubjectHierarchyBenchmark` is new
on this branch, so the base arm has no counterpart row and the benchmark tables below cannot show it as
a delta. What is certain is the shape change described under "Allocations" above, three
`ConcurrentDictionary` plus three `object` plus twelve reference fields becoming one, one and four for a
three level subject. `ConstructThreeLevel` measures 3,383 ns and 1,480 B on the branch.

Comment density in the four base class files is 29%, after a compaction pass took it down from 35%.
What remains records language rules established by compiling the shape rather than by reading the
specification, several of which were found by a review pass breaking the code that contradicted them.
Where the design document owns the full argument, the code states the rule and points at it.

A simplification sweep is included, gated on byte-identical generated output. It left the generator
roughly neutral on size rather than smaller. What it bought is structural: base class reasoning now
lives behind one call instead of eight reaches across a file boundary, and the generated member names
have a single source of truth, which is where the contract check and the hiding check had already
drifted apart once.

The base class reasoning is then split across four files, none over 300 lines: mode selection and the
contract check, the ancestor walk, the table of what the generator emits, and the scans for members
that collide with it. Also a pure move, gated the same way.

## Benchmarks

**Verdict: no regression.** The interception hot path is identical at the machine code level, and the
full suite median over 49 shared rows is +0.48% against a measured noise floor of about 3%.

### What was compared

Both arms run back to back through `scripts/benchmark.ps1` on an Intel i7-11700K pinned to its 3.6 GHz
base clock, turbo disabled, `performance` governor, so the frequency is fixed for the whole run and the
machine is otherwise idle. Ubuntu 24.04, .NET SDK 10.0.100, host .NET 9.0.11. Full BenchmarkDotNet job,
not `-Short`, with `--launchCount 3` and `--memoryRandomization`.

The same pair of commits was measured twice:

| Run | Filter | Base arm | Branch arm |
|---|---|---|---|
| 1 | `*RegistryBenchmark*` | `1280eeb4` | `f290d4a8` |
| 2 | `*` (whole suite) | `1280eeb4` | `f290d4a8` |

`f290d4a8` is this branch merged with `1280eeb4`, so the two arms differ by this branch's commits only,
and run 1 and run 2 are the same comparison repeated on byte identical inputs. That makes the spread
between the two runs a direct measurement of this machine's noise, which is the point of running it
twice.

Both runs measured branch commit `b0af7f2d` against base `1280eeb4`. Both refs have moved since: master
gained #420, and this branch gained nine commits. Those nine are documentation, tests, benchmark
cleanup, and identifier renames inside the generator's own model types, none of which change a single
character of emitted code, so the numbers below still describe what the generator produces. Re-measure
if a later commit changes the emitted output.

### The hot path is unchanged at the machine code level

Timing cannot settle a "no regression at all" question at the few percent level on this hardware, so the
primary evidence is static.

Only the generator project changes. Every runtime assembly on the property read and write path,
`Namotion.Interceptor`, `Namotion.Interceptor.Tracking` and `Namotion.Interceptor.Registry`, is bit
identical between the arms. `git diff --name-only 1280eeb4..f290d4a8 -- src/ ':!*Tests*' ':!*Benchmark*'
':!*Generator*'` returns nothing.

That leaves the generated code inside subject classes. For a root mode subject, which covers 256 of the
270 regenerated files and both subjects the registry benchmarks use, exactly four things change and none
of them is inside a property accessor:

| Change | Runtime effect |
|---|---|
| Three helpers move from `private` to `protected` | None. Both compile to non virtual instance methods, so only the accessibility flag differs |
| `Properties` reads `GetInstanceProperties()` instead of the `_properties` field | Inlined away, see below |
| `AddProperties` reads `((IInterceptorSubject)this).Properties` instead of the field | One interface call, inside a lock, ahead of a `Concat` and a `ToFrozenDictionary`. Not reachable from any constructor |
| New `GetInstanceProperties()` | Only called from the two rows above |

Property accessors, constructors, `DefaultProperties` and the backing fields are byte identical.

Verified against the just in time compiler with `DOTNET_TieredCompilation=0` and
`DOTNET_JitDisasmDiffable=1`, comparing both arms:

- `GetPropertyValue` and `SetPropertyValue`, every instantiation including `Decimal` and `__Canon`:
  byte for byte identical machine code.
- `IInterceptorSubject.get_Properties`: identical machine code, 54 bytes on both arms. The only
  difference anywhere in the dump is a compiler comment moving from `1 single block inlinees` to `2`,
  which is `GetInstanceProperties` being inlined into the same single field read.

No instruction executed by an intercepted read or write differs between the two arms.

### Noise floor

`ServiceOrderResolverBenchmark` contains no `[InterceptorSubject]` type at all, so this change cannot
reach it. Its four rows set the floor for run 2:

| Control row | Δ |
|---|---|
| NoDependencies | +3.08% |
| LinearChain | +0.03% |
| MixedDependencies | -1.47% |
| WithFirstLastGroups | -2.28% |

A 5.4 point spread on code that did not change. Anything inside roughly 3% is not a signal here.

The BenchmarkDotNet error column does not narrow this. It is a confidence interval within a single run,
so it measures precision, not truth, and `--launchCount 3` relaunches the same binary, which reproduces
the same code layout every time. Seven rows exceeded their error band in run 2, the largest at +3.60%,
all inside the control spread. `GenerateSubjectId`, a static method that never touches a subject, was
flagged outside its band in both runs.

### Full suite, run 2

49 shared rows. Median **+0.48%**. Allocations byte identical on 48 of 49 rows.

| Row | base | branch | Δ |
|---|---|---|---|
| `Registry.AddLotsOfPreviousCars` | 56,397,264 ns | 56,768,089 ns | +0.66% |
| `Registry.ChangeAllTires` | 14,923.6 ns | 15,026.1 ns | +0.69% |
| `Registry.Write` | 1,095.2 ns | 1,105.7 ns | +0.96% |
| `Registry.WriteNoOp` | 358.0 ns | 351.4 ns | -1.84% |
| `Registry.Read` | 376.0 ns | 375.8 ns | -0.04% |
| `Registry.IncrementDerivedAverage` | 5,249.6 ns | 5,069.6 ns | -3.43% |
| `Registry.GetOrAddSubjectId` | 30.41 ns | 29.11 ns | -4.28% |
| `SubjectSource.WriteToRegistrySubjects` | 2,924,864 ns | 2,922,661 ns | -0.08% |
| `SubjectUpdate.CreateCompleteUpdate` | 6,694.5 ns | 6,694.1 ns | -0.01% |
| `SubjectTransaction.CommitChanges` (Local) | 34,053.8 ns | 34,136.6 ns | +0.24% |
| `ContextDelegationDepth.WriteWithDelegation` (64) | 636.6 ns | 637.2 ns | +0.09% |

`AddLotsOfPreviousCars` allocates 19,616,720 B against 19,617,329 B, a difference of 609 B on 19.6 MB.

### The same comparison, measured twice

Run 1 and run 2 built the same two commits, so these pairs differ only by measurement:

| Row | run 1 | run 2 |
|---|---|---|
| WriteNoOp | +2.95% | -1.84% |
| Write | -3.37% | +0.96% |
| WriteWithTimestampScope | -5.18% | -0.66% |
| Read | -4.22% | -0.04% |
| DerivedAverage | -2.49% | +1.79% |
| ChangeAllTires | -0.64% | +0.69% |
| AddLotsOfPreviousCars | -0.71% | +0.66% |
| IncrementDerivedAverage | -0.81% | -3.43% |
| GetOrAddSubjectId | -0.21% | -4.28% |
| GenerateSubjectId | -0.15% | +0.36% |
| ReadParents | +0.80% | +0.96% |

Six of eleven flip sign. `WriteNoOp` is the clearest case: it read +2.95% in run 1, which looked like a
regression with non overlapping intervals, and -1.84% in run 2. It is also a row whose machine code was
proven identical above, so neither number can be a cost. Since both runs compared identical inputs, this
table is the noise, and the A/B signal is not distinguishable from it.

### Inheritance rows

New in this branch, so there is no base counterpart to compare against. Absolute numbers from run 2:

| Row | branch |
|---|---|
| RootOnlyGet | 41.75 ns |
| DerivedDeclaredGet | 40.41 ns |
| RootOnlySet | 31.53 ns |
| DerivedDeclaredSet | 30.96 ns |
| BaseDeclaredSetThenGet | 78.59 ns |
| PropertiesAccess | 0.30 ns |
| ConstructThreeLevel | 3,383 ns, 1,480 B |

A property declared on a derived subject costs the same as one declared on a root subject, 40.41 against
41.75 nanoseconds for a read and 30.96 against 31.53 for a write, so the inherited members in derived
mode costs nothing. `BaseDeclaredSetThenGet` performs one write and one read at 78.59 nanoseconds,
against 73.28 for a root write plus a root read. Base declared properties now cost what any other
intercepted property costs, which is the fix rather than a regression: on master that write skipped the
interceptor chain entirely, which is why it was fast and why change tracking never saw it.

### One row that moved, and why it is not a signal

`SubjectSource.WriteToSource` came in at -12.18%, faster, and is the only row in the suite whose
allocation moved, 11,068 B to 12,284 B. It is also the least stable row measured: the base arm's standard
deviation is 13.8% of its mean, with a 5.9% gap between mean and median.

The measured loop calls `PropertyChangeInterceptor.WriteProperty` directly and never goes through a
generated setter, so no changed code is on its path. It ends in `_source.Wait()` on a background
processor with a one millisecond buffer window, so the allocation attributed to one operation depends on
how many buffer flushes happen to land inside it, which makes the figure timing coupled rather than
deterministic.

## Non-regression evidence

Generated output was regenerated for the whole repository and compared against the merge base, which
covers all 360 subjects rather than only the shapes a test was written for.

- 270 generated subject files change, 4,014 lines, resolving into exactly two diff shapes: 256 root
  mode files at 13 changed lines each, 14 derived mode files at 49 each.
- 0 lines unclassified. Every one is an expected category: the three helper modifiers, the new
  `GetInstanceProperties()`, the `Properties` line, the `AddProperties` operand, and in derived
  subjects the removal of the inherited block.
- 0 `DefaultProperties` blocks differ and 0 `.Concat` targets moved, compared as whole blocks rather
  than as key lists, because `ToFrozenDictionary` is last wins so a changed entry would not show up
  as a key difference.

## What changes for you

**Fixed, all broken on master today.** Base-declared properties in a hierarchy were never intercepted,
silently, because the value and `PropertyChanged` both stayed correct (issue #437). Four shapes did not
build at all: a plain class between two subjects (three CS0108), a sealed root subject (CS0628), a hand
written base whose `IRaisePropertyChanged` is explicit (CS0103), and a hand written base without a
static `DefaultProperties` (CS0117). A subject over an ordinary base declaring one of those member names, an MVVM
base for instance, emitted CS0108; that hit seven of ten probed base shapes.

**Newly supported.** A hand written base class can host generated subclasses, against a documented
contract. A hand written subclass can extend a generated subject and use its accessor helpers, which was CS0122
before. Both work across an assembly boundary.

**Breaking, five of them, all loud.**

| | What stops compiling |
|---|---|
| NI0013 | A member named `GetPropertyValue`, `SetPropertyValue`, `InvokeMethod` or `GetInstanceProperties` on a derived subject |
| NI0014 | A member matching an `IInterceptorSubject` member, for example `public object SyncRoot { get; }` |
| `protected` helpers | A hand written subclass declaring one of those four names now gets CS0108, where `private` members hid nothing |
| NI0006 | A `*WithoutInterceptor` method whose stripped name is one the generated half occupies now needs renaming |
| NI0012 | A leaf project on the new package with a referenced model library still built by the old generator. Warning, so it breaks only where warnings are errors |

Two of these are deliberate trades rather than accidents. The wrapper guard rejects by name at any
arity because three signature precise versions each admitted a silent capture, and a rename request is
recoverable where a capture is not. NI0014 now reaches into referenced assemblies, so a matching member
on a third party class above the contract provider is an error a consumer cannot fix in their own code;
it only fires where the member genuinely displaces something.

**Not a regression, but it will read as one.** A base declared write stops taking the no interception
fast path and now performs the interception it was silently skipping, so it costs what any other
intercepted property costs. On the branch a write plus a read on a base declared property measures
78.59 ns, against 73.28 ns for the same pair on a root declared property.

## Progress

- [x] Task 1: test harness and green pins
- [x] Task 2: resolve base class facts from the nearest subject ancestor
- [x] Task 3: sealed root subjects
- [x] Task 4: split the interception members into root mode and derived mode
- [x] Task 5: the subject base contract, NI0011 and NI0012
- [x] Task 6: NI0013 and NI0014
- [x] Task 12: generator simplification sweep (runs after task 6, gated on byte-identical generated output)
- [x] Task 7: hand written base and subclass, end to end
- [x] Task 8: upgrade the two tests that look like coverage and are not
- [x] Task 9: whole repository regeneration gate
- [x] Task 10: benchmarks
- [x] Task 11: documentation
- [x] Task 13: full benchmark suite against master (registry smoke test, then everything)

Maintainer design notes: `docs/design/generator-supported-shapes.md`
User documentation: `docs/generator.md`, `docs/subject-guidelines.md`


src/Namotion.Interceptor.Tests/VirtualPropertyIntegrationTests.cs
```diff
@@ -64,6 +65,41 @@ public void VirtualPropertyChain_WithMultipleLevels_WorksCorrectly()
         Assert.Equal("Bob", manager.Name);
         Assert.Equal(5, manager.TeamSize);
     }
+
+    [Fact]
+    public void WhenWritingThroughAThreeLevelHierarchy_ThenBaseAndDerivedWritesAreIntercepted()
+    {
+        // Arrange: the value alone proves nothing here, because a base-declared property kept its
+        // value even while base-declared writes bypassed the interceptor chain (issue #437).
```
RicoSuter: dont link issues in code please (check other code as well)

src/Namotion.Interceptor.Generator/SubjectMetadataExtractor.cs
```diff
@@ -504,11 +494,31 @@ method.ReturnType is RefTypeSyntax ||
...
                 var methodName = fullMethodName.Substring(0, fullMethodName.Length - InterceptedMethodPostfix.Length);
+
+                // The wrapper is emitted into the same generated half as the plumbing, so a wrapper
+                // named like a plumbing member captures the generated call sites instead of the
+                // plumbing. In derived mode the plumbing is inherited, so this is a silent capture:
+                // the only compiler signal is CS0108, which a consumer without TreatWarningsAsErrors
+                // never sees, and the subject then reports whatever the wrapper returns. An
+                // AddProperties wrapper is quieter still, taking the interface slot from the root's
+                // explicit implementation with no compiler diagnostic at all. NI0013 cannot catch
+                // either, because it scans declared members rather than emitted ones.
```
RicoSuter: compact the docs? is this all needed and not clear from code?

src/Namotion.Interceptor.Tests/VirtualPropertyIntegrationTests.cs
RicoSuter: Done in b0af7f2d. Four comments on this branch carried an issue number; all four now state the fact directly instead.

Six more exist outside this branch (`OpcUaStatusCodeClassifier` #332, `OutboundWriter` #287, `ContextServiceResolutionTests` and `ServiceOrderResolverTests` #380, `ContextConcurrencyTests` #403, `SubjectTransaction` #343, plus a URL in `ReportedIssueTests`). Left alone so this PR stays scoped. Say the word and I will do them as a separate PR.

src/Namotion.Interceptor.Generator/SubjectMetadataExtractor.cs
RicoSuter: Cut from 8 lines to 3 in b0af7f2d. You were right that most of it was restating the call and the diagnostic message. What is kept is the part neither shows: the capture is silent, because in derived mode the only compiler signal is a CS0108 that a consumer without `TreatWarningsAsErrors` never sees, and an `AddProperties` wrapper produces none at all. The long rationale for why the guard is name-only stays on `CollidesWithGeneratedMember`.

src/Namotion.Interceptor.Generator/GeneratedMemberTable.cs
```diff
@@ -0,0 +1,170 @@
...
+/// consumer cannot edit.
+/// </summary>
+internal static class MemberNames
+{
+    public const string GetPropertyValue = "GetPropertyValue";
+    public const string SetPropertyValue = "SetPropertyValue";
+    public const string InvokeMethod = "InvokeMethod";
+    public const string GetInstanceProperties = "GetInstanceProperties";
+    public const string PropertyChanged = "PropertyChanged";
+    public const string RaisePropertyChanged = "RaisePropertyChanged";
```
RicoSuter, line 18: should we use nameof(interface.property) here instead? or not possible?

src/Namotion.Interceptor.Generator/GeneratedMemberTable.cs
```diff
@@ -0,0 +1,170 @@
...
+    /// </summary>
+    public const string DefaultProperties = "DefaultProperties";
+}
+
+/// <summary>
+/// The return type the contract check demands of one plumbing helper. An enum rather than a symbol,
+/// because two of the four are answered from the method itself and one has to be constructed from
+/// the compilation.
+/// </summary>
+internal enum PlumbingReturnKind
```
RicoSuter: not sure about the name "plumbing" seems off in this PR, not clear what this is about

src/Namotion.Interceptor.Generator/GeneratedMemberTable.cs
```diff
@@ -0,0 +1,170 @@
...
+            PlumbingReturnKind.PropertyMetadataDictionary, RequiresLeadingString: false,
+            "protected IReadOnlyDictionary<string, SubjectPropertyMetadata>? GetInstanceProperties()")
+    ];
+
+    // The four fields below each read the ones above them, and a static field initializer runs in
+    // textual order, so reordering them leaves one null and takes the generator down with a
+    // TypeInitializationException on every compilation.
+
+    private static readonly string[] HijackableInterfaceMembers =
+        ["Context", "Data", "SyncRoot", "AddProperties"];
```
RicoSuter, line 107: also here avoid magic strings, possible?

src/Namotion.Interceptor.Generator/GeneratedMemberTable.cs
```diff
@@ -0,0 +1,170 @@
...
+            .ToArray();
+    }
+
+    /// <summary>
+    /// Whether a method matches the RaisePropertyChanged(string) the root emits. Read by the notify
+    /// lookup, which asks whether a base already answers that call, and by the hiding check, which
+    /// asks whether the emitted one needs a 'new'.
+    /// </summary>
+    public static bool IsRaisePropertyChangedSignature(IMethodSymbol method)
+        => method.TypeParameters.Length == 0 &&
```
RicoSuter, line 163: dont we need to also check the name?

src/Namotion.Interceptor.Generator/SubjectBaseContract.cs
```diff
@@ -0,0 +1,265 @@
...
+/// Everything the generator needs to know about the class a subject inherits from: which ancestor
+/// owns the shared plumbing, and whether that ancestor exposes enough of it to be inherited from.
+/// </summary>
+internal static class SubjectBaseContract
+{
+    /// <summary>
+    /// Resolves everything the emitter needs to know about the base class, and reports NI0011 to
+    /// NI0014. A null result means generation is suppressed and the caller must emit nothing.
+    /// </summary>
+    public static BaseClassInfo? Resolve(
```
RicoSuter: rename BaseClassInfo to SubjectBaseClass, or maybe even merge SubjectBaseContract and BaseClassInfo together while Resolve is the static factory method and it creates "itself"? and rename to SubjectBaseClass? what do you think?

src/Namotion.Interceptor.Generator.Tests/GeneratorTestHost.cs
```diff
@@ -1,16 +1,37 @@
...
 using Xunit;
 
 namespace Namotion.Interceptor.Generator.Tests;
 
+/// <summary>
+/// Records every intercepted write, which is what separates "the value looks right" from "the
+/// executor really saw it". Shared, because a generated shape is only proven by running it and more
+/// than one test class needs the same evidence.
+/// </summary>
+internal sealed class RecordingWriteInterceptor : IWriteInterceptor
```
RicoSuter: check that this is always reused, i think i saw another impl of that somewhere

src/Namotion.Interceptor.Generator/GeneratedMemberTable.cs
RicoSuter: Not possible in the generator. It ships as `analyzers/dotnet/cs` and deliberately does not reference `Namotion.Interceptor`, so none of those types are symbols there. Five of the seven names (`GetPropertyValue`, `SetPropertyValue`, `InvokeMethod`, `GetInstanceProperties`, `DefaultProperties`) name members the generator *emits into consumer code*, so even with a reference there would be nothing to point `nameof` at.

Pinned from the test project instead, in 06259d60, since that one does reference the library: `GeneratedMemberTableTests` asserts `MemberNames.PropertyChanged` and `MemberNames.RaisePropertyChanged` against `nameof(INotifyPropertyChanged.PropertyChanged)` and `nameof(IRaisePropertyChanged.RaisePropertyChanged)`. The five emitted names stay unpinnable by construction; the emitter writes their signatures and the compiler checks those, with the 278 snapshot files covering the rest.

src/Namotion.Interceptor.Generator/GeneratedMemberTable.cs
RicoSuter: Same reason: no reference to `Namotion.Interceptor` in the analyzer, so no `nameof`. Pinned by test in 06259d60 instead, through `CollidesWithGeneratedMember(nameof(IInterceptorSubject.Context))` and the other three, so a typo in the array shows up as the guard failing to recognise the real interface member name.

Worth flagging what I found while checking: the obvious structural fix, enumerating `IInterceptorSubject`'s members from the `Compilation` instead of listing names, would be a bug. The interface has five members and this list has four. `Properties` is deliberately excluded, because every subject emits its own explicit implementation of it which always wins, so treating it as hijackable would report every legitimate generated hierarchy. That was documented in the design doc but nothing at the array said so, which is presumably why it read as a lazy copy. There is now a comment there, and a test pinning the exclusion.

src/Namotion.Interceptor.Generator/GeneratedMemberTable.cs
RicoSuter: The name is checked, but by the caller, not here: both call sites reach this through `AccessibleMembers`/`HidableMembers` keyed on `MemberNames.RaisePropertyChanged`, which do `GetMembers(name)`. So every method reaching this predicate already carries that name and a check here would be dead.

Your reading is the finding though. A method called `IsRaisePropertyChangedSignature` that does not check the name is claiming something it does not do. Renamed to `HasRaisePropertyChangedParameters` in 06259d60, and the doc now says the name is the caller's lookup key.

src/Namotion.Interceptor.Generator.Tests/GeneratorTestHost.cs
RicoSuter: You saw right, and there were more than one. The generator test project had **eight** declarations of three classes, every copy byte-identical:

- `RecordingWriteInterceptor` x3 (`GeneratorTestHost`, `GeneratorShapeBehaviorTests`, `ExplicitInterfaceBehaviorTests`)
- `RecordingReadInterceptor` x3 (`BaseClassInterceptionBehaviorTests`, `GeneratorShapeBehaviorTests`, `ExplicitInterfaceBehaviorTests`)
- `RecordingMethodInterceptor` x2 (`BaseClassInterceptionBehaviorTests`, `GeneratorShapeBehaviorTests`)

Two of each predate this branch. What this branch did was add another copy and label it shared, which it was not, since the existing ones kept their own.

Collapsed into `RecordingInterceptors.cs` in 5c9189b1. No behaviour change, because every copy was identical. 2,968 tests green.

Out of scope and left alone: `Namotion.Interceptor.Tests` has its own `RecordingWriteInterceptor`, plus `CountingWriteInterceptor` twice in `ContextSubtreeServiceTests` and `ContextDelegationCycleTests`. Sharing across test projects needs a common test library, which is a separate change.

docs/subject-guidelines.md
```diff
@@ -448,6 +470,163 @@ public partial class HueBridge
 - **HueBridge** (`Namotion.Devices.Philips.Hue`): Injects `IHttpClientFactory` and `ILogger<HueBridge>` alongside the interceptor context for HTTP communication with the Hue Bridge API.
 - **OpcUaSubjectServer** (`Namotion.Interceptor.OpcUa`): Injects OPC UA server configuration and telemetry services.
 
+## Hand-Written Base Classes and Subclasses
```
RicoSuter: should we move this into the generator.md file as last chapter and link to it from this document here instead? this is the "tutorial" page and should not go into that much detail

src/Namotion.Interceptor.Generator/GeneratedMemberTable.cs
RicoSuter: Renamed in 7c5c696e. `plumbing` is now zero occurrences repo-wide.

The set the generator emits into a subject is **interception members**. The four methods generated accessors call are **accessor helpers**, which is a distinction the code already made without naming it.

The part that mattered most: five uses were in diagnostic text, where a reader meets the word with no context. NI0012's title is now "Base class interception members cannot be shared" and its message "...so '{1}' emits its own interception members and the base class's own properties stay unintercepted".

Identifiers: `PlumbingMethods` to `AccessorHelpers`, `PlumbingMethodShape` to `AccessorHelperShape`, `PlumbingReturnKind` to `AccessorHelperReturnKind`, `RootModePlumbingMemberNames` to `RootModeMemberNames`, `FindHiddenPlumbingMembers` to `FindHiddenInterceptionMembers`, `EmitsPlumbingHere` to `EmitsInterceptionMembers`, plus eleven test method names.

Prose was rewritten phrase by phrase rather than word-substituted, because the replacement is plural where the metaphor was a mass noun: "the plumbing lives in source" had to become "the interception members live in source". Verified: emitted output byte identical across all 278 files, no broken doc anchors, 2,968 tests green.

src/Namotion.Interceptor.Generator/SubjectBaseContract.cs
RicoSuter: `BaseClassInfo` is now `SubjectBaseClass` (7c5c696e). The `Info` suffix carried nothing.

I did not merge it into `SubjectBaseContract`, and did not take either resolver name. Two reasons.

On the merge: `SubjectBaseContract` holds about 180 lines of contract checking (`SatisfiesContract`, `HasUsableDefaultProperties`, `HasExpectedReturnType`, `HasAccessibleMethod`) that has nothing to do with being a data carrier. Merged, the record's file would be roughly 250 lines of not-the-record, which is the split we just did, undone.

On the name: "contract" is already the user-facing term. NI0011's message says the base "does not satisfy the subject base contract", and `docs/subject-guidelines.md` documents it under that name. Renaming to `SubjectBaseClassResolver` breaks the tie between the diagnostic a user reads and the class that produced it, and `SubjectBaseClassResolver.Resolve` stutters. Between your two, `Resolver` would beat `Factory` though: the class does not manufacture anything from parameters, it resolves facts about a base that already exists.

Happy to take the resolver name if you would still rather have it.

docs/subject-guidelines.md
RicoSuter: Agreed and done in e78acdee. The contract is reference material rather than tutorial: it is exactly what NI0011 and NI0012 tell a user to satisfy, and both diagnostics are already documented in `generator.md`. It is now that file's last chapter.

`subject-guidelines.md` keeps the ordinary case (generated over generated, which needs nothing from the reader) and links across for the rest. It drops from 698 to 542 lines.

Four things depended on the old location and moved with it:

- `SubjectBaseShapeTests.ReadDocumentedConformingBaseFixture`, which reads the conforming base **out of the documentation** so the documented contract and the tested one cannot drift. It now reads `generator.md`. Worth noting it selects the block by content (`Assert.Single(blocks, b => b.Contains("class TrackedEntityBase"))`), not by position, so the much larger file it now scans does not affect it, and it fails loudly on zero or duplicate matches.
- Two links in `generator.md` that pointed across files and are now same-file anchors.
- Two comments in `SubjectBaseContract` pointing at the contract's location.
- The forward pointer in `subject-guidelines.md`, now a cross-file link.

Verified: 0 broken anchors across all of `docs/`, 244 generator tests including the fixture one, 2,968 solution-wide.

PR #440, opened 2026-08-08
feat: Single ownership for subject bound hosted services

RicoSuter:
One owner for the start, stop and disposal of a hosted service bound to a subject. A hosted service now runs exactly while its subject is in the graph: the `HostedServiceHandler` on the subject's context is the only thing that starts or stops it, and it disposes exactly what it created. Attachments become factories rather than instances handed over, which is what makes a re-attach produce a fresh instance instead of silently starting nothing.

Consumers rename three calls and follow the compiler; there are no shims. See Breaking changes.

- Follow-up: #541, #542, #543, #544
- Related to #540, #456

## Why

- **Double start.** `AddHostedSubject<T>` registered the subject as a hosted service *and* the context started it. Measured `StartAsync` count: 2. On a `BackgroundService` the second call orphans the first `_executeTask` beyond the reach of any cancellation.
- **Context silently dropped.** The context was passed only when a constructor accepted one, which the generator emits only for parameterless shapes. All six HomeBlaze device subjects declare a constructor with dependencies, so every one ran with no tracking, no registry and no lifecycle.
- **Detach never disposed.** `OpcUaClient` worked around it by hand; `OpcUaServer` omitted the dispose entirely. Two copies of one pattern already disagreed on its subtlest step.
- **Re-attach was a silent no-op.** Detach removed the attachment outright, and restarting a stopped instance is not an option because a disposed connector cannot restart.
- **Two contexts on one `IServiceCollection` lost the second handler.** `AddHostedService` dedupes on implementation type, so every subject in that context silently never started.
- **Unbounded shutdown.** A service that did not return from `StopAsync` held the process indefinitely: measured 15 s and counting against a 3 s `ShutdownTimeout`. The drain barrier now observes the token.

## Contract

- **In the graph implies running, out of the graph implies stopped.** Each target drains in append order, so it ends in the state its last event demanded. Execution across targets is concurrent, so this is quiescent consistency rather than a moment by moment guarantee.
- **One instance per target.** A target holds at most one non null instance, and only a start body sets it.
- **Created implies disposed by the same owner.** The handler disposes exactly the instances it created through a factory, once, and never disposes a subject.
- **A subject's stop precedes the disposal of its own attachments**, on both the context detach path and the shutdown path, whenever that stop is not cancelled.
- **The factory must construct on every call.** The handler owns what it creates, so `() => existingInstance` is refused with a fault rather than used after dispose, and `() => serviceProvider.GetRequiredService<T>()` hands the handler something the container also owns.
- **Transitions for one target never interleave; unrelated targets run concurrently.** A hosted service must therefore not detach an attachment from inside its own stop path, and a dispose path must not write a subject typed property or attach a subject.

## Breaking changes

**API.** Three renames and one constraint change. No shims: update the package and follow the compiler.

Two renames, two parameter type changes and one constraint change. No shims: update the package and follow the compiler.

| Was | Now |
|---|---|
| `services.AddHostedSubject<T>()` | `services.AddSubject<T>()` |
| `subject.AttachHostedService(instance)` | `subject.AttachHostedService(() => new T(...))`, returning a handle |
| `subject.DetachHostedService(instance)` | `subject.DetachHostedService(handle)` |
| `subject.GetAttachedHostedServices()` | `subject.GetHostedServiceAttachments()`, handles rather than instances |

`AddSubject<T>` constrains `T` to `IInterceptorSubject` where `AddHostedSubject<T>` constrained it to `IHostedService`. A registered hosted service that is not a subject belongs on `AddSingleton` plus `AddHostedService` instead.

The factory must construct on every call. The handler owns what it creates and disposes it when the subject leaves the graph, so `() => existingInstance` is refused with a fault rather than silently used after dispose, and `() => serviceProvider.GetRequiredService<T>()` hands the handler something the container also owns.

**Behavior.** A detach now stops and disposes what the handler created, where before it disposed nothing. A re-attach after a detach now produces a fresh instance from the factory, where before it silently started nothing. `AddSubject<T>` always supplies the context, where `AddHostedSubject<T>` supplied it only when a constructor accepted one. Log message shape changed: `{ServerUrl}` and `{Path}` collapsed into `{Service}` plus `{Target}`; categories are unchanged.

**Contract.** `AttachHostedService` and `AttachHostedServiceAsync` now enter the lifecycle lock, which widens two documented constraints: the rule for `IStartupCompletionDeferrer` implementers and for hosted service dispose paths goes from "must not write a subject typed property" to "must not attach a hosted service either". Both are updated in `docs/hosting.md` and the design document.

## The model, before and after

```mermaid
flowchart LR
  subgraph before ["Before: one global queue"]
    direction TB
    E1["subject A attaches"] --> Q["single BufferBlock"]
    E2["subject B detaches"] --> Q
    E3["attachment C stops"] --> Q
    Q --> C["one consumer<br/>50 ms each, in series"]
  end
  subgraph after ["After: one chain per target"]
    direction TB
    F1["subject A"] --> G1["chain A"]
    F2["subject B"] --> G2["chain B"]
    F3["attachment C"] --> G3["chain C"]
  end
```

| | Before | After |
|---|---|---|
| Start and stop dispatch | one global queue, one consumer | one serialized chain per target |
| Unrelated services | coupled through the queue | independent |
| Attachment | an instance you hand over | a factory the handler invokes |
| Who disposes | nobody | whoever created it, so the handler |
| Re-attach after detach | silently starts nothing | fresh instance from the factory |
| Registration | `AddHostedSubject<T>` | `AddSubject<T>` |
| Context supply | only if a constructor accepts one | always, after construction |
| N subjects starting | N x 50 ms, in series | constant, pinned by an overlap test |

## Found by review rather than by writing

Every defect below was found after the implementation was complete, by reviewers who did not write it, and each is fixed with a test that fails without the fix.

- **Orphaned service.** A detach racing an attach left a service running for an attachment the subject no longer had: started once, never stopped, never disposed.
- **`configure` ran too late**, so for one constructor shape the subject started with default configuration.
- **The compiler steered callers into the one unsupported shape.** `AttachHostedService(myService)` gives CS0411, whose suggested fix does not work, and the only minimal edit that compiles is `() => myService`, which produced a silent use after dispose. That shape is refused at runtime with a fault.
- **A fourth deadlock shape** the design document claimed did not exist. Three reviewers reached it independently.
- **Two value-equal subjects shared one liveness entry.** Every other durable subject-keyed map in the repository passes `ReferenceEqualityComparer`, the registry and the lifecycle interceptor included; this one did not, so detaching either subject cleared the other's liveness while it was still in the graph and its next start declined, leaving a subject in the graph with nothing running and no fault. Reproduced with plain graph writes.
- **A stop body carried the ambient startup scope.** A body inherits the execution context of the flow that appended it, and stop bodies do not wait for that flow's scope, so a service attached from a stop path was captured by an unrelated caller's scope and parked until that caller closed it. Where the stop path awaited its own attach, it wedged the whole target chain.
- **The handle could not distinguish "being created" from "settled".** Covered under Public API above. The reproduction was committed as a skipped test earlier in this branch and is no longer skipped.
- **A faulted awaited attach leaked an instance a queued start created.** `AttachHostedServiceAsync` publishes the attachment before it awaits the start, so a second start can be appended against the same target while the first is in flight. When the awaited start faulted, the same completion released both the caller's cleanup and that queued body, and nothing ordered the two: the body's ownership read refused it only if the release landed first. A probe that forced the queue measured the body winning between a quarter and a half of the time, each iteration leaving an instance started, never stopped and never disposed after the host had stopped. The fix does not order them, it stops depending on the order: the cleanup appends a stop, which the chain puts behind any start already appended. One line. Found by the final review of the design document rather than of the code, which is the second defect this branch's own documentation turned up.
- **The wrapper declared an invariant and then broke it two lines later.** The OPC UA wrapper states that the status message is the text behind `Error` alone. One commit fixed the stop path by clearing the message ahead of the status, and classified the same symptom elsewhere as a test-side race, working it around in the test. It was not a race: two more sites wrote the pair in the order that leaves `Stopped` or `Starting` standing beside the failure text, and each write is its own pass of the interceptor chain, so the window is wide. The rule turns out to be directional, which is why one order everywhere would not have fixed it: moving to `Error` writes the status first, leaving `Error` clears the message first. One test per site.
- **Mutating the subject before resolving the handler** left an attachment stored, or an instance running and unreachable, when the lookup threw. Fixing the order then exposed a second, more reachable defect: an attach and a context entry read and write the same two facts in opposite orders, so with the lookup first *both* sides can miss. The attach now does both, and the second read counts rather than throwing.

## Public API

`HostedServiceAttachmentState` and `IHostedServiceAttachment.GetState` are new. An attachment exposed `Current` and `Fault` and nothing else, so with both null a consumer could not tell "never started" from "being created right now", from "previous instance disposed on a re-attach", from "detached for good". The one consumer guessed, and guessed wrong: it treated a null `Current` as proof that anything published was stale and dropped it, which on a re-attach takes the tree the factory has just published back out of the graph, because `Root` is subject typed. The six members name the cases, and `Stopped` against `Removed` is the distinction to act on, recoverable against terminal.

`GetState(out T? current)` hands the state and the instance out of one snapshot load. The target keeps them in one field precisely so they cannot disagree, and exposing them as two properties handed that disagreement back to every consumer. There is therefore no `State` property alongside it: one load, one reading, and no way for a caller to pair a state with an instance read a moment apart from it. `Running` is defined as a non null instance, so a reading that hands back null can never say `Running`, and the consumer rule that existed to catch that mismatch is gone rather than documented. The generic interface loses its covariance to make room, since C# will not take a covariant parameter in an `out` position; nothing in the tree used it, and the interface ships for the first time here.

`LifecycleInterceptor.TryRunWhileAttached` is new public surface in `Namotion.Interceptor.Tracking`, because the hosting package needs to record liveness under the graph lock. `InternalsVisibleTo` would also work and the repository already uses it in places, but it would make an internal of Tracking part of Hosting's build contract with nothing stating the constraints that come with running a callback under that lock. A public member carries them in its remarks, which is where they are.

`AttachHostedService` and `AttachHostedServiceAsync` keep their names, and take a factory rather than an instance as the migration table above sets out. On top of that they now enter the lifecycle lock, which they did not before. That widens two documented constraints: the rule for `IStartupCompletionDeferrer` implementers and for hosted service dispose paths goes from "must not write a subject typed property" to "must not attach a hosted service either", and residual hazard 4's cycle drops from three parties to two. Both are updated in `docs/hosting.md` and in the design document.

## Open gaps

| Gap | Impact | Why it is open |
|---|---|---|
| No successful `OpcUaServer` attach is tested | Its start, stop and diagnostics mapping are covered only through the shared attachment host, which the client exercises fully | A successful server attach binds a port, and the OPC UA integration suite hardcodes 4840. Needs a port lease first |
| The diagnostics poll's non-blocking gate is unpinned | A blocking wait there passes every test | It is observationally identical except at a shutdown that must cancel the loop |
| Four narrow properties argued, not pinned | Each is masked by the guard beside it, so deleting one alone leaves the suite green. Three change no outcome at all. The fourth, the liveness read inside the start body, costs a create and teardown against a subject that has already left the graph, which for a connector is a session opened and closed | Labelled in the design document rather than overclaimed |
| The detach benchmark cannot resolve a regression | A future regression on the detach callback would not be caught the way the attach one is | Its work is a dictionary removal, which allocates nothing and costs nanoseconds, so it is only visible against a much smaller baseline than the row provides |

Everything else in the handler is pinned by a test that fails when that mechanism alone is deleted and stays green when its siblings are deleted. That claim is measured, not asserted: a mutation pass over the handler's documented mechanisms killed each by exactly the test the document names, and every survivor is written down rather than quietly left. The inventory has grown with the branch and now runs to sixteen coverage gaps and two labelled equivalent mutants.

## Hazards not guarded against

Deadlock shapes a consumer can build that the library does not detect at runtime, all set out in `docs/design/hosting-service-ownership.md`. Per target chains removed the cross service deadlock that one shared queue had; they do not remove a cycle inside one chain. The status column says whether each is a rule for consumers, an accepted limitation, or an open defect.

| Shape | Kind | Status |
|---|---|---|
| An attachment whose own `StopAsync` detaches itself | Consumer contract | Stated, not enforced. Detecting it means tracking waits across chains, which is not worth its cost against a rule that can be written down |
| Two subjects whose services detach each other's attachments | Consumer contract | The same cycle across two chains. Same status |
| A subject that detaches its own attachment from inside its own stop path | Consumer contract | The only one that occurs in practice. Both OPC UA wrappers had it and were changed, and a regression test pins the fix |
| An `IStartupCompletionDeferrer` whose `DeferCompletion` takes a lock | Consumer contract | Reachable only through a consumer implementation. The constraint is on the interface's own remarks, where an implementer will read it |
| Awaiting a captured start, or its detach, inside its own startup scope | Accepted limitation | Not a rule a consumer can simply follow: `host.StartAsync()` reaches it without anyone writing anything unusual. `HostOptions.StartupTimeout` turns the hang into a failure |
| A cross handler ownership race | Open defect, #576 | Internal rather than a consumer shape, which is why it is not one of the five above. Documented as derived rather than measured, and filed rather than fixed |

## Performance

Measured on a pinned i7-11700K, performance governor with `scaling_min_freq` equal to `scaling_max_freq` and turbo off, verified in sysfs rather than from the BenchmarkDotNet header, which reports the hardware minimum even when correctly pinned. `HostingLifecycleBenchmark` does not exist on master, so the base arm is master with that one file, a project reference and a package reference grafted on, which makes both arms run identical benchmark code against different libraries. The base was re-grafted onto current master so the comparison is against the tree this branch merged, not an older one. Refs: branch `181a5c5d5` against `943db5f91`, the benchmark grafted onto master `7f6182ffc`. The pin was confirmed in sysfs before and after every run; the BenchmarkDotNet headers disagree with each other across arms, which is the documented artifact rather than drift.

**No regression. Allocations are byte identical between branch and master in all twelve rows**, including the arms that enable hosting, and the timings do not clear the noise floor the unreachable rows set in the same run.

20,000 subjects, `InvocationCount=1 IterationCount=40 LaunchCount=3 WarmupCount=8 MemoryRandomization=True`. Before is master, after is this branch.

| Benchmark | Master | Branch | Delta | Allocated |
|---|---:|---:|---:|---|
| AttachWarm, Hosting | 31.09 ms | 31.08 ms | 0.0% | 10,312.56 KB, identical |
| AttachCold, Hosting | 32.44 ms | 31.43 ms | -3.1% | 14,687.95 KB, identical |
| Detach, Hosting | 14.44 ms | 13.81 ms | -4.4% | 5,000.06 KB, identical |
| AttachWarm, HostingOneHosted | 32.00 ms | 33.33 ms | +4.2% | 10,312.69 KB, identical |
| AttachCold, HostingOneHosted | 31.95 ms | 31.38 ms | -1.8% | 14,688.57 KB, identical |
| Detach, HostingOneHosted | 13.90 ms | 14.30 ms | +2.8% | 5,000.16 KB, identical |
| _noise reference_ AttachWarm, None | 28.31 ms | 28.30 ms | 0.0% | 9,218.81 KB, identical |
| _noise reference_ AttachCold, None | 29.04 ms | 28.38 ms | -2.3% | 13,594.20 KB, identical |
| _noise reference_ Detach, None | 13.33 ms | 12.87 ms | -3.5% | 5,000.06 KB, identical |
| _noise reference_ AttachWarm, EmptyHandler | 28.63 ms | 28.22 ms | -1.4% | 9,218.81 KB, identical |
| _noise reference_ AttachCold, EmptyHandler | 28.61 ms | 29.40 ms | +2.8% | 13,594.20 KB, identical |
| _noise reference_ Detach, EmptyHandler | 12.99 ms | 13.23 ms | +1.9% | 5,000.06 KB, identical |
| _noise reference_ ServiceOrderResolver.LinearChain | 1.786 us | 1.807 us | +1.2% | 3.63 KB, identical |

The six noise references enable no hosting, so this change cannot reach them, and they set the floor themselves: they span -3.5% to +2.8%. No hosting row clears that. The two extremes of the whole run both sit in hosting arms and point in opposite directions, Detach/Hosting at -4.4% against AttachWarm/HostingOneHosted at +4.2%, which is sign flipping within one arm rather than a regression. Standard deviations run 2 to 4% of the means, wider than most of the deltas.

**Master was checked for a regression of its own**, because the Sonar rollout refactored `ServiceOrderResolver`, `InterceptorSubjectContext` and the source generator between the first benchmark and this one. New master against old master, 28 rows over `ServiceOrderResolverBenchmark`, `ContextDelegationDepthBenchmark` and `RegistryBenchmark`: allocations identical everywhere except one row 86 B lower out of 19.6 MB, and timings scattered across plus or minus 3% with the signs flipping. Two rows that looked like outliers, `DerivedAverage` at +3.3% and `GetOrAddSubjectId` at +9.7%, were re-run: the first came back at +0.3% and the second at identical medians, 29.06 ns against 29.07 ns, with `GenerateSubjectId` flat at +0.01% as a control. Neither reproduced, so master is unchanged and this branch's base is sound.

The previous description claimed 56.0 B per subject on attach falling to 0.0 B. **That claim is not reproducible with this harness**, and the reason is worth recording rather than hiding: the benchmark's own allocation column moves by exactly that amount depending on which sibling rows run beside it. The same commit reads 9 MB in the hosting arm when `AttachWarm` runs alone and 10.07 MB when all twelve rows run, which over a 20,000 subject graph is 56 B per subject. Measured on four trees, the reading tracks the run configuration rather than the code. Filed as #540, together with a second finding: the benchmark script cannot pass two filter patterns, because BenchmarkDotNet then honours only the positional ones.

## Diff composition

|---|---:|---:|---:|---:|
| Production code | 23 | 2,317 | 742 | +1,575 |
| Tests and benchmarks | 44 | 7,755 | 421 | +7,334 |
| Documentation | 5 | 893 | 103 | +790 |
| Project and CI files | 2 | 5 | 0 | +5 |
| **Total** | **74** | **10,970** | **1,266** | **+9,704** |

Roughly three quarters of it is tests.

Comment density in `Namotion.Interceptor.Hosting` reached 39.7 percent while this branch was being written, against a repository average of 23. The rationale that belongs in the design document was moved there, removing 253 comment lines with no code line changed, leaving the package at 33.0 percent. The inline share, which is what the style rule targets, is 16 percent in the handler and 9 in the target; the remainder is API contract.

## Verification

- [x] Documentation updated: `docs/hosting.md`, `docs/subject-guidelines.md`, `docs/design/hosting-service-ownership.md`
- [x] Unit tests: 4,147 passed, 1 skipped, 0 failed across 35 assemblies. The one skip is a deliberate reproduction of a defect deferred to a follow-up, with a self contained skip reason. The second skip this branch carried is gone: the state work below fixed the defect it reproduced, so that test now runs
- [x] Integration tests: HomeBlaze, MQTT, OPC UA and WebSocket suites all pass in CI on the current head
- [x] Benchmarks: `HostingLifecycleBenchmark`, all twelve rows, plus two noise references, against current master with the benchmark grafted on
- [x] Chaos tests: Connector Tester `opcua-chaos` against the current head, 125 cycles, 125 pass, 0 fail, 700 injected chaos events, 13.2 million value mutations
- [ ] Load tests

**Chaos was run against the current head; load was not.** `opcua-chaos` completed 125 cycles: every cycle passed, 700 chaos events were injected across the server and both clients, and every cycle converged. Managed heap climbs in steps with flat treads and ends flat: the final plateau runs 42 cycles at +0.003 MB per cycle, range 86.9 to 88.0 MB. Over cycles 5 to 100 the slope is +0.087 MB per cycle against +0.114 for the earlier run on this branch, so the current head is flatter than the tree that run already passed, and both sit at the drift the harness produces on unmodified code, which is why a single arm trend is never a regression finding on its own. Convergence is sub-100 ms in 124 of 125 cycles with one 10.0 s outlier on the heaviest profile; the earlier run had one 5.0 s outlier and the same shape otherwise. The `mqtt-chaos` and `websocket-chaos` profiles were not run, and neither were the load profiles: this branch changes when hosted services start and stop rather than how much traffic they carry, so throughput and latency are not the axis it moves.

## Final review round

Four independent reviewers went over the branch with separate lenses: concurrency and the memory model in the hosting core, regressions against master, the HomeBlaze consumer, and error paths across the whole diff. Every finding below was reproduced or verified before being acted on, and the ones that did not survive verification were dropped rather than reported.

**Two were fixed.**

- **A hosted service whose own `StartAsync` threw was never stopped.** `CompleteStart` is not reached, so `Current` stays null and every later stop returns at its null check; the handler never disposes a subject, so such a subject got no cleanup at all. Master stopped it, and so does the generic host, which a probe confirmed rather than assumed. The failure path now stops the instance before disposing it. Pinned by two tests, one for the attachment and one for the subject, both verified to fail with the mechanism removed.
- **The OPC UA wrapper could report `Running` for a disposed source, permanently.** The run loop's unwind reports `Stopped` without the attachment gate, deliberately, because taking it there waits on the transition the unwind runs inside. Every status write that claims the wrapper is doing something now commits under a lock the stop report also takes, against a generation captured where the operation took the gate. This took three attempts: the first guarded only the reconciliation and left the same bug reachable through `StartAsync`, and the second left a capture that compared the generation against itself. Both were caught by review, and all seven commit sites were then enumerated by hand.

**Two were classified rather than fixed.** A double start reachable through `AddSubject` is now #585, and is a narrowing of a defect master has unconditionally rather than a regression. A throwing `ILogger` can skip cleanup at five sites, which was judged not worth defending against, since wrapping logger calls is not standard practice; the one part that stands on its own, cleanup belonging in a `finally`, was applied.

**Known gap in the evidence.** The status ordering fix is not pinned by a test. `RunAsync` calls `StartAsync` before its poll loop and the unwind is downstream of that loop, so a start cannot race its own unwind; the race needs a second start from the UI operation, and on that path the only await returns immediately for both wrappers. Pinning it needs a direct `SingleAttachmentHost` test with a fake owner, which needs `InternalsVisibleTo` that the package does not have today. The fix rests on hand enumeration of every commit site plus two adversarial reviews.

## Follow-ups, not in this PR

- **#541: `MethodPropertyInitializer` cannot re-attach a subject carrying an `[Operation]`**, which is 30 source files in HomeBlaze. It rebuilds its discovered-method set per attach and never checks what is registered, and the generated `AddProperties` throws on a duplicate key. Pre-existing, and it makes the re-attach path this PR builds unreachable in the application until it is fixed, which is why the two wrapper defects below are also deferred.
- **The OPC UA wrapper cannot clear a fault**, and the UI hides the Stop that would let it. Reproduction committed as a skipped test. **Pre-existing**: master carries byte-identical `Start_IsEnabled` and `Stop_IsEnabled`, so this PR neither causes nor fixes it.
- **#542: a transition body inherits the appending flow's ambient `SubjectTransaction`**, the same root cause as the startup scope defect fixed here, so property writes inside `StartAsync` can land in a caller's transaction. The old shared consumer loop was accidentally immune.
- **#543: `OpcUaSubjectClientSource` takes `_structureLock` and then writes into the lifecycle lock**, while its detach path re-enters that lock from inside it. Pre-existing, and it self-deadlocks a non-reentrant semaphore while holding the graph lock.
- **#544: status is written before the state a waiter then reads**, in both the reconcile and the start's catch, which made three tests race. Making status the last write would delete the class rather than the instances.
- **`SingleAttachmentHost.StartAsync` holds the attachment gate across an unbounded wait for the loaded root.** Server only: the interface default returns immediately and only `OpcUaServer` overrides it to await `RootLoaded`. `StopAsync` needs the same gate, and the UI Start operation passes `CancellationToken.None`, so a root that never loads pins the gate and Stop can never run. **Introduced here**: master's wrappers hold no semaphore at all, so a stop stayed reachable throughout. Two fixes, and choosing between them is a behaviour decision rather than a size one.
  - *Bound the wait*, about four lines in `OpcUaServer`. The file already defines `RootLoadWaitTimeout` at ten seconds and already applies it to this same wait on the synchronous `CreateInstance` path, so this makes the two agree and caps the gate hold. A root that legitimately takes longer then fails the start, where master would have waited.
  - *Move the wait outside the gate*, about ten lines in `SingleAttachmentHost`. Keeps master's semantics exactly, an unbounded wait with Stop reachable throughout, but `Status = Starting` is written inside the gate ahead of the wait today, so it reopens the question of what the UI shows while waiting.
- **No test project here sets a hard per-test timeout**, `HomeBlaze.OpcUa.Tests` included, so a deadlock in any of them reaches CI as an unattributed job timeout rather than a named failure. Worth having where a suite drives concurrency deliberately.
- **#575: remove the stop body's 50 ms delay.** The start's half is gone, replaced by the startup scope; the stop's remains a mitigation with no mechanism behind it, costing 50 ms per target on every stop that shutdown then waits out.
- **`OpcUaServer.WaitUntilStartableAsync` writes `Status = Stopped` outside the status lock and without bumping the stop generation.** Safe today, because it writes only that one value while the gate is held and the message is already null, so it can neither win over a stop nor produce a forbidden pair. Recorded because that safety is a property of the current value rather than something the discipline enforces.
- **#585: `AddSubject` can double start a hosted subject that later joins a hosting graph.** The self start the activation performs when the resolved context has no handler records nothing on the target, so a handler that later owns the subject starts it again, orphaning a `BackgroundService` loop past shutdown. Reproduced. **Master is worse**: its registration consults no handler at all, so it double starts whenever such a subject is in a hosting graph. Narrowed here to a misconfiguration and documented in `docs/hosting.md`; closing the remainder needs a second start owner, which is what this PR exists to remove.
- **#576: a cross handler ownership race can leave a subject in the graph with nothing running.** `DetachSubject` releases ownership after it appends its stops, so a *different* handler's take landing between the liveness clear and that release loses the compare and exchange, appends nothing, and is never retried. Needs two hosting enabled contexts over one subject plus concurrent graph mutation on both, which takes a manual `AddFallbackContext` to set up. Recorded in the design document as derived rather than measured: no seam sits between those two statements, so proving it would mean adding one to production code.
- **Cheaper attach and detach callbacks**, tracked as #456. Its figures came from a throwaway harness and read as a regression this PR introduced. They are at exactly the magnitude #540 shows this benchmark moving on its own, and the like-for-like run above finds branch and master byte-identical, so treat the like-for-like run as the answer on whether there is a regression and #456 as the question of whether the callbacks could be cheaper than they are.
- **`AddKeyedSubject<T>`** if a consumer needs several instances of one subject type.
- **`AddSubject` startup is linear in registered hosted types**, because the generic host starts hosted services sequentially. `HostOptions.ServicesStartConcurrently` is the mitigation and belongs to the application author.

## Notes for review

- `docs/design/hosting-service-ownership.md` records the decisions that are counter intuitive, each with its reason, and distinguishes what a test pins from what is supported only by argument. Its `What Has No Test Behind It` section is the honest inventory, including two labelled equivalent mutants.
- Eleven claims written before implementation turned out false against the code and were corrected rather than carried forward. The most consequential: `Interlocked.CompareExchange` ownership does not make the two context case benign, because the duplicate start comes from the *owning* handler receiving a second context attach.
- Five of the document's own "deleting this fails exactly these tests" claims were wrong, each written from reasoning rather than from running the mutation. Four were corrected against measurement. The fifth was caught by the final review, which found it contradicted by another passage in the same document rather than by a run.
- Race tests carry positive controls: each was verified by breaking the mechanism it guards and watching it fail. Where a test could pass with its mechanism deleted, it was rewritten rather than kept, which is how two vacuous guards and three two-core flakes were found.
- Master has been merged in twice. The first time was after #532 and #534 shipped the hosted startup scope. The scope survives; the machinery around it does not, since per-target chains need no readiness scan and no per-start cancellation source. Two behaviours changed rather than carried over: a detach no longer cancels a deferred start, it declines after the scope closes; and a service whose deferred start never ran is no longer stopped at shutdown, having never been created. The second merge brought the Sonar rollout, which conflicted with nothing here and left the suite green at 4,145, but which is why the benchmark above wants re-running.


.gitignore
```diff
@@ -386,3 +386,6 @@ src/HomeBlaze/HomeBlaze/PluginsCache/
 # Performance pass artifacts
 benchmark_*.md
 .claude/scheduled_tasks.lock
+
+# Subagent-driven development session scratch
+.superpowers/
```
RicoSuter: revert this file

src/Namotion.Interceptor.Hosting/SubjectServiceCollectionExtensions.cs
```diff
@@ -0,0 +1,114 @@
...
+        });
+
+        services.AddHostedService<SubjectActivation<T>>();
+
+        return services;
+    }
+
+    private static bool HasContextConstructor<T>()
+    {
+        return typeof(T).GetConstructors(BindingFlags.Public | BindingFlags.Instance)
```
RicoSuter: isnt there an activator api to check whether a ctor parameter exists?

docs/subject-guidelines.md
```diff
@@ -507,13 +528,18 @@ services.AddMySubject(subject =>
...
 {
-    // Context is automatically passed if:
-    // 1. Subject has this constructor parameter, AND
-    // 2. Context is registered in DI or provided via contextResolver
+    // No context parameter. AddSubject attaches the subject after construction,
+    // and runs its configure callback before that attach.
 }
 ```
+
+Declare an `IInterceptorSubjectContext` parameter only when the constructor genuinely needs the context, for example to build child subjects. The one thing worth knowing turns on whether construction actually attaches the subject, which is what the generated context constructor does: `configure` then runs against an attached subject, so its assignments are intercepted and race the start the attach queued. A constructor that declares the parameter and never attaches with it behaves like one that never declared it.
```
RicoSuter: turns on whether construction -> strange sentence

docs/subject-guidelines.md
```diff
@@ -439,20 +455,25 @@ public partial class HueBridge
...
 - **OpcUaSubjectServer** (`Namotion.Interceptor.OpcUa`): Injects OPC UA server configuration and telemetry services.
 
 ## Implementing Hosted Subjects for DI
 
 > See [Hosting](hosting.md) for foundational concepts on hosted subjects and the hosting lifecycle.
 
-When creating a subject library that extends `BackgroundService`, provide a DI extension method using `AddHostedSubject<T>` from `Namotion.Interceptor.Hosting`.
+When creating a subject library that extends `BackgroundService`, provide a DI extension method using `AddSubject<T>` from `Namotion.Interceptor.Hosting`.
+
+`AddSubject<T>` registers the subject as a singleton, constructs it at host start and attaches it to the context. When that context has hosting enabled, the handler on it starts the subject because the subject entered the graph, and host startup waits for that start. When the resolved context has no hosting handler, `AddSubject<T>` starts an `IHostedService` subject itself and stops that same instance at host shutdown, so the subject runs either way. Do not register the same subject with `AddHostedService<T>` as well, because that is a second owner and a second start.
```
RicoSuter: can there be only one subject of type T, is this a limitation we had before? so there can only by one ShellyDevice registered at the same time?

.gitignore
RicoSuter: Reverted. The entry was session scratch and does not belong to this change.

src/Namotion.Interceptor.Hosting/SubjectServiceCollectionExtensions.cs
RicoSuter: Yes, and it was worth changing. There is no Try form, so a check still means catching, but ActivatorUtilities.CreateFactory answers the precise question where the reflection answered a looser one: whether any public constructor mentions the type, which can be true of one ActivatorUtilities cannot call, in which case the check passed and the construction then threw. The decision and the construction are now the same code, and the constructor selection is cached. The probe runs once per registration.

docs/subject-guidelines.md
RicoSuter: Rewritten. It now states the ordering difference plainly: the generated context constructor attaches the subject itself, so configure runs against an attached subject, its assignments are intercepted and tracked, and they race the start that attach queued. Without such a constructor AddSubject attaches after configure, so the assignments are not intercepted and no start can observe them half written. A constructor that declares the parameter but never calls AddFallbackContext with it behaves like one that never declared it.

docs/subject-guidelines.md
RicoSuter: Yes, one instance per type, and it is not new: the registration is TryAddSingleton keyed on T, exactly as AddHostedSubject was. So AddShellyDevice() twice yields one ShellyDevice, and the second call quietly drops its configure and contextResolver. Now documented, with the workaround: to run several instances of one type, construct them and attach them to the object graph rather than registering them, which is what an application managing devices from configuration already does. HomeBlaze is unaffected because it builds devices through SubjectFactory rather than DI.

src/Namotion.Interceptor.Hosting/HostedServiceHandler.cs
```diff
@@ -7,160 +8,324 @@
...
-    public Task StartAsync(CancellationToken cancellationToken)
+    private void AttachSubject(IInterceptorSubject subject)
     {
-        if (_executeTask is not null)
+        if (_gate.State is HostedServiceGateState.Draining or HostedServiceGateState.Drained)
+        {
+            return;
+        }
+
+        _liveSubjects[subject] = 0;
```
RicoSuter, line 130: attach/detach callbacks are very hot path, can we somehow have a fast path here, ie dont do that much work for a subject which is host hosted or has not attachements (also in detach?) any options?

RicoSuter: Detach already has a fast path, at the `subjectTarget is null && attachments.IsEmpty` early return, and it exists for exactly this reason. Attach has none, and there is a specific thing blocking one.

`_liveSubjects[subject] = 0` cannot be conditional on the subject hosting something today. An attachment can be added at any time *after* a subject entered the graph, and that later `AttachHostedService` has no other way to tell a live subject from a detached one. Record only the subjects hosting something at attach time and every later attach onto every other subject is refused. The `TryRemove` in detach is the same fact from the other side: skip it for non-hosting subjects and the set stops tracking the graph.

So the write is the price of supporting attach-after-entry, not an oversight. Measured cost, from the performance review on this branch: about 161 bytes and 275 ns per subject attach, and 1.4 MB retained for a 20,000 subject graph that hosts nothing.

Three ways out, in order of how much I like them:

1. **Take the liveness answer from the registry.** `ISubjectRegistry.KnownSubjects` already knows whether a subject is in the graph, and the lookup is O(1). That deletes `_liveSubjects` entirely and both the write and the removal with it. The cost is a package dependency: `Namotion.Interceptor.Hosting` currently references only `Tracking`. Could be opportunistic, using the registry when the context has one and falling back to the set when it does not, at the price of two code paths through a guard that took several review rounds to get right.
2. **Accept it and document the shape.** It is one concurrent dictionary write per subject entering the graph, on graph mutation rather than on property writes, and the detach side already short circuits everything after it.
3. **Restrict attach to subjects not yet in the graph**, which removes the need for liveness altogether. I mention it only to rule it out: `OpcUaClient` attaches from inside its own `ExecuteAsync`, so it would break the one real consumer.

My suggestion is to measure it in HomeBlaze first rather than reason about it. If a 20,000 subject load is dominated by other work, option 2 stands; if this write shows up, option 1 is a contained follow-up. I would rather not reopen this guard in the current pull request, since it is the piece the reviews spent the most rounds on and it is now pinned by tests on both the attach and the detach side.

RicoSuter: Tracked as #456, with the measurements, the constraint that blocks the obvious fast path, and the options. Keeping this pull request on correctness and simplification.

RicoSuter: Correcting option 1 in my earlier reply: I claimed the registry lookup is O(1), and that is wrong for the call I named.

`ISubjectRegistry.KnownSubjects` returns a lazily rebuilt `ImmutableDictionary` snapshot, and every registry attach and detach nulls it (`SubjectRegistry.cs:217` and `:257`). A read straight after a mutation runs `_knownSubjects.ToImmutableDictionary()`. Reading it once per attach callback would therefore make a graph attach O(N squared) and allocate a fresh immutable dictionary per subject, far worse than the roughly 161 bytes it was meant to save. The usable call is `TryGetRegisteredSubject`, which is O(1) but takes the registry monitor on every attach, against a lock free dictionary write today.

Three more differences I had not accounted for:

- **Drain has no registry equivalent.** `_liveSubjects.Clear()` in `StopAsync` makes every subject not live for that handler, and the comment there states why: a handler that still reports live claims ownership of every attachment added afterwards, appends a start that no-ops, and never releases it. Subjects stay registered after the host stops, so each of the three liveness reads would need a separate gate check. One mechanism becomes two.
- **Liveness is per handler, a registry is per context.** Two contexts sharing one `IServiceCollection` is where the original double start came from, so that distinction carries weight.
- **It puts a second lock inside the chain lock.** `TryTakeOwnershipAndAppendAsync` reads liveness inside `lock (_sync)` and would then take `_knownSubjects` as well. The order looks consistent with the registry own use and I see no cycle, but it is another lock pair in the critical section that took the most review rounds.

So option 1 is a redesign rather than a swap, which weakens it against simply accepting the current shape. Measuring what enabling hosting costs on a graph with nothing hosted comes first either way, and that is the next step in #456.

RicoSuter: Measured it, and #456 now carries the numbers and rewritten options.

On a 20,000 subject graph that hosts nothing, enabling hosting costs +3.09 ms (+12.4%) and 40.0 B per subject on attach, and +1.74 ms (+18.0%) and zero extra bytes on detach. Three findings worth having here:

- The 161 B per subject I quoted earlier was an estimate and it was 4x too high. The real figure is exactly 40 B, one `ConcurrentDictionary` node, byte identical across every run.
- Having one more handler in the chain is not the cost. A lifecycle handler with an empty body is 15.5 ns per subject; `HostedServiceHandler`'s own body is roughly 9x that. Note also that it is an `ILifecycleHandler` only and never enters the property read or write chain, so property access is unaffected.
- Detach adds no allocation at all, so the existing early return is doing its job.

That changes which option is worth taking. The two `HostedServiceGate.State` reads per attach each take a monitor where a volatile int read would do, which is contained inside the gate, adds no dependency and does not touch the ownership guard. Taking liveness from the registry is now clearly the worst option rather than the best, for the reasons in the previous comment plus the fact that it would reclaim only those 40 B.

Nothing here changes this pull request.

RicoSuter: Correcting my own correction: I said the 161 B per subject figure was 4x too high. It was not. My harness ran warmup rounds, and a `ConcurrentDictionary` never shrinks its table on removal, so by the time I measured, `_liveSubjects` already had its buckets sized and I was measuring steady state only.

Both numbers are real and describe different moments:

- **First ever attach of the graph, which is what an application startup does: 164.2 B per subject.** Reproducible across three runs, and it matches an isolated `ConcurrentDictionary<object, byte>` test exactly at 164.3 B per entry into a fresh dictionary. On a 20,000 subject load that is about 3.3 MB, of which 0.8 MB is retained and about 2.5 MB is churn from table growth.
- **Every attach after that: 40.0 B per subject.** One node: a key reference, a next reference, a hashcode and the byte value, on top of the object header and method table pointer, padded to 40.
- **Detach: zero extra bytes either way.**

So the original review figure was right for the case that matters most, a host loading its graph once at startup. #456 now carries both numbers and the correction. The conclusion about which option to take does not change: the two gate reads are still the contained win, and taking liveness from the registry is still the worst option.

src/Namotion.Interceptor.Hosting/SubjectServiceCollectionExtensions.cs
```diff
@@ -0,0 +1,134 @@
...
+
+            // The constructor branch exists only because ActivatorUtilities throws when no constructor
+            // can consume the extra argument. It confers no attachment advantage: the generated
+            // constructor is "C(IInterceptorSubjectContext context) : this()", so the attach happens
+            // after the parameterless constructor body either way.
+            //
+            // The decision is the factory itself rather than a reflection query, so the code that
+            // decides a context taking constructor is usable is the code that then uses it. A
+            // reflection query answers the looser question of whether any constructor mentions the
+            // type, which can be true of one ActivatorUtilities cannot call.
```
RicoSuter: compact to two lines, dont need to explain .net apis

src/Namotion.Interceptor.Hosting/SubjectServiceCollectionExtensions.cs
```diff
@@ -0,0 +1,134 @@
...
+            var instance = context is not null && contextFactory is not null
+                ? (T)contextFactory(serviceProvider, [context])
+                : ActivatorUtilities.CreateInstance<T>(serviceProvider);
+
+            // Ordered ahead of the attach below, which is the only ordering this factory controls. For
+            // the generated constructor the subject is already attached and this changes nothing, but
+            // for the documented "MySubject(IInterceptorSubjectContext? context = null)" shape, which
+            // takes the context and never uses it, the attach below is the first one and running
+            // configure after it would start the subject half configured. The handler's start delay is
```
RicoSuter: compact this, too verbose

src/Namotion.Interceptor.Hosting/SubjectServiceCollectionExtensions.cs
```diff
@@ -0,0 +1,134 @@
...
+    /// <paramref name="configure"/> always runs before the attach this method performs, and the attach
+    /// is what makes a hosting enabled context start the subject. When that attach is the first one,
+    /// which is the case for a <typeparamref name="T"/> that has no context constructor and for one
+    /// that takes a context and ignores it, the subject is fully configured before anything can start
+    /// it and the assignments are neither intercepted nor tracked. When a generated context constructor
+    /// has already attached the subject, that attach cannot be reordered from here, so the assignments
+    /// are intercepted and tracked and they race the start it appended, exactly as they do for a hand
+    /// written <c>new MySubject(context) { Name = "x" }</c>. The three shapes are set out side by side
+    /// in docs/hosting.md.
+    /// </para>
```
RicoSuter: compact this xml docs block

src/Namotion.Interceptor.Hosting/HostedServiceTarget.cs
```diff
@@ -0,0 +1,236 @@
+using Microsoft.Extensions.Hosting;
+
+namespace Namotion.Interceptor.Hosting;
+
```
RicoSuter, line 4: compact the xml docs and doc comments here, only keep what is needed by consumer (xml docs) and not clear from code (code docs, just the why)

src/Namotion.Interceptor.Hosting/InterceptorHostingExtensions.cs
```diff
@@ -1,171 +1,261 @@
 using System.Collections.Immutable;
+using System.Runtime.ExceptionServices;
 using Microsoft.Extensions.Hosting;
 
 namespace Namotion.Interceptor.Hosting;
```
RicoSuter, line 6: also here compact the docs

docs/hosting.md
```diff
@@ -117,39 +140,192 @@ public class PersonBackgroundService : BackgroundService
...
 
 `SourceMonitor` is the one implementation in this repository. It is what makes an attached source count towards source registration from the moment it is attached rather than from the moment it finally starts, so a synchronization wait cannot complete against a tree whose sources have not registered yet. See [Applications That Create Sources at Runtime](connectors-monitoring.md#applications-that-create-sources-at-runtime).
 
+A deferrer runs inside the lifecycle lock, so neither `DeferCompletion` nor the hold's `Dispose` may block on anything that can be waiting for that lock. A lock of the deferrer's own is allowed, but only where the acquisition order is already fixed: it is safe when nothing held under that lock ever waits on anything that needs the lifecycle lock, so the two are always taken in the same order, and unsafe as soon as they can be taken in either. The constraint is written on `IStartupCompletionDeferrer`, and an implementation that follows it cannot take part in the deadlock: see [A deferrer that takes a lock of its own](design/hosting-service-ownership.md#4-a-deferrer-that-takes-a-lock-of-its-own).
+
+## Migrating from the Previous API
```
RicoSuter: remove migration chapter, noone has to migrate outside (here already done)

src/Namotion.Interceptor.Hosting/SubjectServiceCollectionExtensions.cs
RicoSuter: Done. The ActivatorUtilities explanation is gone; what is left is only the part that is not a .NET fact, that the factory is the decision rather than a reflection query, because reflection answers the looser question of whether a constructor mentions the type.

src/Namotion.Interceptor.Hosting/SubjectServiceCollectionExtensions.cs
RicoSuter: Cut to three lines: before the attach, because for a shape whose first attach is that one, configuring afterwards would let a handler start the subject half configured. The walk through the three constructor shapes now lives only in docs/hosting.md.

src/Namotion.Interceptor.Hosting/SubjectServiceCollectionExtensions.cs
RicoSuter: Cut from 19 lines to 6. Kept the two facts a consumer cannot get from the signature: one instance per type and a second call throws, and that configure runs before the attach so the shapes differ in whether the assignments race the start. The side by side comparison stays in docs/hosting.md.

src/Namotion.Interceptor.Hosting/HostedServiceTarget.cs
RicoSuter: Done, 124 comment lines down to 71. Removed what the code says, for instance "set under the chain lock" above a lock statement, and the .NET facts, for instance that CompareExchange carries its own fence. Kept the concurrency reasoning, since none of it is derivable from the code and two of those arguments exist because a review found a real defect when they were missing.

src/Namotion.Interceptor.Hosting/InterceptorHostingExtensions.cs
RicoSuter: Done, 86 down to 72. Same rule. I also applied it to HostedServiceHandler.cs, which had no thread but was the worst of them at 262 comment lines, now 214. Across the four files it is 107 lines with no change to anything claimed. Say the word if you wanted the handler left alone.

docs/hosting.md
RicoSuter: Removed, along with the last comparison to the old registration helper in docs/subject-guidelines.md. Checked before deleting that the chapter's one piece of standalone advice, not to translate an instance attach into a factory that captures it, is already covered by "The factory must construct", which is where a reader with that problem lands. Nothing linked to the section.

src/Namotion.Interceptor.Connectors/Monitoring/SourceMonitor.cs
```diff
@@ -258,7 +258,16 @@ public IDisposable DeferWaitCompletion()
     }
 
     /// <inheritdoc />
-    /// <remarks>Explicit, so <see cref="DeferWaitCompletion"/> stays this type's only surface.</remarks>
```
RicoSuter, line 261: do we need this remarks or is it clear from design docs and code? why is the IStartupCompletionDeferrer.DeferCompletion not called DeferWaitCompletion or vice-versa?

src/Namotion.Interceptor.Hosting/HostedServiceAttachment.cs
```diff
@@ -0,0 +1,93 @@
...
+    /// anything in between widens the window it exists to close.
+    /// </remarks>
+    HostedServiceAttachmentState GetState(out IHostedService? current);
+
+    /// <summary>
+    /// What the attachment is doing right now. Separates a start in flight, a stop in flight and a
+    /// terminal attachment from a settled one with nothing running, which <see cref="Current"/> and
+    /// <see cref="Fault"/> reading null cannot.
+    /// </summary>
+    HostedServiceAttachmentState State { get; }
```
RicoSuter: should we maybe remove this property when GetState already provides a safer way?

src/Namotion.Interceptor.Hosting/HostedServiceAttachment.cs
```diff
@@ -0,0 +1,93 @@
...
+}
+
+/// <inheritdoc />
+public interface IHostedServiceAttachment<T> : IHostedServiceAttachment
```
RicoSuter, line 34: where is the generic variant needed/exposed?

src/Namotion.Interceptor.Hosting/HostedServiceGate.cs
```diff
@@ -0,0 +1,96 @@
...
+}
+
+/// <summary>
+/// Startup and shutdown gate for hosted service transitions. The state only ever moves forward:
+/// NotStarted to Running to Draining to Drained, or NotStarted straight to Draining when a host is
+/// stopped without having started.
+/// </summary>
+internal sealed class HostedServiceGate
+{
+    private readonly object _sync = new();
```
RicoSuter: use Lock and rename to _lock?

src/Namotion.Interceptor.Hosting/HostedServiceGate.cs
```diff
@@ -0,0 +1,96 @@
...
+        }
+
+        _opened.TrySetResult();
+    }
+
+    /// <summary>
+    /// Completes once the gate has left <see cref="HostedServiceGateState.NotStarted"/>. Callers must
+    /// then read <see cref="State"/> and decide what to do; the wait itself carries no verdict.
+    /// </summary>
+    public Task WaitForOpenAsync() => _opened.Task;
```
RicoSuter, line 90: no cancellation token?

src/Namotion.Interceptor.Hosting/HostedServiceGate.cs
```diff
@@ -0,0 +1,96 @@
...
+
+    /// <summary>
+    /// Completes once the drain has begun. Lets a transition parked on something a caller controls,
+    /// which is a startup scope and nothing else, be released by shutdown instead of holding the
+    /// drain's barrier for the whole shutdown deadline. Continuations run asynchronously, so what was
+    /// parked cannot resume on the thread that begins the drain.
+    /// </summary>
+    public Task WaitForDrainingAsync() => _draining.Task;
```
RicoSuter, line 98: no cancellation token?

src/Namotion.Interceptor.Hosting/HostedServiceHandler.cs
```diff
@@ -8,240 +9,374 @@
 namespace Namotion.Interceptor.Hosting;
 
 [RunsAfter(typeof(ContextInheritanceHandler))]
-internal class HostedServiceHandler : IHostedService, ILifecycleHandler, IDisposable
+internal sealed class HostedServiceHandler : IHostedService, ILifecycleHandler
 {
-    private ILogger? _logger;
+    // A workaround, not a design choice: docs/design/hosting-service-ownership.md#the-50-ms-delay.
+    private const int TransitionDelayMilliseconds = 50;
```
RicoSuter, line 15: do we still need this? should we fixed in a follow up issue/pr?

src/Namotion.Interceptor.Tracking/IStartupCompletionDeferrer.cs
```diff
@@ -13,11 +13,39 @@ namespace Namotion.Interceptor.Tracking;
 /// Take a hold before queueing the work and dispose it once the work has run, including on failure.
 /// Holds are counted. Taking one never un-completes a signal that has already fired.
 /// </para>
+/// <para>
+/// <b>The constraint on an implementation.</b> Both halves can run under the lifecycle lock:
```
RicoSuter: can we compact this to a simple contract description?

src/Namotion.Interceptor.Hosting/SubjectActivation.cs
```diff
@@ -0,0 +1,56 @@
...
+    public SubjectActivation(IServiceProvider serviceProvider)
+    {
+        _serviceProvider = serviceProvider;
+    }
+
+    public async Task StartAsync(CancellationToken cancellationToken)
+    {
+        // Resolving constructs and attaches the subject, which makes the handler append its start.
+        var subject = _serviceProvider.GetRequiredService<T>();
+
```
RicoSuter:
```suggestion
```

src/Namotion.Interceptor.Hosting/HostedServiceTarget.cs
```diff
@@ -0,0 +1,430 @@
...
+/// <summary>
+/// One managed thing: a subject implementing <see cref="IHostedService"/>, or a factory attachment.
+/// Owns a serialized transition chain, so transitions for one target never interleave while
+/// transitions for unrelated targets run concurrently.
+/// </summary>
+internal sealed class HostedServiceTarget
+{
+    private readonly object _sync = new();
```
RicoSuter: use Lock and rename to  _xyzLock?

src/Namotion.Interceptor.Hosting/HostedServiceTarget.cs
```diff
@@ -0,0 +1,430 @@
...
+    /// landing between an install and its record nulls the owner, finds no record to retire, and leaves
+    /// one that no later release can ever match.
+    /// </summary>
+    /// <remarks>
+    /// Its own lock rather than <c>_sync</c>. A context detach releases every attachment target it
+    /// enumerates, including ones whose chain lock a concurrent attach is holding, so releasing under
+    /// the chain lock deadlocks that pair. Nothing held here ever takes another lock, and the take
+    /// enters it while already holding <c>_sync</c>, so the one order is chain lock then this.
+    /// </remarks>
+    private readonly object _ownershipSync = new();
```
RicoSuter: use Lock and rename to _ownershipLock?

src/Namotion.Interceptor.Hosting/HostedServiceHandler.cs
```diff
@@ -8,240 +9,374 @@
...
 
-    private async Task ExecuteAsync(CancellationToken stoppingToken)
+    private void DetachSubject(IInterceptorSubject subject)
     {
-        // The long-lived action loop must not inherit the scope in which the host was started.
-        _startupScope.Value = null;
-        _logger ??= _loggerResolver();
+        // Read before anything is allocated: a completion source per subject is 1.76 MB of garbage per
+        // detach of a 20,000 subject graph, under the lifecycle lock. "Has ever hosted", not "hosts
+        // now", for the reason on TryGetHostedServiceAttachments.
```
RicoSuter: do we need to keep this 3 line comment?

src/Namotion.Interceptor.Hosting/HostedServiceHandler.cs
```diff
@@ -8,240 +9,374 @@
...
+
+        var attachments = subject.GetHostedServiceAttachments();
+        if (subjectTarget is null && attachments.IsEmpty)
+        {
+            // Liveness is recorded only for a subject that hosts something: every reader of it holds
+            // a target. MarkLiveIfAttached covers the one moment that is not true.
+            return;
+        }
+
+        _liveSubjects[subject] = 0;
```
RicoSuter, line 115: cant two attaches on different context of the same subject (assign to one, detach and attach to another context) race here?

src/Namotion.Interceptor.Hosting/HostedServiceTarget.cs
```diff
@@ -0,0 +1,430 @@
...
+    /// outranks <see cref="HostedServiceAttachmentState.Faulted"/> because the mark is terminal for the
+    /// attachment while the fault is not: every start appended after it is refused, and the one start
+    /// that can still run, one queued ahead of the mark, clears the fault on its way past. Ranked the
+    /// other way a target nothing can attach to again would read as the recoverable one of the two.
+    /// <see cref="HostedServiceAttachmentState.Stopping"/> outranks
+    /// <see cref="HostedServiceAttachmentState.Removed"/> because an explicit detach marks detached and
+    /// then appends its stop, so both hold while that stop runs and "still shutting down" is the more
+    /// urgent of the two; it settles into <see cref="HostedServiceAttachmentState.Removed"/>.
+    /// </remarks>
+    public HostedServiceAttachmentState State => GetState(out _);
```
RicoSuter: as questioned somewhere else, remove and only expose getstate()?

src/Namotion.Interceptor.Hosting/HostedServiceGate.cs
RicoSuter: Done, and it was the right call rather than only style: these three were the only `readonly object` locks left in the package, against 34 `Lock` uses elsewhere in the repository including Tracking and Registry.

Renamed to `_lock`.

src/Namotion.Interceptor.Hosting/HostedServiceTarget.cs
RicoSuter: Done, and it was the right call rather than only style: these three were the only `readonly object` locks left in the package, against 34 `Lock` uses elsewhere in the repository including Tracking and Registry.

Renamed to `_chainLock`.

src/Namotion.Interceptor.Hosting/HostedServiceTarget.cs
RicoSuter: Done, and it was the right call rather than only style: these three were the only `readonly object` locks left in the package, against 34 `Lock` uses elsewhere in the repository including Tracking and Registry.

Renamed to `_ownershipLock`. The rename also drops the `_sync` / `_ownershipSync` pairing, which read as two halves of one thing when the lock order between them is load bearing.

src/Namotion.Interceptor.Hosting/HostedServiceGate.cs
RicoSuter: Deliberate, and now recorded in the code. `BeginDraining` completes both sources unconditionally, so nothing parked on either outlives the start of shutdown. A token would be wrong rather than merely unnecessary: both are awaited inside transition bodies, where a caller's cancellation must never abort a start that is already creating an instance.

src/Namotion.Interceptor.Hosting/HostedServiceGate.cs
RicoSuter: Deliberate, and now recorded in the code. `BeginDraining` completes both sources unconditionally, so nothing parked on either outlives the start of shutdown. A token would be wrong rather than merely unnecessary: both are awaited inside transition bodies, where a caller's cancellation must never abort a start that is already creating an instance.

src/Namotion.Interceptor.Hosting/HostedServiceAttachment.cs
RicoSuter: Removed. `GetState` is the only reading now, so the pair cannot be split by construction. It cost nothing: `HostedServiceAttachmentState` is new on this branch so master is unaffected, there were no production consumers of `State` since `SingleAttachmentHost` already used `GetState` at both call sites, and the ten test assertions moved to `GetState(out _)`. Public API snapshot accepted.

src/Namotion.Interceptor.Hosting/HostedServiceTarget.cs
RicoSuter: Removed. `GetState` is the only reading now, so the pair cannot be split by construction. It cost nothing: `HostedServiceAttachmentState` is new on this branch so master is unaffected, there were no production consumers of `State` since `SingleAttachmentHost` already used `GetState` at both call sites, and the ten test assertions moved to `GetState(out _)`. Public API snapshot accepted.

Same change as the thread on `IHostedServiceAttachment`.

src/Namotion.Interceptor.Hosting/HostedServiceAttachment.cs
RicoSuter: Three places: it is the return type of `AttachHostedService<T>` and `AttachHostedServiceAsync<T>`, and `SingleAttachmentHost` holds an `IHostedServiceAttachment<TService>` so `Current` comes back typed without a cast.

Worth knowing that the typed `GetState(out T? current)` is what forced dropping `out T` covariance, since C# will not take a covariant parameter in an `out` position. Keeping the generic form is what buys the consumer the cast free read, so I left it as it is.

src/Namotion.Interceptor.Hosting/SubjectActivation.cs
RicoSuter: Removed.

src/Namotion.Interceptor.Hosting/HostedServiceHandler.cs
RicoSuter: Cut to two lines. Both facts earn their place: the measured allocation is why the read sits before anything is allocated, and "has ever hosted" rather than "hosts now" is a correctness condition rather than an optimisation.

src/Namotion.Interceptor.Tracking/IStartupCompletionDeferrer.cs
RicoSuter: Compacted from 28 lines to 9. What is left is the rule an implementer has to follow; the derivation, and the cycle it avoids, stay in the design document it already links. The duplicate remark on `DeferCompletion` itself is gone, since it restated the type level one.

src/Namotion.Interceptor.Connectors/Monitoring/SourceMonitor.cs
RicoSuter: Cut to one sentence. The paragraph this branch added restated the interface's rule; what remains says the constraint holds here and why, with the derivation in the interface remarks.

On the naming: the mismatch predates this branch, both names are on master unchanged. `DeferCompletion` is the interface's vocabulary in Tracking, which knows nothing about waits, and `DeferWaitCompletion` is this type's own. The honest observation is that the explicit implementation exists only to keep two names alive, so unifying them would delete this member and its remarks together. That is a public rename across two packages, so it is out of scope here.

src/Namotion.Interceptor.Hosting/HostedServiceHandler.cs
RicoSuter: Filed as #575. The start body already lost its half of this to the startup scope, which answers the hazard rather than guessing at a duration. What remains on the stop side is a mitigation with no mechanism behind it, costing 50 ms per target on every stop and passing `CancellationToken.None`, so shutdown waits it out.

Not done here because it changes shutdown timing for every connector running on this handler, which wants its own measurement pass and chaos run.

src/Namotion.Interceptor.Hosting/HostedServiceHandler.cs
RicoSuter: Yes, across two handlers, though not for the sequence you describe.

**Sequentially it is safe.** `DetachSubject` releases ownership before it returns, so the attach behind it finds a null owner and wins the exchange. Re-attaching to the same context is safe for a different reason: `TryTakeOwnership` treats an owner that is already this handler as success, so the ordering against its own release does not matter.

**Concurrently and across two different handlers there is a window.** `DetachSubject` clears liveness, appends its stops, and releases last. A second handler's compare and exchange landing between the clear and the release finds the first handler still installed, so the take fails, nothing is appended, and nothing retries it. The first handler then nulls the owner, leaving the subject live for the second handler with the target owned by nobody, nothing running and no error recorded anywhere.

It is the same failure mode the draining handler guard in `TryTakeOwnershipAndStart` exists to prevent, and it is reachable only with two hosting enabled contexts over one subject plus concurrent graph mutation on both, which takes a manual `AddFallbackContext` to set up.

Recorded in the design document under Liveness and filed as #576 rather than fixed here. Two reasons. No seam sits between the liveness clear and the release, so proving it would mean adding one to production code, which is why the document records it as derived rather than measured. And a fix is new design rather than a reordering, because the handler that lost the exchange has to learn that ownership became free; releasing before appending the stops is not available, since that was measured and reopens a defect where a second handler's start lands ahead of the first handler's stop on a shared chain.

src/Namotion.Interceptor.Connectors/Monitoring/SourceMonitor.cs
```diff
@@ -258,7 +258,11 @@ public IDisposable DeferWaitCompletion()
...
+    /// <remarks>
+    /// Explicit, so <see cref="DeferWaitCompletion"/> stays this type's only surface. The interface's
+    /// locking constraint holds: the take acquires nothing, and the release takes _lock in the order this
+    /// type already establishes for itself through <see cref="HandleLifecycleChange"/>.
+    /// </remarks>
```
RicoSuter, line 265: why not name the method here DeferWaitCompletion as well?

PR #455, opened 2026-08-12
Scheduled per-property change subscriptions

RicoSuter:
Adds scheduled delivery for per-property change subscriptions, renames the inline entry points so their hazard is visible at the call site, and fixes the transaction defects this feature makes ordinary.

## Why

Per-property subscriptions deliver on the writing thread, inside the write, before the setter returns. That makes them concurrent, unblockable, and unable to throw: an observer exception propagates out of the setter and suppresses later deliveries for that write.

Those constraints are right for the channel and wrong for a large class of consumers. An observer that pushes a value to a device, writes a row, or updates a UI cannot honour them, so today it has to build its own bridge to a scheduler. Doing that correctly is harder than it looks. Composing `ObserveOn` dedicates a private thread to every subscription that signals, which is unaffordable at one subscription per property, and an exception reaching an `ObserveOn` sink escapes the scheduler work item, which on `Scheduler.Default` terminates the process and on `TaskPoolScheduler` silently ends the stream.

## Added

**Scheduled delivery.** `Subscribe(callback, scheduler, onError)` and the matching `SubscribeToProperty` forms queue the change on the writing thread and drain it in bounded batches on a caller-supplied `IScheduler`. Delivery is serialized per subscription, and observer, `onError` and scheduler exceptions are reported rather than propagated into the write. They return a `ScheduledPropertySubscription` exposing `PendingCount` and `IsFaulted`.

This is a hand-rolled dispatcher rather than `ObserveOn`. Owning the dispatch buys three things `ObserveOn` cannot: failure handling instead of a permanent silent wedge, two struct copies instead of four, and a handle that is not a live Rx sink castable to `IObserver<T>`. It is not an allocation win, and the cost table below says so.

**A benchmark.** A write-side case covering the channels, kept as a regression guard.

## Changed

**The inline entry points are renamed** to `SubscribeInline` and `SubscribeToPropertyInline`, so `Subscribe` and `SubscribeToProperty` now mean the safe scheduled form. Previously the most hazardous entry point in the library had the shortest and most inviting name, and the safe path was distinguished only by an extra argument.

The rule behind it: if delivery happens on the writing thread, the method name says so. The corollary is that safety is never controlled by an argument value, which is why the scheduled overloads reject inline schedulers rather than quietly meaning something else. Passing `ImmediateScheduler` there would not deliver your change on your thread, it would make whichever writer wins the counter race drain every other writer's changes inside its own setter.

**`IPropertyChangeObserver`'s contract is now conditional.** It still must never throw on the inline path, while the scheduled path reports to `onError` and keeps the subscription live. The XML docs state both cases, since that is what appears in IntelliSense while implementing the interface. No existing implementation needs to change.

**Documentation** gained an API routing table, a delivery-guarantees table and a channel-cost table with measured figures, and the surrounding prose was cut back so the added material does not grow the page.

## Fixed

Transaction defects that are pre-existing but that this feature makes ordinary. A thread-owning scheduler such as `EventLoopScheduler` starts its worker thread on first use, `Thread.Start` flows the starting flow's `ExecutionContext`, and Rx never restores it per work item, so a scheduler first used inside a transaction hands its thread a frozen ambient transaction for life. This subscription's own scheduling is suppressed, so it cannot create such a thread; a caller sharing the scheduler with their own work can.

- **Cross-transaction corruption.** `Dispose` returned its pending-changes dictionary to a static pool without nulling the field, so a flow holding a disposed transaction wrote into and read from a dictionary another live transaction already owned, under a different lock. `Dispose` now nulls the field under the lock and returns the buffer afterwards, so the aliasing is structurally impossible rather than guarded at each site. In a probe of 300 dispose-versus-writer races, master lost 300 writes and this branch lost none.
- **An exception out of a property setter.** Capture threw when it lost a race with a cross-flow dispose. It now falls through to the real write with the speculative origin finalization rolled back, because an exception escaping a setter on a scheduler thread ends the process.
- **A permanently wedged flow.** The nested-transaction check read `CurrentTransaction.Value != null`, so a thread born inside a transaction threw "Nested transactions are not supported" on every `BeginTransactionAsync` for the rest of its life once that transaction was disposed elsewhere.
- **Foreign pending changes.** `GetPendingChanges` returned the changes of whichever transaction last owned the pooled dictionary.
- **An exception for an ordinary write.** The context-binding check resolved the interceptor with `TryGetService`, which throws when a context resolves more than one, and context inheritance leaves exactly that. It now tests membership, which is the question actually being asked.
- **Derived cascades stranded on a frozen flow.** Suppression is now keyed on whether the ambient transaction can still serve a getter from its pending buffer, so a transaction already torn down stops suppressing and the cascade runs. Keying on the disposed flag alone would not do, because dispose sets that flag before releasing the buffer.
- **Write timestamps for writes that never happened.** A write dropped by an interceptor still advanced `TryGetWriteTimestamp()`, which feeds OPC UA and MQTT source timestamps and the read-after-write staleness check that discards a fresh device read when the stored timestamp is not older.

A dedicated review of this area compared branch against master under identical probes and found no input, configuration or interleaving where the branch is worse. Under three seconds of mixed transactional stress master produced `NullReferenceException` ×9,248 plus collection-corruption errors, and the branch produced only documented conflict exceptions.

## Contract points worth reading

- Serialization is **per subscription, not per observer**. One observer shared across subscriptions is invoked concurrently and must synchronize itself.
- Dormancy is **not symmetric with disposal**. Detaching the subject stops acceptance but not the drain, so a change accepted before the detach is still delivered. Disposal drops what is queued.
- The queue is **unbounded, and a drained backlog is not returned**. `ConcurrentQueue` keeps the largest segment it ever allocated, so a subscription costs its peak backlog from then on until disposal. Watch `PendingCount`, and recycle a subscription that has spiked rather than expecting it to shrink.
- **A faulted subscription is silent.** Without `onError` there is no signal, and `PendingCount` reads zero, which is indistinguishable from healthy and idle. Pass `onError`, or poll `IsFaulted`.
- **A blocking observer needs a scheduler of its own.** The drain owns its scheduler thread for the whole call, so on a single-threaded scheduler shared with other subscriptions it starves them. `Scheduler.Default` grows its pool instead.
- Ambient `AsyncLocal` state, `Activity.Current` and logger scopes do not flow to observers. Suppression covers the work item, not the context a scheduler's worker thread was born with, and a thread-owning scheduler starts its thread on first use. Scheduling from the subscription is suppressed, so such a thread is born clean; the hazard is scheduling your own work on the same scheduler for the first time inside a transaction scope.

## Public API

```csharp
// Inline: delivered on the writing thread, inside the write, before the setter returns.
IDisposable SubscribeInline(this PropertyReference property, IPropertyChangeObserver observer);
IDisposable SubscribeInline(this PropertyReference property, PropertyChangeCallback callback);

IDisposable SubscribeToPropertyInline<TSubject, TValue>(this TSubject subject,
    Expression<Func<TSubject, TValue>> propertySelector, IPropertyChangeObserver observer)
    where TSubject : IInterceptorSubject;
IDisposable SubscribeToPropertyInline<TSubject, TValue>(this TSubject subject,
    Expression<Func<TSubject, TValue>> propertySelector, PropertyChangeCallback callback)
    where TSubject : IInterceptorSubject;

// Scheduled: serialized per subscription, isolated from the write.
ScheduledPropertySubscription Subscribe(this PropertyReference property,
    IPropertyChangeObserver observer, IScheduler scheduler, Action<Exception>? onError = null);
ScheduledPropertySubscription Subscribe(this PropertyReference property,
    PropertyChangeCallback callback, IScheduler scheduler, Action<Exception>? onError = null);

ScheduledPropertySubscription SubscribeToProperty<TSubject, TValue>(this TSubject subject,
    Expression<Func<TSubject, TValue>> propertySelector,
    IPropertyChangeObserver observer, IScheduler scheduler, Action<Exception>? onError = null)
    where TSubject : IInterceptorSubject;
ScheduledPropertySubscription SubscribeToProperty<TSubject, TValue>(this TSubject subject,
    Expression<Func<TSubject, TValue>> propertySelector,
    PropertyChangeCallback callback, IScheduler scheduler, Action<Exception>? onError = null)
    where TSubject : IInterceptorSubject;

public sealed class ScheduledPropertySubscription : IDisposable
{
    public int PendingCount { get; }
    public bool IsFaulted { get; }
    public void Dispose();
}
```

`scheduler` is required with no default, which keeps the scheduled overload set at arity 4 and 5 disjoint from the inline set at 2 and 3. No package reference is added.

## Breaking changes

Four methods are renamed. Nothing is removed.

| Before | After |
|---|---|
| `property.Subscribe(observer)` | `property.SubscribeInline(observer)` |
| `property.Subscribe(callback)` | `property.SubscribeInline(callback)` |
| `subject.SubscribeToProperty(selector, observer)` | `subject.SubscribeToPropertyInline(selector, observer)` |
| `subject.SubscribeToProperty(selector, callback)` | `subject.SubscribeToPropertyInline(selector, callback)` |

Migration is mechanical and the compiler finds every site, since the scheduled overloads keep the old names and differ by arity. No production code in this repository called the old names. This is a hard rename with no `[Obsolete]` forwarders, deliberately: a forwarder would preserve the inviting name on the dangerous path, which is the thing being removed.

Worth a release-note callout: mechanically fixing the compile error by adding a scheduler switches four behaviours at once, since the write no longer waits for the observer, exceptions are reported instead of propagated, the queue is unbounded, and disposal drops what is queued.

One behaviour change beyond the rename: a write on a flow whose ambient transaction has been disposed now reaches the model and recalculates its dependents, where before it was captured into a pooled dictionary and lost or misattributed.

## Change size

| Dimension | Files | Added | Removed |
|---|---:|---:|---:|
| Production code | 8 | 743 | 61 |
| Tests | 17 | 1820 | 65 |
| Benchmarks | 1 | 128 | 2 |
| Public API snapshot | 1 | 16 | 4 |
| Documentation | 3 | 92 | 35 |
| **Total** | **30** | **2799** | **167** |

Production code is essentially one new file, the 367-line dispatcher, plus 175 lines of extension methods of which a large share is XML documentation. The transaction area is 141 added lines across two files.

## Performance

Measured on one Apple M4 Max running .NET 9.0.10 arm64. Allocation figures were bit-identical across runs and are quoted as bytes; timings could not be pinned on this hardware and appear only as factors between rows from the same run.

| Channel | Per write | Per delivery | Held per live subscription | Write time, inline = 1 |
|---|---|---|---|---|
| `GetPropertyChangeObservable()` | none | none | about 5,712 bytes per additional subscriber, plus one dedicated dispatch thread each | about 2.2 |
| `CreatePropertyChangeQueueSubscription()` | none | none | about 5,536 bytes per additional subscription | 1.9 |
| `SubscribeInline` | none | in the write | about 172 bytes | 1.0, the reference |
| `Subscribe` with a scheduler | 17 to 26 bytes | 160 bytes keeping up, none under backlog | about 5,607 bytes | 2.6 |

These compare one identical write, not equal workloads. A context-level channel delivers every property's changes, so watching one property out of five hundred means paying for the other 499 as well, which is why 2.6 is usually the cheaper choice despite being the highest number in the column.

Setup dominates rather than steady state: almost all of a scheduled subscription's footprint is its empty `ConcurrentQueue`, about 5,376 bytes, so a thousand of them cost roughly 5.35 MB against 0.16 MB for a thousand inline ones. On a wide model that, not per-write cost, is what decides the channel.

The transaction changes cost nothing where no transaction is open, and about 1.9% on the capture path, roughly 1.8 nanoseconds, from two volatile reads and a membership test replacing a length switch.

## Testing

47 new tests. Coverage includes the counter settling to zero with an empty queue after quiescence, the batch handoff and its snapshot refresh, concurrent writers never re-entering one subscription, dispose racing enqueue and racing a mid-flight delivery, both scheduler-failure shapes, ambient state not flowing including during a transaction commit, dormancy in both directions, thread economy, and the transaction and derived-property cases above.

The dispatcher went through a dedicated adversarial gate and several independent reviews, which between them ran more than a hundred mutants plus concurrency probes on arm64. Reviews also drove three defects out of the branch: an unfenced upstream publication that would have orphaned subscriptions on x86-64 while being accidentally safe on arm64, a drain that snapshotted its counter once and so doubled its scheduler work items, and `IsFaulted` reading false inside the `onError` handling the fault.

## Not in this PR

`GetInlineChangeObservable`, a single-property `IObservable`, was developed and then dropped. Serializing an inline sequence means holding a lock across the handler while on the writing thread, and that lock joins a cycle with the exclusive transaction lock: a writer holding the gate blocks on the semaphore while a commit holding the semaphore blocks in its apply pass on the gate. Reproduced at defaults with a single subscriber whose handler writes no property and takes no lock of its own. Rx's `Synchronize` has the same shape, so leaving serialization to the consumer moves the deadlock rather than removing it. It returns when serialization has a design that holds nothing across user code.

The instruments that produced the cost table were run once and then removed, since the benchmark workflow runs every benchmark on every pull request twice. Only the write-side guard is carried.

Deferred, each additive: a bounded queue with a drop policy, demand-driven conflation, current value at subscribe time, async serialized handlers, and per-change ambient context.

## Follow-ups

- #467, derived cascades dropped for the three kinds of write a transaction never captures. Records why the two obvious fixes fail.
- #466, whether two transaction interceptors resolvable from one subject should be rejected at registration. The crash is fixed here.
- #461, creating the subscription queue lazily, which measured a 23.6x reduction in retained memory for idle subscriptions but wants profiling first.
- #462, `ChangeQueueProcessor` dropping its buffered changes on shutdown without counting them.

`GetPropertyChangeObservable(scheduler)` flows the writer's `ExecutionContext` to subscribers for the same reason and with the same consequences as the paths fixed here. It is released API, so changing it needs its own decision.


src/Namotion.Interceptor.Tracking.Tests/Change/ScheduledPropertySubscriptionProtocolTests.cs
```diff
@@ -0,0 +1,607 @@
...
+
+        using var subscription = new PropertyReference(person, nameof(Person.FirstName))
+            .Subscribe(
+                (in SubjectPropertyChange change) =>
+                {
+                    received.Add(change.GetNewValue<string?>());
+                    if (received.Count == 1)
+                    {
+                        // Written once only, so the drain terminates instead of feeding itself forever.
+                        person.FirstName = "two";
```
RicoSuter, line 111: is this test really working as expected because we're using equality interceptor, or is this not active?

RicoSuter: The equality interceptor is not active here. `WithPropertyChangeSubscriptions()` registers only `PropertyChangeInterceptor`; the equality check comes from `WithEqualityCheck()`, which this context does not call. So every write raises a change, including a write of the same value.

The test does not depend on that either way, since `"one"` and `"two"` differ and would both pass an equality check. Verified by switching all 20 protocol tests to `WithFullPropertyTracking()`: 20/20 still pass.

What the absence of the equality check does change is the `received.Count == 1` guard, which is load-bearing rather than defensive. Without it the observer would write on every delivery and, with nothing suppressing equal values, the drain would feed itself forever.

The assertion discriminates: removing the snapshot refresh in `Drain` fails exactly this test and no other.

RicoSuter:
Superseded by #468 and #469. The transaction correctness work and scheduled-subscription feature were split into those two focused PRs, fully reviewed, and merged in order. The replacement PRs also include the subsequent correctness and documentation fixes.

PR #458, opened 2026-08-14
refactor(tracking): reconcile ordered container relationships

RicoSuter:
## Summary

Replace mutable child-index refresh with one staged, immutable relationship reconciliation model shared by lifecycle tracking, Registry, and parent tracking.

- Preserve one relationship per source occurrence in exact collection or dictionary enumeration order.
- Treat dictionary keys as opaque metadata. Reconciliation never calls key equality or hashing.
- Atomically replace complete relationship groups in Registry and `WithParents()` views.
- Keep membership and reference counting distinct from occurrence metadata: duplicate occurrences create duplicate relationships but one membership contribution per parent property and lifecycle authority.
- Make detach use canonical processed state, so in-place backing mutations cannot strand or detach the wrong child.

The implementation is independent of #419 and #440 and is designed against current `master`.

## Semantics

- `Children`, Registry `Parents`, and tracked `GetParents()` preserve exact occurrence order, including duplicates.
- Re-keyed or moved occurrences receive new immutable relationship objects. Unchanged dictionary keys are reusable only when the key object is the same reference.
- Previously returned arrays and retained relationship objects remain frozen.
- Readers overlapping a write can observe an old or new coherent generation. All enabled views agree after the write and synchronous callbacks become quiescent.
- An empty relationship group is the detach publication.
- Assigning the same non-string enumerable instance is an explicit structural refresh when normal equality handling suppresses the setter. It does not emit normal property or derived notifications.
- Different-property callback writes remain supported during ordinary reconciliation. Same-property re-entry is rejected. Same-subject structural writes during initial attach are rejected before terminal backing mutation.

## API changes

Breaking changes are intentional for this PR.

- Remove `IPropertyLifecycleHandler.RefreshCollectionProperty(PropertyReference, object?)`.
- Add `IPropertyRelationshipHandler.ReconcileChildRelationships(PropertyReference, ReadOnlySpan<SubjectPropertyRelationship>)`.
- Add immutable `SubjectPropertyRelationship` with `Parent`, `Child`, and `Index`.
- Add optional `SubjectLifecycleChange.Relationship` occurrence metadata.
- `SubjectRegistry` and `ParentTrackingHandler` implement the new relationship contract.
- Existing `Children`, `Parents`, and `GetParents()` signatures remain unchanged, but their contents now preserve occurrences instead of collapsing them to distinct memberships.

## Correctness and concurrency

Reconciliation enumerates once and stages a complete generation before mutating graph state. Matching uses subject reference identity and occurrence number. Commit applies removals in reverse old order, additions in source order, then publishes the immutable generation.

The branch includes deterministic coverage for:

- duplicate occurrences, order, re-key, reorder, removal, and same-instance refresh;
- hostile subject and dictionary-key equality/hash implementations;
- frozen snapshots and old-or-new reader publication;
- attach, detach, re-entrant detach, callback failure, retry, and empty-group cleanup;
- multiple lifecycle authorities, including three-authority same-property unwind;
- concurrent same-property and different-property writes, setter-only terminal ordering, detach/reattach, source-originated writes, and cache publication races;
- initial-attach enumeration, membership callbacks, and relationship callbacks attempting structural writes;
- HomeBlaze path invalidation, subject reference identity, and stale cache publication.

After its findings were fixed, the original independent whole-branch reviewer completed a final re-review with 0 Critical, 0 Important, and 0 Minor issues on implementation head `3f16e6f0`. The following commit only removes the internal design artifact from the PR.

## Performance

Static analysis shows average linear reconciliation in source occurrences plus distinct memberships. Registry provisional membership updates use a temporary reference-identity overlay, avoiding the earlier quadratic immutable-array rebuild. Public snapshot reads are cached and allocation-free after construction.

Focused allocation regressions are green:

- Registry provisional update scaling improved from a 60.33x allocation ratio for 8x input growth to 9.00x.
- HomeBlaze repeated invalidation of empty caches improved from 4,130,064 bytes to 0 bytes for 2,001 callbacks after warmup.

Exact occurrence storage and frozen snapshots intentionally cost one relationship per occurrence and executing lifecycle authority when relationship consumers are enabled. Lifecycle-only mode does not materialize those relationship objects.

The final BenchmarkDotNet comparison is intentionally pending for the other machine. The benchmark project builds cleanly and the target filter selects 9 methods and 34 parameterized cases. The required three-launch comparison uses the final benchmark definitions backported byte-for-byte to pinned production base `868a4d109d53b24805c9ee180efbf5029ee12c1a`:

```powershell
pwsh scripts/benchmark.ps1 -Filter "*ChildIndexRefreshBenchmark*","*RegistryBenchmark.AddLotsOfPreviousCars*","*ParentLookupBenchmark*","*ServiceOrderResolverBenchmark.LinearChain*" -LaunchCount 3 -BaseBranch benchmark/pr458-base-868a4d10
```

Allocation and 4-to-1000 scaling are the primary acceptance signals. `LinearChain` is only the timing noise reference.

## Verification

Final head: `fdb1948220d71c2ac5c61444bec38fc872be9da1`

- Debug and Release solution builds: passed, 0 warnings, 0 errors.
- Non-integration solution suite: 3,248 passed across 25 projects, 0 failed, 0 skipped.
- Tracking: 559/559.
- Registry: 205/205.
- HomeBlaze Services: 230/230.
- Tracking and Registry public API snapshots: passed.
- Registry and HomeBlaze allocation regressions: passed.
- Benchmark Release build: passed, 0 warnings, 0 errors; 9 methods and 34 cases selected.
- `git diff --check`: clean; no `.received.txt` files.
- GitHub Actions: general tests and HomeBlaze, MQTT, OPC UA, and WebSocket integration jobs passed.
- WebSocket integration also passed locally in Release, 34/34.

Connector implementation code did not change, so the standalone multi-hour Connector Tester exercise was not run. Connector unit projects are included in the non-integration solution result above.

## Review surface

Against merged `master` `868a4d109d53b24805c9ee180efbf5029ee12c1a`: 49 files, +8,645/-715.

| Area | Files | Added | Removed |
| --- | ---: | ---: | ---: |
| Production | 17 | 2,117 | 602 |
| Tests | 28 | 6,157 | 108 |
| Benchmarks | 2 | 284 | 0 |
| Documentation | 2 | 87 | 5 |

The core Tracking, Registry, and parent-consumer changes are one coupled consistency boundary. HomeBlaze contains the small downstream cache adaptation required by retained-membership re-key and reorder semantics.

## Intentional behavior changes from master

- Duplicate child occurrences are visible as duplicate ordered relationships instead of one collapsed entry.
- Dictionary relationships and all parent views follow current source order and keys.
- Distinct subjects that compare equal by value remain distinct graph identities.
- Same-instance enumerable assignment refreshes structural metadata without normal write notifications.
- Relationship handlers receive synchronous complete-group replacement callbacks and must be thread-safe across lifecycle authorities.
- The existing multi-authority ownership limitation for overlapping authorities remains documented; this PR does not depend on context-cardinality or hosting-lifecycle proposals.


RicoSuter:
## Benchmark results

Measured `341fb6bb` against `master` (`8a87c39a`). `ChildIndexRefreshBenchmark` exists only on this branch, so I backported the file unchanged onto the base and both arms carry it, which is what makes those rows comparable at all. It compiles on `master` as is.

Setup: CPU pinned to 3.6 GHz, no turbo, `performance` governor, checked through sysfs before and after the run rather than from the BenchmarkDotNet header, which printed two different maxima for the two arms. `LaunchCount=3`, memory randomization on, worktree outside the repository, machine otherwise idle.

Noise floor from `ServiceOrderResolverBenchmark`, which touches no subject in either its benchmarks or its setup: `+1.8%`, `+2.5%`, `+1.7%`, `-0.3%`. Anything under roughly 3% below is unresolved.

### Where it gets faster

The common case, rewriting a container whose members and order did not change, is much cheaper, because there is no position dictionary to build and no sort to run.

| Row | master | branch | delta |
|---|---:|---:|---:|
| `RewriteCollectionSameOrder`, n=1000 | 152.6 us | 105.7 us | **-30.8%** |
| `ReverseCollection`, n=4 | 1,398 ns | 1,232 ns | -11.8% |
| `ShiftCollectionByOne`, n=4 | 1,379 ns | 1,220 ns | -11.5% |
| `RewriteCollectionSameOrder`, n=4 | 1,085 ns | 961 ns | -11.4% |
| `SubjectUpdate.CreatePartialUpdate` | 2,557 ns | 2,422 ns | -5.3% |
| `SubjectUpdate.CreateCompleteUpdate` | 6,381 ns | 6,068 ns | -4.9% |
| `SourcePathProvider.TryGetPropertyFromSegment` | 47.3 ns | 45.1 ns | -4.6% |
| `SourcePathProvider.TryGetPath` | 251.7 ns | 243.9 ns | -3.1% |

The last four are the container-kind byte, and they are small but consistent in direction across four independent rows.

Allocations improve wherever children actually move, because the span's boxed index is reused instead of a fresh box per changed child:

| Row | master | branch | delta |
|---|---:|---:|---:|
| `ShiftCollectionByOne` / `ReverseCollection`, n=1000 | 72,104 B | 48,104 B | **-33.3%** |
| same rows, n=4 | 392 B | 296 B | -24.5% |

24,000 B at n=1000 is exactly 24 B times 1000, the re-boxing that is now gone.

### Where it gets slower: the refresh is quadratic in displaced children

These rows are `TrackParents=false`, so they isolate `RefreshChildIndices` itself.

| Row, n=1000 | master | branch | delta |
|---|---:|---:|---:|
| `ReorderDictionary` | 115.4 us | 642.8 us | **+457%** |
| `ReverseCollection` | 249.0 us | 687.3 us | **+176%** |
| `ShiftCollectionByOne` | 241.3 us | 321.6 us | **+33%** |
| `RekeyOneDictionaryEntry` | 109.9 us | 121.4 us | +10.5% |
| `RewriteDictionarySameKeys` | 111.2 us | 120.3 us | +8.3% |

`ReverseCollection` is the apples-to-apples one, since `master` already reordered collections. The mechanism is in the loop: per incoming child there is a linear scan from `slot`, and then a displaced child costs `RemoveAt(position)` plus `Insert(slot, ...)`, which are both O(n). Inside a loop over n children that is O(n^2), where `master` built a position dictionary in O(n) and finished with a single `List.Sort` at O(n log n).

`ReorderDictionary` at +457% compares new work against no work, since `master` returned early for anything that was not a collection, so read it as the cost of the fix rather than as a slowdown of the same job. It is worth flagging anyway, because `Dictionary<K,V>` enumeration order is not stable across mutation, so a rewrite that reuses the same logical content can still present a different order and pay the full quadratic reorder without anything having semantically moved.

At n=4 every collection row is faster, so this is purely a scale effect. The question is whether containers large enough to matter get reordered in practice; a few hundred entries is where the two curves cross.

### Parent tracking pays a full pass even when nothing moved

`ParentTrackingHandler.RefreshChildIndices` walks every child and calls `UpdateParentIndex`, which scans the parents set to find the entry before discovering the index is unchanged. Comparing the two `TrackParents` arms on this branch at n=1000:

| Row | `TrackParents=false` | `TrackParents=true` | cost |
|---|---:|---:|---:|
| `RewriteCollectionSameOrder` | 105.7 us | 164.3 us | +55% |
| `RewriteDictionarySameKeys` | 120.3 us | 185.0 us | +54% |

Both of those rows move no index at all. Skipping the callback when the index compares equal would take most of that back, and would also help the quadratic rows.

### The write rows are a layout artifact, not a cost

`RegistryBenchmark.Write` came out `+9.3%` and `WriteWithTimestampScope` `+9.4%`, and both reproduced on an independent repeat run at `+6.2%` and `+8.7%` with a flat control. That looked real, and it is not. `Write` and `WriteWithTimestampScope` are exactly the rows that enter `LifecycleInterceptor.WriteProperty` and then early-out on the type predicate, while `WriteNoOp` short-circuits earlier in the equality handler and is unaffected, so the enlarged method was the obvious suspect.

Diffing the JIT output for all four `WriteProperty[System.Decimal]` instantiations, with `DOTNET_JitDisasmDiffable=1` and tiering off, gives one changed line in 1140:

```
< ; 0 inlinees with PGO data; 65 single block inlinees; 20 inlinees without PGO data
> ; 0 inlinees with PGO data; 85 single block inlinees; 21 inlinees without PGO data
```

Not one machine instruction differs, and that header line is the one `docs/benchmarking.md` says to strip before believing a diff. So no work was added to the scalar write path. The repeat reproducing it is not evidence either way, since both runs used the same pair of binaries and a placement difference reproduces on every launch instead of averaging out.

### Unaffected

`AddLotsOfPreviousCars` `+0.3%`, `ChangeAllTires` `-0.8%`, `ReadParents` `-0.7%`, `Read` `-0.7%`, `GetOrAddSubjectId` `-0.4%`, `GenerateSubjectId` `-0.0%`, all inside the noise floor. `KnownSubjectsSnapshot` moved from 0.10 ns to 0.19 ns, which is below the timer floor and says nothing.

### Verdict

No regression on the write path, on attach/detach, or on path resolution, and a real improvement on unchanged-order rewrites and on allocations. The one genuine regression is the quadratic reorder, which is invisible at n=4 and severe by n=1000. If large containers only ever get appended to or replaced wholesale, this ships as is; if any of them get reordered or re-keyed, restoring an O(n) placement, for instance building the target order into a scratch list and swapping rather than `RemoveAt`/`Insert` per child, would remove both that and most of the parent-tracking overhead.

Issue #466, opened 2026-08-15
Decide whether multiple transaction interceptors should be rejected at registration


RicoSuter:
`SubjectTransactionInterceptor.WriteProperty` resolves the interceptor bound to the written property's context with `TryGetService`, which throws when a context resolves more than one:

```csharp
var subjectInterceptor = context.Property.Subject.Context.TryGetService<SubjectTransactionInterceptor>();
```

Context inheritance makes that reachable: `ContextInheritanceHandler` adds the parent subject's context as a fallback on attach, and `GetServices` walks fallbacks, so a subject built on one transaction-enabled context and attached into a graph rooted on another resolves two. `WithTransactions()` guards against double registration only when the fallback already exists at call time.

Reproduced with two aggregated contexts both calling `WithTransactions()`, an ambient transaction, and an ordinary non-derived write:

```
System.InvalidOperationException: There must be exactly one service of type
  Namotion.Interceptor.Tracking.Transactions.SubjectTransactionInterceptor
```

The exception escapes a property setter. The surrounding code says why that is the wrong failure mode: the capture-race path a few lines below deliberately falls through rather than throwing, because "an exception escaping one on a scheduler thread or a bare thread takes the process down". With scheduled per-property subscriptions this is no longer hypothetical, since observers run on scheduler threads.

Derived properties do not reach it, because the `!context.Property.Metadata.IsDerived` guard on the enclosing condition short-circuits first.

## Suggested direction

Membership rather than exactly-one resolution:

```csharp
var isBoundToThisContext = context.Property.Subject.Context
    .GetServices<SubjectTransactionInterceptor>()
    .Contains(transaction.Interceptor);
```

`GetServices` returns `ImmutableArray<T>`, so this is allocation-free and measured marginally cheaper than `TryGetService`, which calls it and then switches on the length.

Worth deciding separately whether two transaction interceptors on one subject is a configuration that should be rejected at registration time instead, since the write path then has to answer which one owns a capture.

RicoSuter:
The throw is fixed in #455 (`bd0556d8`): the write path now tests membership with `GetServices<SubjectTransactionInterceptor>().Contains(transaction.Interceptor)` instead of requiring exactly one, with the two-context repro as a regression test. The deliberate throw for a transaction bound to a different context is unchanged.

[...]

Issue #465, opened 2026-08-14
Derived cascade suppression infers capture from ambient transaction state and drops cascades


RicoSuter:
`DerivedPropertyChangeHandler.WriteProperty` decides whether to suppress a derived cascade by inspecting ambient transaction state after the inner write has returned:

```csharp
if (SubjectTransaction.HasActiveTransaction &&
    !context.Property.Metadata.IsDerived &&
    SubjectTransaction.Current is { IsCommitting: false, IsDisposed: false } transaction &&
    context.Property.Subject.Context.GetServices<SubjectTransactionInterceptor>().Contains(transaction.Interceptor))
{
    return;
}
```

The intent is "this write was captured and the commit will replay its cascade". Ambient state cannot express that, because the transaction interceptor reads `SubjectTransaction.Current` on the way into the write while this reads it again on the way out, and user code running inside `next` can change it in between.

## Reproduces today

Stock `WithFullPropertyTracking().WithTransactions()`. No transaction is ambient when the write starts, so the transaction interceptor takes its `!HasActiveTransaction` fast path and the write reaches the model. Inside the same write, `PropertyChangeInterceptor` dispatches synchronously to an observer that opens a transaction, and `TransactionAwaiter.GetResult()` publishes it into the ambient slot of the flow the write is running on. When control unwinds to the derived handler, all four conditions hold and the cascade is suppressed for a write nothing captured.

```csharp
var context = InterceptorSubjectContext.Create().WithFullPropertyTracking().WithTransactions();
var person = new Person(context) { LastName = "Doe" };
_ = person.FullName;

SubjectTransaction? ambientTransaction = null;
var changedProperties = new List<string>();
using var subscription = context
    .GetPropertyChangeObservable(ImmediateScheduler.Instance)
    .Subscribe(change =>
    {
        changedProperties.Add(change.Property.Name);
        if (change.Property.Name == nameof(Person.FirstName) && ambientTransaction is null)
        {
            ambientTransaction = context
                .BeginTransactionAsync(TransactionFailureHandling.BestEffort)
                .GetAwaiter()
                .GetResult();
        }
    });

person.FirstName = "Jane";
ambientTransaction?.Dispose();

Assert.Equal("Jane Doe", person.FullName);                              // passes
Assert.Contains(nameof(Person.FullName), changedProperties);            // fails
Assert.Contains(nameof(Person.FullNameWithPrefix), changedProperties);  // fails
```

[...]

PR #468, opened 2026-08-16
Fix transaction lifecycle and derived tracking correctness

### LINKED ISSUE #467: Derived cascades are dropped for writes a transaction never captured
`DerivedPropertyChangeHandler` suppresses a derived cascade while the ambient transaction can still serve reads from its pending buffer, because a dependent's getter reads through the transaction and would otherwise be computed from uncommitted state and published. The commit replays the captured writes and their cascades run there instead.

That is correct for writes the transaction captured. It is wrong for writes it did not, and there are three of those. In each, no commit will ever replay the cascade, so the dependents stay stale until an unrelated write to the same dependency.

## The three cases

**A derived property written inside a transaction.** Capture skips derived properties (`SubjectTransactionInterceptor` checks `!IsDerived`), so the write lands immediately. Its cascade is suppressed and never replayed.

```csharp
// Nickname is [Derived] with a setter, NicknameWithPrefix depends on it
using (await context.BeginTransactionAsync(TransactionFailureHandling.BestEffort))
{
    person.Nickname = "Rico";
}
// NicknameWithPrefix is never announced
```

**A transaction open on an unrelated context.** The ambient slot is process-wide while capture is per-context, so a transaction on context A suppresses the cascade for a write to context B, which may have no transaction interceptor at all. B never captures the write, so B has no commit to replay it.

**A transaction that becomes ambient during an uncaptured write.** No transaction is ambient when the write starts, so the interceptor takes its fast path and the write lands. A synchronous change observer then opens one, and `TransactionAwaiter.GetResult` publishes it into the ambient slot of the flow the write is running on. By the time the cascade is decided, a live transaction is bound to this context for a write nothing captured.

All three had tests on the branch that fixed them; the tests were removed when the fix was reverted, because the fix reopened a worse defect (publishing a dependent computed from uncommitted state, which survives rollback).

## Why the obvious fixes do not work

Deciding on `context.IsWritten` ("did the value reach the model") fixes all three and reintroduces the uncommitted publish, because it answers only whether the *trigger* landed, while the cascade's *inputs* come from other properties that the transaction may be holding.

Narrowing the suppression to a non-disposed transaction reopens the rollback case, because `Dispose` sets the disposed flag before it releases the pending buffer, so there is a window where the transaction reads as disposed and still feeds getters.

## Suggested direction

Mask the ambient transaction while the cascade evaluates dependent getters, so a recalculation always computes from committed state regardless of what the flow is holding. The cascade can then run unconditionally: during capture it announces the committed value, and the commit replays the captured writes and announces the updated one.

The cost to weigh is that a derived property recalculated inside a transaction would no longer reflect that flow's pending writes in its announced value and `LastKnownValue`, while a direct read of the same property still would, because that goes through the normal read path. Whether that divergence is acceptable is the decision this needs. It also lands in `EvaluateAndStabilize`, which carries the stabilization loop and the per-property locks, so it wants its own review rather than being folded into unrelated work.

RicoSuter:
## Summary

- Make transaction capture, commit, terminal completion, and disposal ownership atomic, including stale ambient transaction handling.
- Reject access carrying a live ambient transaction while a commit owns the frozen snapshot, while allowing only synchronous commit-owned model replay.
- Preserve direct transaction read-your-writes while ensuring derived tracking observes landed model values after rollback, failure, transformation, and reentrant evaluation.
- Scope cascade suppression to writes that actually landed, so vetoed writes and unrelated writes retain the correct behavior.
- Keep multiple and inherited transaction interceptor registrations correct while optimizing the common singleton context-binding path.

## Contract

- Direct reads inside a live transaction return pending values.
- After terminal commit completion, the still-ambient transaction is inactive: reads see the model and writes land normally instead of entering a dead buffer.
- Property access carrying a live ambient transaction during an in-progress commit throws. Synchronous conflict detection, apply, revert, hooks, and derived cascades owned by that commit remain valid.
- Disposal makes the ambient transaction inactive immediately. Ordinary raw writes may therefore interleave with an already-started external-writer commit and may later be overwritten by its frozen replay.
- Derived tracking and shared derived values are calculated from landed model state.
- Commit publishes and cascades changes in deterministic landed order.
- Nested and concurrent derived evaluation uses invocation-owned, thread-local recording state.
- Settable-derived writes retain their current immediate, non-transactional behavior; #471 tracks the separate design decision.
- No public API is added or removed.

## Behavior change for transaction writers

Misuse of a live transaction during commit now fails loudly. While `CommitAsync()` is in progress, reading or writing a property on a flow carrying that live transaction throws `InvalidOperationException`. Previously, such writes could bypass the frozen transaction snapshot and land directly in the model. Synchronous model replay owned by the commit, including property hooks, change handlers, and derived cascades, remains authorized.

Custom `ITransactionWriter` implementations must construct source operations from the supplied change snapshot, writer-owned configuration, and any required subject state captured before `CommitAsync()`. They must not read or write subject properties from `WriteToSourcesAsync` or `RevertSourceWritesAsync`; even a getter reads landed sibling state outside the frozen snapshot and can make the source payload inconsistent.

## Verification

- `dotnet build src/Namotion.Interceptor.slnx --no-restore`
  - Passed with 0 warnings and 0 errors.
- `dotnet test src/Namotion.Interceptor.slnx --filter "Category!=Integration" --no-build --no-restore`
  - Passed: 3,103 tests, 0 failed, 0 skipped, including 474 tracking tests.
- Targeted transaction, terminal-state, frozen-snapshot access, rollback, veto, nested derived, derived-of-derived, reentrancy, disposal boundary, multi-context, and mutation-sensitivity tests pass.
- Independent correctness and concurrency reviews found no blocking defect. Final minor feedback was incorporated with deterministic public-path coverage and documentation corrections.
- Detailed cross-revision benchmark evidence and controls are recorded in [the review comment](https://github.com/RicoSuter/Namotion.Interceptor/pull/468#issuecomment-5306973304). Reported allocations were unchanged, and timing deltas stayed within the measured control noise.

## Scope

This PR contains only the transaction and derived-tracking correctness foundation. Scheduled property subscriptions follow in #469, stacked on this branch.

The concrete reproduction from #465 is fixed here; that issue remains closed as superseded by #467.

Fixes #467
Addresses #466
Follow-up: #471

### DIFF
--- a/docs/design/tracking-derived-properties.md
+++ b/docs/design/tracking-derived-properties.md
@@ -20,7 +20,7 @@ public partial class Person
 
 The source generator detects `[Derived]` and sets `IsDerived = true` on `SubjectPropertyMetadata`, but does not generate a backing field. The getter body is the user's original C# expression.
 
-Derived properties can also be added at runtime via `RegisteredSubject.AddDerivedProperty<T>(name, getValue, setValue)`. These work identically with the dependency tracking system — the getter/setter lambdas are wrapped in interception calls and dependencies are recorded on first evaluation.
+Derived properties can also be added at runtime via `RegisteredSubject.AddDerivedProperty<T>(name, getValue, setValue)`. These work identically with the dependency tracking system: the getter/setter lambdas are wrapped in interception calls and dependencies are recorded on first evaluation.
 
 ## Interceptor Ordering
 
@@ -53,7 +53,7 @@ Dependencies (RequiredProperties):      Used-by properties (UsedByProperties):
 - **Dependencies** (`data.RequiredPropertiesSpan`): Which properties a derived property reads. Stored as a private `PropertyReference[]` with separate count, accessed via `ReadOnlySpan` under `lock(data)`. Buffer is reused when capacity is sufficient to avoid allocation on re-evaluation.
 - **Used-by properties** (`data.GetUsedByProperties()`): Which derived properties depend on this property. Stored as private `PropertyReferenceCollection`, updated via lock-free CAS.
 
-The two use different data structures because of their access patterns. Dependencies are always read and written under `lock(data)` (the derived property's own data), so a plain array with buffer reuse is sufficient. Used-by properties are read without any lock (`WriteProperty` iterates dependents via `GetUsedByProperties()` which does `Volatile.Read`) and mutated from multiple lock scopes (`Remove` is called under the *derived* property's lock, not the *source* property's lock — so two derived properties detaching concurrently can call `Remove` on the same source's `UsedByProperties`). This requires the lock-free CAS copy-on-write wrapper.
+The two use different data structures because of their access patterns. Dependencies are always read and written under `lock(data)` (the derived property's own data), so a plain array with buffer reuse is sufficient. Used-by properties are read without any lock (`WriteProperty` iterates dependents via `GetUsedByProperties()` which does `Volatile.Read`) and mutated from multiple lock scopes (`Remove` is called under the *derived* property's lock, not the *source* property's lock, so two derived properties detaching concurrently can call `Remove` on the same source's `UsedByProperties`). This requires the lock-free CAS copy-on-write wrapper.
 
 Both directions are needed:
 - Dependencies enable cleanup when a derived property is detached (remove itself from all dependencies' used-by properties).
@@ -82,7 +82,7 @@ sequenceDiagram
     participant Full as FullName (derived)
     participant Sub as Subscribers
 
-    Note over App,Sub: Attach — evaluate getter, build dependency graph
+    Note over App,Sub: Attach - evaluate getter, build dependency graph
     App->>Full: AttachProperty(FullName)
     Full->>FN: reads FirstName (recorded as dependency)
     Full->>LN: reads LastName (recorded as dependency)
@@ -90,7 +90,7 @@ sequenceDiagram
     Note over LN: UsedBy += [FullName]
     Note over Full: LastKnownValue = initial value
 
-    Note over App,Sub: Write — cascade via UsedByProperties
+    Note over App,Sub: Write - cascade via UsedByProperties
     App->>FN: FirstName = "Jane"
     FN->>Full: lookup UsedBy → RecalculateDerivedProperty
     Full->>FN: evaluate getter (reads FirstName)
@@ -98,7 +98,7 @@ sequenceDiagram
     Note over Full: commit new LastKnownValue
     Full->>Sub: notify(oldValue, newValue)
 
-    Note over App,Sub: Detach — clean up both directions
+    Note over App,Sub: Detach - clean up both directions
     App->>Full: DetachProperty(FullName)
     Note over FN: UsedBy -= [FullName]
     Note over LN: UsedBy -= [FullName]
@@ -117,7 +117,7 @@ AttachProperty(FullName)
       data.LastKnownValue = EvaluateAndStabilize(data, FullName)
       SetWriteTimestamp(...)
     catch:
-      // Getter threw — value will be computed on the next dependency write.
+      // Getter threw - value will be computed on the next dependency write.
 ```
 
 `EvaluateAndStabilize` handles the dependency recording, generation check, and stabilization loop internally (see Section 4 for details). If the getter throws during initial evaluation (typically during concurrent state transitions), the exception is caught and `LastKnownValue` remains null. The next dependency write triggers recalculation which retries with consistent state.
@@ -136,7 +136,7 @@ EvaluateAndStabilize(data, FullName)
     dependenciesChanged = data.UpdateDependencies(FullName, recordedDeps, recorder)
     if !dependenciesChanged || Volatile.Read(_writeGeneration) == generationBefore:
       return result                              // common path: no concurrent write
-    // Concurrent write detected — stabilize
+    // Concurrent write detected - stabilize
     for iteration in 0..<MaxStabilizationIterations:
       StartRecordingTouchedProperties()
       result = getter.Invoke(subject)
@@ -148,7 +148,7 @@ EvaluateAndStabilize(data, FullName)
     DiscardActiveRecording()                     // always clears recorder buffer
 ```
 
-On the common path (single-threaded construction), `_writeGeneration` is unchanged and the loop is skipped entirely — zero extra getter evaluations. If a concurrent write is detected, the full stabilization loop runs for correctness.
+On the common path (single-threaded construction), `_writeGeneration` is unchanged and the loop is skipped entirely: zero extra getter evaluations. If a concurrent write is detected, the full stabilization loop runs for correctness.
 
 ### 2. Read (recording)
 
@@ -184,7 +184,7 @@ WriteProperty(FirstName = "Jane")
       RecalculateDerivedProperty(FullName, timestamp)
 ```
 
-The `_writeGeneration` increment uses `Interlocked.Increment` (full fence) so that each concurrent writer produces a unique counter value — no lost increments. `AttachProperty`/`RecalculateDerivedProperty` detect concurrent writes via `Volatile.Read` (acquire semantics). The full fence from `Interlocked.Increment` pairs with `Volatile.Read`'s acquire semantics to guarantee that committed property values are visible when the counter change is observed.
+The `_writeGeneration` increment uses `Interlocked.Increment` (full fence) so that each concurrent writer produces a unique counter value: no lost increments. `AttachProperty`/`RecalculateDerivedProperty` detect concurrent writes via `Volatile.Read` (acquire semantics). The full fence from `Interlocked.Increment` pairs with `Volatile.Read`'s acquire semantics to guarantee that committed property values are visible when the counter change is observed.
 
 **Derived properties with setters** (created via `AddDerivedProperty<T>(name, getValue, setValue)`) have both a getter and a setter. The setter modifies internal state as a side effect, but the actual property value is always determined by the getter. When the setter is called, `WriteProperty` detects `HasRequiredProperties && SetValue != null` and triggers recalculation to re-evaluate the getter and fire a change notification with the correct computed value.
 
@@ -194,7 +194,7 @@ The `_writeGeneration` increment uses `Interlocked.Increment` (full fence) so th
 
 The following sequence diagrams illustrate the concrete execution flow of `RecalculateDerivedProperty`. The single-threaded trace shows the straight-line happy path where all concurrency guards are inert. The concurrent trace shows how `IsRecalculating` serializes notification delivery to prevent stale out-of-order notifications.
 
-**Single-threaded recalculation** — all concurrency guards are no-ops:
+**Single-threaded recalculation**: all concurrency guards are no-ops:
 
 ```mermaid
 sequenceDiagram
@@ -215,7 +215,7 @@ sequenceDiagram
     W->>D: finally: IsRecalculating = false
 ```
 
-**Concurrent write during notification** — the scenario that `IsRecalculating` serialization prevents:
+**Concurrent write during notification**: the scenario that `IsRecalculating` serialization prevents:
 
 ```mermaid
 sequenceDiagram
@@ -248,7 +248,7 @@ sequenceDiagram
     A->>D: finally: IsRecalculating = false
 ```
 
-Without the `IsRecalculating` serialization (the pre-fix behavior), Thread B could have entered `RecalculateDerivedProperty` during Thread A's notification delivery, completed a full recalculation with the newer value, and delivered its notification first. Thread A would then resume and deliver its stale notification second — violating the invariant that the last notification reflects the final computed value.
+Without the `IsRecalculating` serialization (the pre-fix behavior), Thread B could have entered `RecalculateDerivedProperty` during Thread A's notification delivery, completed a full recalculation with the newer value, and delivered its notification first. Thread A would then resume and deliver its stale notification second, violating the invariant that the last notification reflects the final computed value.
 
 #### Algorithm
 
@@ -267,7 +267,7 @@ RecalculateDerivedProperty(FullName, timestamp)
 
   // Outer loop: handles post-notification RecalculationNeeded without recursion.
   // The try-finally wraps the entire loop so IsRecalculating stays true during
-  // NotifyDerivedPropertyChanged — this serializes notification delivery with
+  // NotifyDerivedPropertyChanged - this serializes notification delivery with
   // recalculation, preventing stale notifications from being delivered after newer ones.
   try:
     for outerIteration in 0..<MaxStabilizationIterations:
@@ -330,16 +330,16 @@ NotifyDerivedPropertyChanged(derivedProperty, data, sequence, newValue, oldValue
 The getter is evaluated inside `EvaluateAndStabilize` **without holding `lock(data)`** (when `callerHoldsLock` is false). The lock is acquired only briefly for `UpdateDependencies`. If the getter throws (typically during concurrent state transitions), the exception is caught and `LastKnownValue` remains unchanged. The concurrent writer's `WriteProperty` cascade will re-trigger recalculation with consistent state.
 
 The `RecalculationNeeded` flag is set under `lock(data)` by three sources when `IsRecalculating` is true:
-- **Concurrent `RecalculateDerivedProperty`**: Another write triggered recalculation but the current one is in progress — the concurrent call bails and signals the flag.
-- **`AttachProperty`**: The property is being reattached while recalculation is in progress — the evaluation result may be stale.
-- **`DetachProperty`**: The property is being detached while recalculation is in progress — the evaluation result is invalid.
+- **Concurrent `RecalculateDerivedProperty`**: Another write triggered recalculation but the current one is in progress, so the concurrent call bails and signals the flag.
+- **`AttachProperty`**: The property is being reattached while recalculation is in progress, so the evaluation result may be stale.
+- **`DetachProperty`**: The property is being detached while recalculation is in progress, so the evaluation result is invalid.
 
-Phase 3 checks the flag before committing. If set, the stale result is discarded and the inner loop re-evaluates with fresh state. Because `IsRecalculating` stays true during notification delivery, any concurrent write sets `RecalculationNeeded` and bails — the outer loop picks it up after notification completes and re-evaluates with the latest state.
+Phase 3 checks the flag before committing. If set, the stale result is discarded and the inner loop re-evaluates with fresh state. Because `IsRecalculating` stays true during notification delivery, any concurrent write sets `RecalculationNeeded` and bails, and the outer loop picks it up after notification completes and re-evaluates with the latest state.
 
 The generation check inside `EvaluateAndStabilize` avoids re-evaluation when dependencies change but no concurrent write occurred. The stabilization loop only runs when a concurrent write is actually detected.
 
 Key details of the change notification:
-- **Notifications outside lock but inside `IsRecalculating`**: `NotifyDerivedPropertyChanged` fires `SetPropertyValueWithInterception` and `RaisePropertyChanged` without holding `lock(data)` (preventing deadlock with `lock(_attachedSubjects)`), but while `IsRecalculating` is still true. This serializes notification delivery with recalculation — no concurrent recalculation (and thus no competing notification) can start during delivery. Two additional guards provide defense-in-depth: a `RecalculationSequence` check and a `ReferenceEquals` check on `LastKnownValue`. See the "Deadlock prevention" section for details.
+- **Notifications outside lock but inside `IsRecalculating`**: `NotifyDerivedPropertyChanged` fires `SetPropertyValueWithInterception` and `RaisePropertyChanged` without holding `lock(data)` (preventing deadlock with `lock(_attachedSubjects)`), but while `IsRecalculating` is still true. This serializes notification delivery with recalculation, so no concurrent recalculation (and thus no competing notification) can start during delivery. Two additional guards provide defense-in-depth: a `RecalculationSequence` check and a `ReferenceEquals` check on `LastKnownValue`. See the "Deadlock prevention" section for details.
 - **Timestamp inheritance**: The derived property receives the same timestamp as the write that triggered the recalculation, ensuring consistent timestamps within a mutation context. This also holds under an explicit-null scope (`WithChangedTimestamp(null)`): storage remains the never-written sentinel, but trigger and cascade dependents share a single captured publishing time so change events stay consistent.
 - **Local origin by default**: Origin is stamped per write and nothing inherits it, so the recalculation notification publishes with a `Local` origin without any scope. The local model computed the value and no source confirmed it, so the change flows to bound sources like any local write.
 - **`NoOpWriteDelegate`**: Since derived properties have no backing field, the write delegate is a no-op (`static (_, _) => { }`). The call to `SetPropertyValueWithInterception` exists solely to fire the change notification through the interceptor chain (observable, queue, etc.) with the correct old and new values.
@@ -349,7 +349,7 @@ Key details of the change notification:
 
 When a subject is removed from the object graph, `DetachProperty` cleans up both directions.
 
-`DetachProperty` uses `TryGetDerivedPropertyData()` and returns early if no tracking data exists. This avoids `ConcurrentDictionary.GetOrAdd` allocations for source properties that were never dependencies — a significant performance optimization when detaching subjects with many non-dependency properties (e.g., detaching 1000 cars where only 1 of 18 properties per car participates in derived property tracking).
+`DetachProperty` uses `TryGetDerivedPropertyData()` and returns early if no tracking data exists. This avoids `ConcurrentDictionary.GetOrAdd` allocations for source properties that were never dependencies, a significant performance optimization when detaching subjects with many non-dependency properties (e.g., detaching 1000 cars where only 1 of 18 properties per car participates in derived property tracking).
 
 For properties with tracking data, both cases are handled in a single `lock(data)` block, followed by Case 2 cleanup outside the lock:
 
@@ -371,7 +371,7 @@ DetachProperty(property)
       derivedData.RemoveRequiredProperty(property)
 ```
 
-The single lock ensures `IsAttached`, dependency cleanup (Case 1), and used-by snapshot (Case 2) are atomic — no window where a concurrent thread could see `IsAttached=false` but `UsedByProperties` still populated.
+The single lock ensures `IsAttached`, dependency cleanup (Case 1), and used-by snapshot (Case 2) are atomic: no window exists where a concurrent thread could see `IsAttached=false` but `UsedByProperties` still populated.
 
 The lock serializes with `UpdateDependencies`' used-by Add (which also acquires `lock(depData)` on the same object). This ensures either:
 - The used-by property was added before the snapshot → we see it and clean up the dependency.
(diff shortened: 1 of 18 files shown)


RicoSuter:
## Review: ready to merge, no blockers

Reviewed `fix/transaction-derived-correctness` at 63e08e77 against master at 8a87c39a in a clean worktree. Correctness is sound, allocations are unchanged, and the tests are genuine regression tests rather than tests written to the new behavior. Five suggestions below, all non-blocking.

### Findings: what is broken on master and fixed here

Each one reproduced on master, not inferred from reading.

| # | Defect on master | Severity | Evidence |
|---|---|---|---|
| 1 | `DiscardActiveRecording()` tests `_recorder.IsRecording` (`_depth > 0`), so a reentrant evaluation's inner `finally` pops the **outer** frame and wipes its dependencies. The outer's own `FinishRecording()` drives `_depth` to -1 and throws `IndexOutOfRangeException`, swallowed. The next derived property attached anywhere on that thread hits `_frames[_depth++]`, throws again, is swallowed by `AttachProperty`, and attaches with an empty dependency set. It never recalculates again, silently, and the recorder then looks healthy. | High | Transaction-free repro fails on master, passes here. Nothing in master's suite triggers reentrancy, which is why CI is green while carrying this. |
| 2 | A write racing `Dispose()` was captured after the pending dictionary had been cleared and returned to `PendingChangesPool`, so a later transaction could rent a dictionary holding a foreign pending change. | High | New disposal-race tests fail on master. |
| 3 | A thread born inside a transaction keeps it in its `AsyncLocal` slot for life. Reads served stale values, writes were swallowed into a recycled dictionary, and `BeginTransactionAsync` threw "Nested transactions are not supported" forever on that flow. | High | Frozen-flow tests fail on master. Realistic trigger is an Rx `EventLoopScheduler` worker. |
| 4 | Any ambient transaction suppressed derived cascades globally, including for subjects in contexts that have no transactions registered. Because `SubjectTransactionInterceptor` is `[RunsBefore(DerivedPropertyChangeHandler)]` and capture returns without calling `next`, the only writes reaching that check had already landed, so their cascades were dropped with no capture to replay them. | High | `WhenTransactionIsOpenOnAnotherContext_...` fails on master with `["FirstName"]`. |
| 5 | Derived getters read pending transaction values, so tracking state could be computed from writes that never landed and survive rollback. Worse, the pending return short-circuits before `DerivedPropertyChangeHandler.ReadProperty`, so those properties were not recorded as dependencies at all. | High | Rollback and model-view tests fail on master. |
| 6 | `TryGetService<SubjectTransactionInterceptor>()` threw "There must be exactly one service" whenever context inheritance resolved two. | Medium | Inherited-interceptor test fails on master. |
| 7 | Vetoed writes bumped `_writeGeneration` and cascaded. | Low | Veto tests fail on master. |

### Issue status

| Issue | Tracker | Actual | Notes |
|---|---|---|---|
| #467 | open, "Fixes" | **Fully fixed** | All three cases have tests here and all three fail on master. The PR takes exactly the suggested direction (mask the ambient transaction during cascade evaluation) plus the `IsWritten` gate, which is the combination the issue says is required. The issue flags the read-versus-tracking divergence as "the decision this needs"; this PR makes that decision and documents it. Correct to auto-close on merge. |
| #466 | open, "Addresses" | **Partially fixed** | Write path fixed exactly as the issue suggests. `SubjectTransaction.BeginTransactionAsync` (`SubjectTransaction.cs:237`) still resolves with `TryGetService` and still throws the identical `InvalidOperationException` on the same doubly-resolving context. Verified on this branch: write succeeds and captures, begin throws. "Addresses" rather than "Fixes" is the right wording; keep the issue open with the residual narrowed to the begin path plus the undecided question of whether double registration should be rejected at configuration time. |
| #465 | **closed as not planned** | **Fixed here** | The issue's verbatim repro fails on master (`["FirstName"]`, `FullName` never announced) and passes on this branch. The tracker says not planned while the fix is shipping. Worth reopening so the merge closes it, or noting it in the PR body. |

Two follow-on asks from #465 are not implemented: sealing `DerivedPropertyChangeHandler` and a guard asserting every `SubjectTransactionInterceptor` sorts ahead of every `DerivedPropertyChangeHandler`. I think both can now be dropped rather than carried, because the `IsWritten` gate makes the suppression decision order-independent: a captured write leaves `IsWritten` false whether the derived handler sits inside or outside the transaction interceptor, and the read-side masking keeps a subclass that runs outside capture from publishing uncommitted state, which is the hazard those asks guarded. Worth recording as a decision rather than leaving as an omission.

### Verification

- Build 0 warnings, 0 errors. Full non-integration suite green, about 2,400 tests.
- Connector suites run unfiltered, since CI path filters skip them for shared-library PRs. All green including the integration tests CI would have missed: OPC UA 295, WebSocket 131, MQTT 75, Connectors 633.
- **Tests are real.** Every test file from this PR copied verbatim onto master compiles unchanged, and 23 fail there against 0 here.
- No public API change; `PublicApi` snapshots pass untouched.

### Regressions vs master

None found, in behavior or performance.

Targeted benchmark, `LaunchCount 3`, memory randomization on, CPU pinned (governor `performance`, min and max both 3.6 GHz, `no_turbo=1`). Ignore BenchmarkDotNet's per-arm "Max: 0.80GHz" and "Max: 1.55GHz" header lines, which is the unreliable-header artifact `docs/benchmarking.md` documents.

- **Allocations byte-identical on all 8 rows**, Gen0 and Gen1 included.
- Timings span -4.6% to +1.8% in a run whose own noise reference (`ServiceOrderResolver.LinearChain`) moved +0.4%, and `DerivedDeclaredSet` flips sign between mean and median. Nothing clears the floor in either direction.
- This also settles the one cost worth looking for: `FinalizeOrigin` now routes through `GetFinalOrigin()` and clears `_attempted` on the common Local path, a store the old early return skipped, on the hottest path in the library. `RootOnlySet` is exactly that path and moved faster, so it is not measurable.
- Two paths no benchmark reaches, both bounded by inspection: the derived-read bypass adds one thread-static read per intercepted read while a transaction is open, and the capture path swapped `TryGetService` for `GetServices`, which is a cached dictionary lookup that no longer pays the length switch. Neither allocates.

### Behavior changes consumers will notice

Not regressions, but they belong in release notes rather than only in the design doc.

1. **Derived properties have two views inside a transaction.** Direct reads return pending values; `LastKnownValue`, dependencies, timestamps and notifications use landed state. `WhenATransactionIsRolledBack_...` shows it: inside the transaction `subject.Combined` reads `"uncommitted|d1"` while subscribers are told `"committed|d1"`. That is what makes rollback safe, but a subscriber reading the property back on the transaction's flow sees a different value than the notification carried.
2. **Derived-with-setter writes inside a transaction now cascade and publish immediately**, from landed inputs. Connectors bound to those dependents see a change during the transaction that they did not before. It is correct; previously model and announced state diverged permanently.
3. **Writes on a flow holding a disposed transaction now land in the model** instead of being swallowed, and `BeginTransactionAsync` no longer throws forever on such a flow.
4. **Vetoed writes no longer cascade or bump the write generation**, aligning with the guard `PropertyChangeInterceptor` already had.

### Pre-existing gap this PR does not close, worth its own issue

**Derived properties with a setter are not transactional.** The capture guard excludes them (`!context.Property.Metadata.IsDerived`), unchanged from master. Probed on this branch:

```
inside tx:      Plain=P1, Derived=D1     (both read back as written)
after rollback: Plain=P0, Derived=D1     (the derived write survived)
```

It is sharper than it looks, because a `[Derived]` partial auto-property gets a real backing field, so storage-wise it is indistinguishable from a plain property. Marking a settable property `[Derived]` silently opts it out of transactions with no diagnostic. Expression-bodied derived properties have a null setter and can never be written, so the guard only ever excludes properties that are plain stored writes, which are exactly the ones that could be captured.

This PR does not make the semantics worse, and it does make the model self-consistent where master left it both non-transactional and permanently stale. It does make the gap louder for connectors, since the non-transactional write is now announced mid-transaction and a rollback will not retract it.

Caution for whoever fixes it: the guard is load-bearing for the cascade path. `NotifyDerivedPropertyChanged` republishes through `SetPropertyValueWithInterception` and re-enters the same write chain, so simply capturing derived writes would swallow every cascade into the transaction. A fix must distinguish a user write to a derived property from a cascade republication.

### Suggestions, all non-blocking

1. **Add a transaction-free regression test for finding 1.** The existing reentrancy test pins the symptom at the scene but not the blast radius, which is the part that bites a host: one reentrant getter silently kills the next derived property attached on that thread. A test in the shape "reentrant getter runs, then an unrelated subject on the same thread still tracks its derived property" fails loudly on any regression of `ownsActiveRecording`. Written and verified against both branches, happy to push it.
2. **Doc and code drift on the phase markers.** This strips `// Phase 1/2/3` from `RecalculateDerivedProperty`, but `docs/design/tracking-derived-properties.md` navigates by those labels about fifteen times, including an embedded code excerpt that still shows them, and one surviving in-code comment still says "Phase 3 never runs to clear it".
3. **Two load-bearing why-comments went out with the em dash cleanup** in `SubjectTransaction.cs`: the `_commitStarted` reset rationale recording which failures are retryable and why everything past `FinishCommit` is terminal, and why the deferred exclusive-lock release must wait for commit completion. Neither is recoverable from the code. Worth restoring in shortened form.
4. **`TryCaptureChange` checks `ThrowIfCommittingConcurrently()` before `_isDisposed`**, while `TryGetPendingValue` checks disposed first. `Dispose()`'s comment promises a waiting capture "observes it and continues normally", which does not hold in the narrow window where the transaction is both disposed and committing: the write throws out of the setter instead of falling through. Reaching it needs a dispose and a commit start inside the guard's TOCTOU window, so unlikely rather than impossible, but given the motivating scenario is an exception escaping a setter on a bare thread, matching the read path's ordering closes it in one line.
5. **State the rollback consequence in the user-facing doc.** `docs/tracking-transactions.md:404` says "Derived writes are not captured and can recalculate and notify immediately while a transaction is open", which is accurate but leaves the reader to derive "not captured" into "not transactional" into "survives rollback". One sentence saying a derived write is not rolled back would save that inference. Same for the dual view: the framing that lands is "per-flow reads are transactional, shared tracking state is not", rather than describing two views of one property.
6. **`DerivedPropertyWriteGenerationTests` asserts an exact getter call count against the process-global `_writeGeneration`.** `DisableParallelization` covers other collections in the assembly, but a background thread elsewhere in the process can still bump it. Worth remembering if it ever flakes in CI.

### Recommendation

**Merge. No blockers.**

Optional before merge, and only this one: suggestion 1, the transaction-free regression test for finding 1. That is the most dangerous defect in the PR and nothing currently fails on its blast radius. Everything else is a follow-up.

Tracker actions alongside the merge, so the issue state stays truthful:

- **Reopen #465.** It is closed as not planned, but its verbatim repro passes here and fails on master, so the fix is shipping in this PR. Reopening lets the merge close it for the right reason.
- **Keep #466 open**, narrowed to what is left: `BeginTransactionAsync` still resolving with `TryGetService`, plus the undecided question of whether two transaction interceptors on one subject should be rejected at configuration time. The "Addresses #466" wording in the description is already correct and should not become "Fixes".
- **#467 needs nothing.** It is fully fixed and will auto-close correctly.
- **Open a new issue** for the derived-with-setter transaction gap described above. It is pre-existing, not caused here, and folding it into this PR would be scope creep. Include the cascade-republication caution, since that is what makes the obvious fix wrong.
- **Record a decision on #465's two follow-on asks** (sealing `DerivedPropertyChangeHandler`, and the chain-ordering assertion) rather than leaving them unaddressed. My read is that both can be dropped now that the `IsWritten` gate makes the decision order-independent, but that should be written down somewhere rather than inferred from their absence.

RicoSuter:
## Review: correctness, regressions, performance

**Verdict: sound. Ready to merge.** One optional follow-up nit at the end, nothing blocking.

Reviewed `63e08e77` in a clean worktree against `origin/master` (`8a87c39a`). Build: 0 warnings, 0 errors. Unit suite (`--filter "Category!=Integration"`): **2,456 tests green** across 15 assemblies, including the `PublicApi` snapshots, so the "no public API added or removed" claim holds.

### The new tests are load-bearing

Reverting only the four production files to `origin/master` and leaving the tests in place fails **23 tests**. The coverage pins real behavior rather than restating what already worked.

### One of the fixed bugs is more serious than the description reads

`DerivedPropertyRecorder.IsRecording` is `_depth > 0`, so master's `DiscardActiveRecording()` could not distinguish "I still own an unfinished frame" from "an outer evaluation owns one". When a getter triggers a reentrant derived evaluation, the inner `finally` popped and cleared the **outer** frame. Because `_recorder` is `[ThreadStatic]` and thread-pool threads are reused, the damage outlives the operation that caused it.

I measured the blast radius. `WhenSourcePropertyChanges_ThenDerivedPropertyFiresPropertyChanged` is a pre-existing test this PR does not modify. Against master's production code it **passes alone**, but **fails** when the new reentrancy test runs alongside it:

```
Failed ...WhenSourcePropertyChanges_ThenDerivedPropertyFiresPropertyChanged
  Assert.Contains() Failure: Item not found in collection      // "FullName" never fired
```

So on master a single reentrant getter silently stops derived change notifications for every later evaluation on that thread. `ownsActiveRecording` is the right fix, and it deserves a line in the PR body: "reentrant evaluation" undersells a thread-static corruption with cross-operation reach.

### The pooled pending dictionary had two owners

Master returned `_pendingChanges` to `PendingChangesPool` from `Dispose` while the field still pointed at it, so a racing `TryCaptureChange`, `GetPendingChanges`, or `CommitAsync` could read or mutate **another transaction's** buffer. Nulling the field under `_pendingChangesLock` and handing ownership to whichever of `Dispose`/`EndCommit` observes the other makes the return exactly-once, and `CommitAsync`'s new `_pendingChanges is null` branch returns instead of committing a foreign snapshot.

### The design answers the objections #467 raised

#467 warned that deciding on `context.IsWritten` alone reintroduces the uncommitted-publish defect (it answers whether the *trigger* landed, not where the cascade's *inputs* come from), and that narrowing suppression to a non-disposed transaction reopens the rollback case, because `Dispose` sets the disposed flag before it releases the pending buffer. This PR does not trade one for the other:

- `IsWritten` gates the cascade (did the trigger land),
- the recorder-scoped read bypass gates the cascade's inputs (`SubjectTransactionInterceptor.ReadProperty` skips the pending overlay while `IsRecordingDerivedProperty`), and
- `TryGetPendingValue` re-checks `IsDisposed` **inside** `_pendingChangesLock`, so a disposed transaction can never feed a getter even inside the window #467 named.

### `GetFinalOrigin` / `FinalizeOrigin` split is equivalent

Checked line by line, because it sits on the terminal write path of every write. The only behavioral difference is that `FinalizeOrigin` now assigns `_attempted = default` in the already-Local case where master returned early, and that is a no-op: `PendingOrigin.TryConsume` leaves `_attempted` at `default` when nothing was stamped, `ChangeOrigin.Source` is null whenever `Kind` is Local, and `SentValue` is only read behind the non-Local branch. `WhenFinalOriginIsResolvedBeforeWrite_ThenTheAttemptedOriginRemainsAvailableToTheTerminal` pins the property the fall-through depends on: resolving the origin no longer consumes it, so a write that loses the capture race still finalizes normally at the terminal.

Resolving the origin and the timestamp before taking `_pendingChangesLock` is also a real contention win, since the origin comparison can run user equality code.

### Regression risk for consumers: low

The `IsWritten` guard changes nothing for any built-in interceptor:

- `ValidationInterceptor` is `[RunsBefore(SubjectTransactionInterceptor)]`, so it is outer of the derived handler, and it throws rather than vetoing.
- `PropertyValueEqualityCheckHandler` is `[RunsFirst]`, so its equal-value short-circuit already bypassed the derived handler entirely.
- `LifecycleInterceptor` and `PropertyChangeInterceptor` always call `next`, and `PropertyChangeInterceptor` already carried the same `IsWritten` guard.

That leaves only user-supplied interceptors placed inner of `DerivedPropertyChangeHandler` that silently skip `next`, or that forward a **copy** of the context rather than forwarding by ref. Both now stop cascading, which is the intended fix, and the `IWriteInterceptor` doc update calls the copy case out explicitly.

Two semantics worth knowing about, both deliberate and both tested:

- Inside a transaction, a direct read of a derived property still returns pending values while its cached and announced value comes from the landed model. Those can disagree until commit.
- Within a getter, reads bypass the pending buffer but the getter's own writes are still captured into it, so a getter that writes and then reads the same property sees the pre-write model value. `WhenAGetterStartsATransaction_ThenItsStoredValueStillUsesTheModelView` pins exactly this.

### Performance and allocations

**No new allocations, on any path.** `GetServices<SubjectTransactionInterceptor>()` hits the per-`ContextState` service cache and returns a boxed `ImmutableArray<T>`; unboxing copies one reference field. It is strictly cheaper than master's `TryGetService`, which calls `GetServices` and then switches on the length, so #466's "measured marginally cheaper" holds. `ContainsByReference` is `NoInlining` and only reached when `Length != 1`. The nullable `_pendingChanges` adds nothing, and `GetPendingChanges()` returns `Array.Empty` when detached.

Measured with `scripts/benchmark.ps1 -LaunchCount 3 -BaseBranch origin/master`, CPU pinned via sysfs to 3.6 GHz with turbo off and the machine otherwise idle (the BDN header's `Max:` field is not a reliable pin check and disagrees between the arms here).

| Row | master | this PR | Δ | reachable? |
|---|---:|---:|---:|---|
| `ServiceOrderResolverBenchmark.LinearChain` | 1,798.99 ns | 1,782.52 ns | -0.92% | no (subject-free control) |
| `RegistryBenchmark.WriteNoOp` | 342.51 ns | 350.06 ns | +2.20% | no (equality check short-circuits first) |
| `RegistryBenchmark.Read` | 375.01 ns | 384.63 ns | +2.57% | no (read path untouched here) |
| `RegistryBenchmark.ReadParents` | 0.3368 ns | 0.3394 ns | +0.77% | no |
| `RegistryBenchmark.Write` | 1,028.42 ns | 1,036.64 ns | +0.80% | yes |
| `RegistryBenchmark.WriteWithTimestampScope` | 900.03 ns | 882.86 ns | -1.91% | yes |
| `RegistryBenchmark.DerivedAverage` | 241.91 ns | 241.23 ns | -0.28% | yes |
| `RegistryBenchmark.IncrementDerivedAverage` | 5,023.08 ns | 5,063.91 ns | +0.81% | yes |

**Allocated, Gen0 and Gen1 are byte-identical on every row** (3720 B, 140 B, 0 B elsewhere).

Reading it the way `docs/benchmarking.md` prescribes: the four rows the change cannot reach span **-0.92% to +2.57%**, which is this run's own noise floor. Every reachable row sits inside that band, and the largest reachable delta (-1.91%) is smaller in magnitude than the largest unreachable one (+2.57%). No regression is visible above the floor.

A separate run over the same two commits covering `SubjectTransactionBenchmark.CommitChanges` (all four modes) and `SubjectHierarchyBenchmark.RootOnlySet` / `DerivedDeclaredSet` / `DerivedDeclaredGet` agrees: identical allocations on every row (5072 / 12697 / 12697 / 43073 B), timings from -4.6% to +1.8%, sign-mixed, against a +0.39% reference.

I could not reproduce the "unrelated ambient cascades improved by about 16.5%" figure from the standard suite, since no benchmark holds a transaction open across a cascade. The mechanism is real (the cascade path no longer does an `AsyncLocal` read when a transaction is open elsewhere) but the number needs whatever ad-hoc harness produced it to be reproducible.

### Nit (optional, non-blocking)

`TryGetPendingValue` checks `IsDisposed` **before** `ThrowIfCommittingConcurrently()` (`SubjectTransaction.cs:105` vs `:111`), but `TryCaptureChange` checks it **after** (`:141` vs `:143`). A transaction that is both disposed and committing therefore lets reads fall through gracefully while making writes throw `InvalidOperationException` out of a property setter, which is the failure mode #466 names as the wrong one ("an exception escaping one on a scheduler thread or a bare thread takes the process down"). The window is narrow, since the interceptor's `IsCommitting: false` guard filters most of it, but hoisting the `IsDisposed` check above `ThrowIfCommittingConcurrently()` in `TryCaptureChange` would make disposal win consistently and let the write land, matching the fall-through this PR just generalized.

### Not covered

Connector integration suites were not run. This PR does not touch connector code and all connector unit suites are green (Connectors 633, OPC UA 262, MQTT 64, WebSocket 97), but it does change the shared write path that they sit on, and CI's path filters skip those integration suites for shared-library changes.

RicoSuter:
Review follow-up disposition

Applied from the two reviews:

1. Added the transaction-free reentrancy blast-radius regression, including mutation proof against the old recorder cleanup.
2. Restored the Phase 1, Phase 2, and Phase 3 code markers so the design document and implementation remain aligned.
3. Restored the retryable-failure and deferred exclusive-lock ownership rationale.
4. Made disposal win over concurrent-commit validation in TryCaptureChange, with deterministic coverage that a still-live commit continues to throw.
5. Clarified the user documentation: per-flow reads are pending-aware, shared tracking state uses the landed model, and settable-derived writes survive rollback and disposal.

The exact getter-count note was informational; the test remains unchanged and serialized under the existing collection convention.

The proposed DerivedPropertyChangeHandler sealing and interceptor-order assertion were intentionally not added. The context.IsWritten gate makes capture suppression depend on whether the write landed, and recorder-active model reads prevent uncommitted inputs from being published. The decision is therefore order-independent, and the additional restrictions would not protect a remaining invariant.

Tracker disposition:

- #465 remains closed as superseded by #467, and the PR body records that its concrete reproduction is fixed here.
- #466 remains open and the PR continues to say Addresses because BeginTransactionAsync still has the broader multiple-resolution question.
- #467 remains Fixes.
- #471 now tracks the separate settable-derived transaction design, including the cascade-republication caution.

A later unbiased whole-PR review also found two transaction terminal-boundary defects not covered by the original comments. Commit e05b8b05 fixes them with deterministic RED/GREEN and mutation-sensitive tests: post-terminal writes can no longer enter a dead buffer, and unrelated access cannot escape a frozen asynchronous commit snapshot. The scoped re-review approved the correction with no Critical, Important, or Minor findings.

RicoSuter:
## Final review disposition

All actionable review feedback has now been applied where it improves correctness or clarity:

- Replaced the probabilistic disposal/capture loop with one deterministic public-path test gated in user equality code, without a production test seam.
- Added deterministic end-to-end coverage for the documented disposal-during-external-writer behavior: the raw write lands, then frozen replay may overwrite it.
- Corrected the transaction overview and derived-tracking design pseudocode.
- Tightened the PR contract wording to distinguish access carrying the live ambient transaction from ordinary raw access after disposal.
- Replaced the standalone performance percentage with the linked, reproducible benchmark evidence and measured noise interpretation.

Suggestions that would broaden the contract or add redundant restrictions remain intentionally out of scope, as recorded in the earlier disposition comment.

Final exact head `34a513bd` was reviewed independently twice against `master`; both reviews reported no Critical, Important, or Minor findings. Local verification passed with a clean build and 3,103 non-integration tests, including 474 tracking tests. Final GitHub checks are green.

**Ready to merge.**

RicoSuter:
## Re-review at `e05b8b05`: the terminal-boundary work

Reviewed the delta since the last review, `63e08e77..e05b8b05`, in a clean worktree.

**Verdict: ready to merge.** No correctness or threading defect found in either new commit. One documentation gap worth closing first, described under Findings.

### `4e187a40` is exactly the review dispositions

Phase 1/2/3 markers restored so the design document and the code agree again; the two load-bearing why-comments restored in shortened form; and `TryCaptureChange` now checks disposal before `ThrowIfCommittingConcurrently()`. That last ordering is now consistent across all three lock-guarded methods, `TryGetPendingValue`, `TryCaptureChange`, and the new recheck, so disposal wins over committing everywhere rather than in one path.

The reentrancy blast-radius test is there and is the right shape: a reentrant getter runs, then an unrelated `Person` on the same thread still tracks `FullName`. It is transaction-free, so it fails on the recorder-ownership regression directly rather than through the transaction machinery. That was my one "optional before merge" item and it is closed.

### `e05b8b05` fixes two real defects

**1. Post-terminal writes entering a dead buffer.** After `FinishCommit` cleared the pending dictionary, a later write on the same flow was still captured into that cleared dictionary, so it was silently swallowed: never applied to the model, never committed. The fix gates on `_isCommitted` (and a null buffer) in both the interceptor and the lock-guarded capture.

**2. Access escaping a frozen commit snapshot.** While `CommitAsync` is in flight, property access on a flow carrying that transaction previously fell through to the model, which is mid-mutation and whose snapshot is frozen. It is now rejected, except for the commit's own synchronous model replay, authorized through a `[ThreadStatic]` scope.

I checked the parts most likely to go wrong:

- Both `_isCommitting` and `_isCommitted` are `volatile bool`, so the interceptor's lock-free fast-path reads are properly ordered. A stale read is conservative in the safe direction: it can only send the caller down the slower path, which rechecks under the lock.
- The `CommitModelAccessScope` never spans an `await`. In all three uses the scope closes before the next await, which matters because the authorization is thread-static and a continuation can resume on a different thread. `RevertSourceWritesSafelyAsync` sits outside the scope in the writer path.
- The new recheck takes `_pendingChangesLock` on the interception path, which is a lock the interceptor did not previously take. No user code ever runs under that lock, and the derived handler evaluates getters outside `lock(data)` by design, so there is no lock-ordering inversion in either direction.
- The hot path is unchanged: everything still sits behind `if (!SubjectTransaction.HasActiveTransaction) return next(...)`.

### Verification

- Full non-integration suite: **3,102 passed across 26 projects, 0 failed**.
- **Mutation check.** Reverting only the three production files to `63e08e77` and keeping the new tests fails **7 tests**, including all three terminal tests, both blocked-writer tests, and the disposal-during-commit test from `4e187a40`. The new tests are load-bearing and the behavior genuinely changed.
- **Connector suites run unfiltered**, which closes the gap I flagged last time, since CI's path filters skip these for shared-library changes: Connectors 633, OPC UA 295, MQTT 75, WebSocket 131. All green, integration tests included.
- **Defect injection** on the new mechanisms: making the commit authorization always-true fails 3 tests, always-false fails 33 (so the commit's own replay genuinely depends on it), and making the recheck never reject fails 3. Removing the write-side terminal guard from **both** places it is enforced fails the 3 terminal tests. Single-point removal stays green because the interceptor check and the capture check each suffice alone, which is worth knowing when reading mutation results on this file.

### Performance: no regression

`e05b8b05` vs `63e08e77`, `-LaunchCount 3`, memory randomization on, CPU pinned via sysfs (governor `performance`, min and max both 3.6 GHz, `no_turbo=1`; the BDN header's `Max:` line disagrees between arms and is the unreliable-header artifact).

| Row | 63e08e77 | e05b8b05 | Δ | Allocated |
|---|---:|---:|---:|---|
| `ServiceOrderResolver.LinearChain` (noise reference, unreachable) | 1.750 μs | 1.766 μs | +0.91% | identical |
| `CommitChanges` Local | 33.613 μs | 33.401 μs | -0.63% | identical |
| `CommitChanges` SingleSource | 41.617 μs | 41.343 μs | -0.66% | identical |
| `CommitChanges` SingleSource+Local | 39.588 μs | 39.465 μs | -0.31% | identical |
| `CommitChanges` MultiSource | 46.549 μs | 46.205 μs | -0.74% | identical |

Allocations, Gen0 and Gen1 are byte-identical on every row. Every reachable row moved *faster* by less than the unreachable reference moved slower, so nothing clears the floor. That matches the mechanism: one extra volatile read per intercepted access while a transaction is open, and three thread-static writes per commit.

### Findings

**1. `ITransactionWriter` does not document the new constraint.** This is the one I would fix before merge.

`WhenWriterAccessesAmbientPropertyDuringCommit_ThenCommitReportsConcurrentAccess` pins the behavior: a writer that merely *reads* one subject property during `WriteToSourcesAsync` fails the entire commit with `SubjectTransactionException`, inner `InvalidOperationException` saying "commit is in progress", and the transaction is terminal with nothing applied. The interface remarks are otherwise exhaustive, down to slot marking and pooled-buffer ownership, so the omission stands out. The message the implementer sees, "The transaction writer threw an exception or violated its contract during commit", does not point at the actual rule.

The built-in `SourceTransactionWriter` is unaffected: it only calls `change.Property.TryGetSource(...)`, which is a property-data lookup and not a getter through the interception chain. So nothing in this repository breaks. The exposure is entirely to custom writers, and a sentence in the `WriteToSourcesAsync` and `RevertSourceWritesAsync` remarks saying to use the values in `changes` and not to touch subject properties on this flow would close it.

**2. The rejection is stricter than the writer phase requires.** In `CommitWithWriterAsync`, `WriteToSourcesAsync` runs *before* `ApplyLocalChanges`, so at that point the local model has not been mutated at all and a read would be perfectly coherent. Rejecting it is conservative rather than necessary. A uniform rule is easier to document and reason about than a phase-dependent one, so I would keep it, but it is worth recording as a deliberate choice rather than an accident.

**3. The read-side terminal guard has no observable effect today.** Removing `_isCommitted` from both `TryGetPendingValue` and the interceptor's read fast path leaves the whole suite green. That is not a missing test: `_isCommitted` is only ever set together with a cleared pending dictionary, on both the success and terminal-failure paths, so a post-terminal read already falls through to the model on its own. It is defense in depth against a future path that sets the flag without clearing the buffer. The write-side guard, by contrast, is genuinely load-bearing.

### Behavior changes for the release notes

The first one is the one consumers are most likely to hit, and it is a change in kind rather than degree.

1. **A property getter can now throw.** While a commit is in progress, a read on a flow carrying that transaction throws `InvalidOperationException` instead of silently returning the landed model. Same for writes, including settable-derived writes. Access on a flow without the ambient transaction is unaffected and still reads the model.
2. **A committed or terminally failed transaction is inactive for interception before disposal.** Later reads and writes use the model normally. Retryable failures stay active and keep capturing.

### Aside, not about this PR

`scripts/benchmark.ps1` left my worktree checked out on the base commit after the run, because the restore step cannot put back a detached HEAD. `docs/benchmarking.md` warns that the script checks the base out in your working tree, but not that it may not put it back. I re-ran the affected step after restoring; the benchmark numbers above are from a run where both arms were measured correctly.

RicoSuter:
Final transaction-writer documentation follow-up

Applied the review clarification in commit 4518f57d. The transaction guide and ITransactionWriter XML now explicitly state that both reads and writes are forbidden from WriteToSourcesAsync and RevertSourceWritesAsync while the live ambient transaction is committing. Getter reads are called out because sibling or landed-model state is outside the frozen snapshot. Writers must use the supplied snapshot, writer-owned configuration, and any required subject state captured before CommitAsync.

No runtime behavior or API signature changed. The existing writer-read regression passes, the Tracking project builds with 0 warnings/errors, two independent reviews reported no findings, and all final GitHub checks are green.

PR #469, opened 2026-08-16
Add scheduled property subscriptions

RicoSuter:
## Summary

- Rename writer-thread per-property APIs to `SubscribeInline` and `SubscribeToPropertyInline` so delivery behavior is explicit. This is an intentional API break.
- Add scheduled per-property subscriptions with serialized delivery per subscription, isolated observer and scheduler failures, bounded drain batches, and a named handle exposing `PendingCount` and `IsFaulted`.
- Add deterministic scheduler and protocol coverage for enqueue/settle races, terminal ownership, disposal, faults, reentrant delivery, transactions, `ExecutionContext`, and custom inline schedulers.
- Add concise consumer documentation and a bounded end-to-end scheduled-delivery benchmark for future local comparisons.

## Migration

The intentional API break is approved. Existing writer-thread calls migrate as follows:

- `PropertyReference.Subscribe(observer)` becomes `PropertyReference.SubscribeInline(observer)`.
- `subject.SubscribeToProperty(selector, observer)` becomes `subject.SubscribeToPropertyInline(selector, observer)`.
- The `Subscribe` and `SubscribeToProperty` names now require a scheduler and provide scheduled delivery.

## Contract notes

- Serialization is per subscription. An observer shared by multiple subscriptions may still be invoked concurrently.
- The queue is unbounded. `PendingCount` is exact after writes and deliveries become quiescent.
- Disposal drops queued work. A delivery already in flight may still invoke the observer or finish after disposal returns.
- The built-in immediate and current-thread scheduler singletons are rejected. A custom scheduler can still execute inline, in which case callback latency and current ambient state remain on the writer thread.
- Automatic writer `ExecutionContext` capture is suppressed for asynchronous scheduling. Existing ambient state on a scheduler-owned worker is not cleared.
- Observer failures call `onError` on the scheduler execution thread. A synchronous `IScheduler.Schedule` failure calls it on the calling thread, which can be the writer thread. Shared handlers must be thread-safe, and a slow synchronous failure handler delays the setter.
- Scheduling failures are reported through `onError`. `IsFaulted` is true only when the fault transition wins over concurrent or reentrant disposal.

## Verification

- Full solution build: 0 warnings, 0 errors.
- Non-integration solution suite: 3,140 passed, 0 failed, 0 skipped.
- Tracking suite: 511 passed.
- Focused scheduled and public API verification: 33 passed after the final documentation and XML clarification.
- Deterministic custom-inline regression reproduced the old nested schedule depth of 3 and verifies final depth 1 with all 2,059 changes delivered and settled.
- Benchmark project builds cleanly. No BenchmarkDotNet comparison was run because this PR makes no quantitative performance claim.
- Independent correctness, contract, and stack-integration reviews found no blocking defect. The newest deep-threading review found no bug; its optional production seam and expanded stress-suite suggestions were intentionally not added.

Depends on #468.


RicoSuter:
## Review: 469 vs 468 only

Reviewed `f851ed3a` against `63e08e77` (PR #468 head) in a clean worktree. #468 itself was not re-reviewed.

**Verdict: ready to merge**, after the one item under "Needs a decision" below. No correctness defect found, no performance or allocation regression for existing consumers.

### Scope

The diff touches four files in shipped libraries. Two are XML doc only (`IPropertyChangeObserver`, `PropertyChangeCallback`), one is the rename plus the four new overloads (`PropertyChangeSubscriptionExtensions`), one is the new type (`ScheduledPropertySubscription`). `PropertyChangeInterceptor`, `PropertyChangeSubscription`, and the core assembly are untouched. Everything else in the diff is tests, docs, and the benchmark. `System.Reactive` was already a `Tracking` dependency, so `IScheduler` in the public API adds no new package reference.

### Correctness

I worked through the two counters independently of the tests and they hold up.

`_wip` is a work-in-progress counter, not just a gate. `Enqueue` publishes to the queue before incrementing, and only the `0 -> 1` transition schedules. `_wip` decreases only in the drain's `finally`, so it can reach zero only after a drain's loop has ended and all its `Deliver` calls have returned. That gives the two properties the design needs:

- **At most one drain loop runs at a time**, so per subscription delivery is serialized. The `0 -> 1` enqueue path and the `finally` reschedule path are mutually exclusive: when the `finally` subtract lands on zero it does not reschedule and the next enqueue does, and when it lands non-zero the concurrent enqueue saw a non-zero counter and did not.
- **No lost wakeup.** Every accepted change either belongs to a drain that is already scheduled or triggers a schedule.

`_scheduleRequestCount` is the second trampoline (886a44ef), and it caps at 2: the in-progress `ScheduleDrain` loop holds one count, the drain it submitted can request at most one more. `observed` is therefore always 1 when it re-enters, so the loop never coalesces two requests into one submission. That is what keeps an inline scheduler at constant stack depth instead of recursing, which `WhenAnAcceptedInlineSchedulerDrainsMultipleBatches_ThenScheduleCallsDoNotNest` pins at `MaximumScheduleDepth == 1` over 2,059 changes and three batches.

The two snapshot-ordering comments are correct, not just plausible:

- `Deliver` reads `_onError` then `_observer`, and `TransitionOutOfLive` clears `_observer` then `_onError`. There is no interleaving that yields a live observer paired with a null handler.
- `SubmitDrain` reads `_onError` before the state check (f851ed3a), so disposal landing after the check cannot erase the handler a scheduler failure needs.

Reentrancy is handled without a special case: an observer that writes the same property during delivery enqueues while `_wip >= 1`, so it does not schedule and the successor is picked up by the running drain. `MaxBatch = 1024` bounds a drain so a hot property cannot pin a scheduler thread indefinitely.

Two things I confirmed are deliberate rather than bugs. A drain that returns early on a terminal state leaves an un-subtracted `_wip` residue forever, which is inert because the subscription is terminal and nothing but the test hook reads it. `Dispose` after a fault is a no-op because the fault already ran the full cleanup, including the upstream disposal and the process-wide count decrement, which `WhenTheSchedulerThrows_ThenItIsReportedAndDoesNotReachTheWriter` asserts.

The test suite is genuinely adversarial rather than illustrative: the deterministic schedulers make the interleavings chosen rather than raced, `RunUntilIdle` has a budget so a self-sustaining reschedule loop fails red instead of hanging, and `RecordingScheduler` catches escapes that would otherwise kill the host silently. `WhenManyWritersRaceOneProperty` probes reentrancy across 16,000 concurrent writes and asserts settle to zero.

### Verification I ran

- Full solution build: 0 warnings, 0 errors.
- `dotnet test src/Namotion.Interceptor.slnx --filter "Category!=Integration"`: 2,807 passed, 0 failed, 0 skipped, across the 17 assemblies that report results. The other 10 projects in the solution are HomeBlaze and device suites whose non-integration set is empty; exit code was 0 throughout. This does not match the 3,130 across 26 projects in the PR body, which is worth a look, though nothing failed either way.
- Public API snapshot is regenerated and consistent with the new surface.

### Performance and allocations

**No regression is possible for existing consumers, and the measurement agrees.** The write hot path is not in the diff, and a method rename in a static extension class cannot reach it. Local run, `-LocalOnly`, CPU pinned at 3.6 GHz with `no_turbo=1` and the performance governor (the BenchmarkDotNet header reporting `Max: 0.94GHz` is the known unreliable-header case, the OS values are the real ones):

| Method | Mean | Allocated |
|---|---:|---:|
| WriteIdle | 138.6 ns | 0 B |
| WriteWithListenerOnOtherProperty | 142.0 ns | 0 B |
| WriteWithListenerOnSameProperty (inline) | 190.3 ns | 0 B |
| WriteAndWaitForScheduledListenerOnSameProperty | 4,744.8 ns | 120 B |

The idle, listener-elsewhere, and inline rows are still zero allocation and sit where they always have. I deliberately did not run a two-arm comparison against #468: with identical hot-path source, that run measures only code placement, and `docs/benchmarking.md` records an identical-tree run producing a +2.2% median with single rows at +32% and +45%. The diff scope is the stronger evidence here.

**On the new path, read the 4.7 us carefully.** That benchmark waits on an `AutoResetEvent` for the observer to finish, so it is a one-change-at-a-time end-to-end latency dominated by the thread pool handoff, not the cost of the subscription machinery, and it is not a throughput number. Because every write blocks until its delivery completes, batching never engages, so the 120 B/op is close to a worst case: one `IScheduler.Schedule` submission per change plus queue segment growth. Under any real burst a single drain covers up to 1,024 changes, so the submission cost amortizes away and the residual per-change cost is the queued `SubjectPropertyChange` slot. The static `DrainAction` and the `AsyncFlowControl` struct from `SuppressFlow` keep the subscription itself from adding closure or boxing allocations on top of what the scheduler charges.

### Needs a decision before merge

**The rename is a break on an API already published to NuGet.** `PropertyReference.Subscribe(observer)` and `subject.SubscribeToProperty(selector, observer)` shipped in v0.9.0 and v0.9.1 (added by #377). This PR keeps both names and gives them new, scheduler-requiring signatures. Nothing in this repo consumed them outside tests and the benchmark, and the compiler catches every external call site, so nobody gets a silent behavior change. But per AGENTS.md a public API break needs explicit approval, and neither the PR body nor the docs state the migration. Two things would close it:

1. Put the mapping somewhere a consumer will find it: old `Subscribe` becomes `SubscribeInline`, old `SubscribeToProperty` becomes `SubscribeToPropertyInline`, and same-name-new-signature means scheduled delivery.
2. Consider whether reusing the old names for the new semantics is what you want. `SubscribeScheduled` would make the break unambiguous and leave `Subscribe` free. The compiler catches the difference either way, so this is a judgment call, but it is much cheaper to make now than after the next release.

### Behavior notes for consumers, not blocking

- **The queue is unbounded with no backpressure and no overflow signal.** A slow or stalled observer grows the heap without limit, and the only indicator is polling `PendingCount`. `WhenTheSchedulerAcceptsWorkAndNeverRunsIt_ThenTheSubscriptionGoesQuietWithoutReporting` documents the worst version of this: a scheduler that accepts work and never runs it accumulates silently, with no error and `IsFaulted` still false. That is consistent with `CreatePropertyChangeQueueSubscription`, so it is a known house style rather than a surprise, but for an industrial host it is the failure mode most likely to bite. An optional bounded capacity with a drop policy and an overflow callback would be a good follow-up.
- **Suppressing `ExecutionContext` flow also drops `Activity.Current` and any logging scope.** The docs describe this accurately in `ExecutionContext` terms, but a consumer running OpenTelemetry will experience it as orphaned spans in the observer. One sentence naming tracing and logging scopes would save someone a debugging session.
- **The synchronous-scheduler guard is reference equality against two singletons.** Everything else is accepted, including schedulers that create a thread per work item. Worth a caution in the scheduler-choice paragraph.
- **`PendingCount` walks `ConcurrentQueue` segments rather than being O(1).** Fine as a diagnostic, worth a note so nobody polls it in a tight loop against a large backlog.

### Minor docs point

`docs/tracking.md` lost the paragraph describing inter-channel dispatch order: that each interceptor dispatches its queue first, then its Rx observable, then the per-property listeners it resolved, and that under aggregated contexts the innermost instance resolves the listeners so they can run before an outer context's queue and Rx channels. That is consumer-visible ordering and it now lives only in the comments on `PropertyChangeInterceptor`. The same commit also dropped the "Observable features and limitations" and "Queue performance characteristics" lists, which are unrelated to scheduled delivery. Restoring the ordering paragraph would be worth it, the other two are your call.

RicoSuter:
## Deep correctness and threading pass on `ScheduledPropertySubscription`

Breaking change is settled, so this pass was only about whether the new type is correct under concurrency. Four independent methods, all on `f851ed3a`.

**Result: no correctness or threading bug found.** One coverage gap worth knowing about, described at the end. Nothing blocking.

First, a correction to my earlier comment: the full suite is **3,130 passed across 26 projects, 0 failed**, exactly as the PR body says. My lower count came from running with `--no-build` over stale artifacts, so several projects never reported. There is no discrepancy.

### 1. The two counter invariants, proved

`_wip` is a work-in-progress counter, and the proof that matters is that it never reaches zero while a drain is running. It is only decremented in the drain's `finally`, so during a drain's loop it is at least one. That gives the two properties the design needs:

- **At most one drain loop runs at a time.** A `0 -> 1` enqueue cannot happen during a drain, so the only other scheduler is the drain's own `finally`, and the two are mutually exclusive: when the `finally` subtract lands on zero it does not reschedule and the next enqueue does; when it lands non-zero, the concurrent enqueue saw a non-zero counter and did not.
- **No lost wakeup.** Every accepted change either belongs to an already scheduled drain or triggers one.

From that, `_scheduleRequestCount` provably caps at 2, so `observed` is always 1 and the coalescing loop can never merge two requests into one submission.

### 2. Adversarial tests I wrote

Eleven tests beyond the PR's suite, all passing, and **5 consecutive runs of all 46 scheduled tests with zero failures** (no flakiness):

- **Exact FIFO over 50,000 changes** through 49 batch handoffs: every value arrives once, in commit order.
- **Inline-scheduler mutual exclusion under 8 concurrent writers, 32,000 writes.** This one is the direct observation of the invariant: with a scheduler that runs work inline, two concurrent drains would put two threads inside `Schedule` at once, so a shared depth counter would exceed one. It stays at exactly 1.
- **40,000 writes across 8 threads on the pool scheduler**: no loss, no duplicates, no observer reentrancy, settles to zero.
- **300 rounds of disposal racing an active drain and an active writer**, each round asserting complete teardown, observer released, no escape into the scheduler, and the process-wide count back to zero.
- **Subscribe and dispose churn under sustained writes**, plus two independent subscriptions on one property, plus `ExecutionContext` flow restoration on the writer thread after both a successful schedule and a scheduler failure.

### 3. Defect injection: are the tests actually pinning the protocol?

Twelve injections, each rebuilt and run against the full tracking suite. The load-bearing elements are all pinned:

| Injection | Result |
|---|---|
| Remove the `ScheduleDrain` coalescing trampoline | RED (3 tests) |
| Schedule on every enqueue instead of only `0 -> 1` | RED (10 tests) |
| Never reschedule after a batch | RED (5 tests) |
| Drop `ExecutionContext` suppression | RED (2 tests) |
| Do not dispose the upstream on the terminal transition | RED (7 tests) |

Two injections stayed green for a good reason rather than a bad one: removing the terminal-state check in `Enqueue`, and removing the mid-loop terminal check in `Drain`. Both are provably redundant, because after disposal `_queue` is null and `Deliver` also null-checks `_observer`. They are belt and braces, and it is worth knowing that the null checks, not the state checks, are what actually stop delivery after disposal.

### 4. Race widening: a scheduling delay at every interleaving point

The strongest evidence, given no model checker was available here. I inserted a `Thread.Yield()` at each of **13 interleaving points** in the protocol, one at a time, which turns a window that is normally a few instructions wide into one another thread can land in, then re-ran the invariant suite. Points covered: both `Enqueue` windows, three `Drain` windows including between dequeue and delivery and before the `finally` subtract, the `ScheduleDrain` submit-then-consume window, both `SubmitDrain` windows, the `Deliver` snapshot window, all three `TransitionOutOfLive` windows, and the `Create` window.

**All 13 stayed green.** Exact FIFO, no loss, no duplicates, no reentrancy, settle to zero, complete teardown, and flow restoration all held at every widened point.

### 5. The create-time guard, verified deterministically

The window between `SubscribeInline` returning and the `Interlocked.Exchange` in `Create` is too narrow to hit by racing. Over 400 rounds against a hot writer it never triggered once. So I widened it inside `Create` and made it deterministic:

- With the guard: every round tears down cleanly, count back to zero, no listener data left behind.
- Without the guard: **round 0 leaks the upstream permanently**, count stuck at 1, and it never recovers.

That is the failure mode the `PropertyChangeSubscription` comment warns about, where one leaked subscription keeps every write in the process on the slower listener-check path for the process lifetime. The guard is correct and load-bearing.

### The one coverage gap

Three orderings implement the same property, that **an error report survives a disposal racing it**:

1. `SubmitDrain` reads `_onError` before the state check (this is what `f851ed3a` fixed).
2. `Deliver` reads `_onError` before `_observer`.
3. `TransitionOutOfLive` clears `_observer` before `_onError`.

Swapping any of them leaves the entire suite green. I confirmed they are not merely decorative by swapping each and widening the window the swap creates: the scheduler failure and the observer exception are both silently dropped, while the pristine ordering reports them. So the code is right and the comments are accurate, but **nothing pins any of the three**, and a future refactor could undo `f851ed3a` without a single test going red.

I also checked whether a regression test is writable from outside the type. It is not: on the correct ordering there is no window to aim at, and within one write `SubmitDrain` runs to completion inside a single listener callback, so no other listener can interleave. Pinning this would need an internal seam. Given the consequence is a dropped error report during teardown, leaving it as a review-time invariant is defensible. Worth a decision rather than a silent omission.

Two smaller notes from reading the same paths:

- Queued changes are safe to hold indefinitely. `SubjectPropertyChange` is a readonly struct whose boxed value holders are allocated per change with no pooling anywhere in `Change/Performance`, so nothing can mutate underneath a queued change.
- `SuppressFlow` is entered inside the `try` with the catch outside it, so the writer thread's flow is restored even when `Schedule` throws. A leak there would have been nasty, and it is covered.

Happy to open a PR against this branch with the eleven tests if you want them, particularly the exact-FIFO one and the inline-scheduler depth one, since that pair pins ordering and mutual exclusion directly rather than by proxy.

RicoSuter:
## Final review disposition

All actionable review feedback has now been applied where it improves the consumer contract:

- The intentional API break is approved and the exact migration mapping is in the PR description.
- The consumer-visible synchronous dispatch-order paragraph was restored.
- XML and consumer docs now describe inline and scheduled per-property channels accurately.
- `onError` threading is explicit: observer failures report on the scheduler execution thread, while a synchronous `IScheduler.Schedule` failure reports on the calling thread and can delay the setter.
- PR2 was synced onto the exact final PR1 head.

The newest deep-threading review found no correctness or threading bug. Its optional internal timing seam and eleven additional stress tests were intentionally not added. The three ordering invariants are correct by the paired snapshot/clear proof and existing deterministic protocol tests; adding production hooks solely to pin unobservable instruction windows would increase complexity without fixing a current defect. Bounded queues and overflow policy likewise remain separate feature decisions, while the current unbounded semantics are documented.

Final exact range `34a513bd...7760a62f` was reviewed independently twice; both reviews reported no Critical, Important, or Minor findings. Local verification passed with a clean build and 3,140 non-integration tests, including 511 tracking tests.

**Ready to merge after final GitHub checks complete.**

RicoSuter:
Final stack synchronization

PR2 is now stacked on exact PR1 head 4518f57d at head 7a9f9ce0. Two independent reviewers verified that the restacked PR2 binary diff is identical to the previously approved 18-file package, with no PR1 files leaking into the PR2 diff and no documentation conflict. All final GitHub checks are green.

Issue #462, opened 2026-08-14
ChangeQueueProcessor drops its buffered changes on shutdown without counting them


RicoSuter:
`ChangeQueueProcessor.ProcessAsync` never flushes what it has buffered when it stops. On the buffered path changes accumulate in `_changes` and are written only by the periodic timer, and the shutdown sequence cancels that timer instead of draining it:

```csharp
finally
{
    try { await linkedTokenSource.CancelAsync().ConfigureAwait(false); } catch { }
    await flushTask.ConfigureAwait(false);
}
```

The flush loop's `WaitForNextTickAsync(linkedTokenSource.Token)` throws, the `catch` swallows it as an expected cancellation, and the buffer is dropped. Since the same linked token is passed to `_writeHandler`, a flush already in progress can also abort part-way through its batch.

The window is up to one buffer interval, 8ms by default, plus whatever was in flight. It is silent: nothing is logged and nothing is counted, unlike the bounded-queue overflow path which maintains `DroppedChangeCount`.

## Whether this is worth changing

Losing changes when a process stops is a normal thing for a connector to do, and the window is small, so this may well be working as intended. Three things argue the other way:

- This is `StopAsync`, a graceful stop, not a crash. A caller that awaits it has some reason to expect the queue to be handed over rather than dropped.
- The behaviour depends on configuration. With `bufferTime` at zero the immediate path writes each change as it is dequeued, so nothing is buffered to lose. Only the batching configuration drops data, which is the one a throughput-minded consumer picks.
- The connect window next door goes to considerable trouble over exactly this loss, parking writes into the retry queue and reconciling them. The shutdown path has no equivalent.

## What a fix would have to weigh

A final flush needs its own token rather than the linked one, since the linked one is already cancelled by then, plus a bounded timeout. The cost is that shutdown can now block: an unreachable source turns every stop into a wait for that timeout, which is worse than losing 8ms of changes for most consumers. So a fix probably wants a deliberate opt-in and a documented bound, not an unconditional drain.

At minimum, counting and logging what was dropped would make the loss visible instead of silent.

RicoSuter:
#481 fixes the first of the two losses named here and leaves the second, so this should stay open rather than close with it.

[...]

Issue #362, opened 2026-07-08
Connectors: internal reconnects skip the write retry-queue reconcile (source-wins divergence not enforced)


RicoSuter:
## Summary

The write retry-queue reconcile runs only at **base-loop** reconnects, never at **connector-internal** reconnects. During a transport auto-reconnect handled inside a connector (MQTT managed-client reconnect, OPC UA SDK reconnect, WebSocket reconnect), local writes that were captured and queued during the outage are re-loaded past but not reconciled against the freshly reloaded source state. They are then flushed blindly on the next write, so the "source changed during the outage, source wins" drop decision is skipped.

This was surfaced while reviewing #355. It is **pre-existing on master**, not introduced by that PR.

## Where

The reconcile (`SubjectSourceBase.ReconcileRetryQueueAsync`, previously `ReapplyRetryQueue`) is invoked from exactly one place: `SubjectSourceBase.ExecuteAsync`, immediately before `ProcessAsync`, after `DrainOwnedWritesToRetryQueue`. The base retry loop is where drain plus reconcile plus a fresh processor run sequentially on the pump task.

The connectors, however, reconnect on their own concurrent monitor / health-check task and call `LoadInitialStateAndResumeAsync` directly, without any reconcile:

- `MqttSubjectClientSource.OnReconnectedAsync` (calls `LoadInitialStateAndResumeAsync`)
- `WebSocketSubjectClientSource.ReconnectAndResumeAsync` (calls `LoadInitialStateAndResumeAsync`)
- `OpcUaSubjectClientSource` restart path and `SessionManager.PerformFullStateSyncIfNeededAsync` (both call `LoadInitialStateAndResumeAsync`)

Each connector's `StartListeningAsync` returns a `BackgroundTaskLifetime` that runs its monitor loop, and the base then runs `ProcessAsync` for the whole connected lifetime. So the connector-internal reload happens **concurrently with the running `ProcessAsync`**.

## Impact

- Writes are **queued, not lost** (bounded, drop-oldest `WriteRetryQueue`), and they do reach the source on the next write.
- The only defect: for connector-internal reconnects, the 3-way reconcile decision (send if current, restore if at baseline, **drop if the source diverged**) is not applied. A stale local write based on a pre-outage baseline can overwrite a source value that changed during the outage, violating the documented source-wins guarantee for that case.
[...]

Issue #352, opened 2026-06-21
ChangeQueueProcessor: model queue overflow as an explicit policy with an observable drop signal


RicoSuter:
## Background

`ChangeQueueProcessor` has an opt-in `maxQueueDepth` bound (added in #350). When the buffered queue overflows, the oldest unprocessed change is silently dropped and `DropCount` is incremented. That is adequate for a passive history log (a bounded gap is acceptable), but it is not a complete design for sync connectors, where a dropped change is a downstream divergence that should trigger a resync.

A silent `DropCount` is observability, not a policy. The "what do I do about an overflow" half is missing.

## Proposal

Model overflow as an explicit, visible policy on `ChangeQueueProcessor`:

- An overflow policy: `Unbounded` (default), `DropOldest`, `DropNewest`.
- An overflow signal (callback or observable) so the consumer can react: a history store accepts the gap, a sync connector triggers a resync, an operator gets an alarm.
- Consider bounding on a more meaningful unit than raw pre-dedup change count. The queue coalesces by property at flush, so a burst on a single property inflates the raw count even though it dedups to one write. Options: distinct pending properties, or estimated bytes.

## Consumers

- History stores (Phase A): `DropOldest`, accept the gap.
- Sync connectors (OPC UA / MQTT / WebSocket): on overflow, trigger a full resync rather than accept silent loss. This is why exposing a raw drop knob on the connector configurations was reverted in #350; the safe connector behavior needs this policy plus a resync.

## Out of scope (done in #350)

#350 keeps only the low-level opt-in `maxQueueDepth` on `ChangeQueueProcessor` and removes the `MaxQueueDepth` config properties from the connectors. This issue tracks the proper overflow-policy design for when a connector or store actually needs bounded behavior.

PR #480, opened 2026-08-19
feat: Stable-ID subject update protocol and ID-resolving pipeline

RicoSuter:
Replaces the index-based subject-update protocol with a stable-ID one. Subjects carry a stable base62 ID, and structural collection and dictionary updates now send complete state as `items` referencing subjects by `id` (plus `key` for dictionary entries) instead of positional insert, remove and move operations. This is the first of three stacked PRs that together bring the capability of #197 onto current master; it also lands the lifecycle batch scope the apply path depends on, and bumps the WebSocket wire protocol to version 2.

The work was re-implemented from #197 rather than rebased, because that branch had diverged for six months across exactly the layers it rewrites. The decision paid for itself: the review chain on this PR found three separate defects in the ported code, each of which leaves a receiver holding a registered, rooted, default-valued subject that never converges, and none of which any suite on #197 would have caught, in two cases because that branch had deleted the tests that would have failed.

Related to #197. Supersedes its protocol and pipeline work; that PR closes once all three stack PRs are open and reviewed.

## Why

#197 has been open since February and cannot land as-is. Master reworked interception, lifecycle ordering, change origins and the connector base under it, so a rebase would mean resolving six months of semantic drift in one step, with no green checkpoint until the end. Splitting it into three reviewable PRs gives each layer a chance to prove itself, and lets the parts master has since solved differently be dropped rather than merged.

The three defects found here are the substantive argument. Each one is silent, permanent, and invisible to the counters that exist to catch exactly this:

- The inbound drop tripwire counted the root subject as dropped on every inbound update, because `root` is a mapping hint that never resolves in the receiver's registry. The only production signal for silent drops was rising during healthy operation.
- A partial update whose only structural change was a reorder or a remove marked nothing complete, so `completeSubjectIds` serialized as absent, so the receiver read "everything is complete" and fabricated default-valued subjects for ids it could not resolve. A counted, self-healing gap in one update became an uncounted permanent corruption in the next.
- Deferred attribute updates ran after the retry pass that consumes subject ids, so a subject reachable only through a subject-valued attribute of a newly created subject was counted as dropped, then created and never populated.

## Contract

- A partial update always carries `completeSubjectIds`, empty if nothing in it is complete. Only a complete update omits the field, where absence still means "every subject is complete". A receiver must never create a subject for an id outside that set: doing so materializes a default-valued instance that no later update repairs, because the sender will not resend state it believes the receiver already has.
- Structural updates carry complete state. A collection or dictionary property update describes the full resulting membership, so a receiver applies it by replacement rather than by replaying operations. Collection order is meaningful and comes from array position; dictionary entries are keyed and their order is not.
- Every write the applier makes to the graph carries the update's `ChangeOrigin`, so echo suppression holds and a connector never pushes a value back to the source it just received it from.
- `SubjectUpdateDiagnostics` counts what the pipeline drops, and is the only production signal for it: this protocol deliberately ships no content-divergence detector. Read the counters as a rate, with one exception now known: a model whose subjects hold parent pointers back to the root produces a recurring, not transient, count, because such a reference marks permanent divergence rather than a gap that later state repairs. `SubjectRegistryDiagnostics.DuplicateSubjectIdAttaches` is the companion signal: a nonzero value means something fabricated a subject, which no other counter observes.
- A subject created by an inbound update is populated before it enters the graph, and a second reference to that same subject within the same update resolves to the instance already created rather than making another. A reference back to the update root resolves to the local root. Its initial values therefore bypass the write-interceptor chain: no validation, no derived recalculation, no change events for those values, and a caller-supplied `transformValueBeforeApply` does not run for them. Lifecycle correctness is unaffected, because attach seeds the last-processed values from the backing store.
- `LifecycleInterceptor.CreateBatchScope` is thread-affine. It must be disposed on the thread that created it, and nested scopes must close in reverse order of opening, both of which a `using` block gives you.

## Breaking changes

API, in `Namotion.Interceptor.Connectors`:

- `SubjectCollectionOperation` and `SubjectCollectionOperationType` are deleted, along with `SubjectPropertyUpdate.Operations`. Structural changes now travel as complete state in `Items`. A caller that built operations by hand should send the resulting membership instead.
- `SubjectPropertyItemUpdate.Index` (`required object`) becomes `Id` (`required string`), the item's stable subject id. The previous `Id` (`string?`) becomes `Key` (`string?`), the dictionary key, null for collection items whose order comes from array position.
- `SubjectPropertyUpdate.Count` is removed. It existed to bounds-check index-based operations and has no role once every structural update carries complete membership.
- `SubjectUpdate.Root` becomes `string?`. A partial update need not touch the root subject.
- Added: `SubjectUpdate.CompleteSubjectIds`, and the `SubjectUpdateDiagnostics` static class.
- Added in `Namotion.Interceptor.Tracking`: `LifecycleInterceptor.CreateBatchScope`. See the thread-affinity note under Contract.

Behavior:

- The WebSocket wire protocol goes from version 1 to 2, a hard break with no dual-shape support. A version 1 peer is rejected at the handshake: the server answers `Hello` with `ErrorCode.VersionMismatch` and closes, and the client rejects a mismatched `Welcome`. Both ends must be upgraded together. The protocol has no external consumers yet, which is why this is a break rather than a negotiation.
- The same shape change reaches AspNetCore HTTP payloads, which serialize `SubjectUpdate` from the same types. Structural properties now emit `id` and `key` instead of `index`, and `root` may be absent on a partial update.
- `ContextInheritanceHandler` now keys fallback-context removal on the context-detach flag rather than on a zero reference count. For a subject attached in two independent lifecycle graphs, the old predicate removed neither fallback and the new one removes the fallback for the graph being left. This is a fix, and it is a behavior change outside the protocol story.

Contract:

- A source or transaction writer must build its payload from the changes it is given, never by reading subject properties. That constraint was always in force under transactions and is now stated on `ISubjectSource.WriteChangesAsync` and in the connector guide, where implementers meet it. See #476.

## Performance

Not measured yet. The benchmark comparison is an open gate, listed under Verification. Two things a reviewer should expect it to show: partial-update building is on the connector hot path and this change rewrites it, and stable ids replace small per-update integers with 22-character base62 strings on every subject key and item reference, so payloads grow. The size delta will be recorded alongside the benchmark rather than left to inference.

| Benchmark | Before | After | Delta | Allocated |
|---|---:|---:|---:|---|
| _pending, see Verification_ |  |  |  |  |
| _noise reference_ |  |  |  |  |

## Diff composition

|---|---:|---:|---:|---:|
| Production code | 26 | 1,284 | 983 | +301 |
| Tests and benchmarks | 95 | 4,786 | 2,207 | +2,579 |
| Documentation | 7 | 590 | 218 | +372 |
| Scripts and CI | 2 | 181 | 0 | +181 |
| **Total** | **130** | **6,841** | **3,408** | **+3,433** |

## Verification

- [x] Documentation updated
- [x] Unit tests
- [x] Integration tests
- [ ] Benchmarks
- [ ] Load tests
- [ ] Chaos tests

Unit tests: all 26 assemblies green, including Connectors 779, ConnectorTester 117, Tracking 524, Registry 154. Integration tests: WebSocket 56 of 56, run in isolation. Documentation: `docs/connectors-subject-updates.md` rewritten for the stable-ID protocol, `docs/connectors-websocket.md` updated for version 2, `docs/design/tracking-lifecycle.md` updated for the batch scope, `docs/connector-tester.md` for the new profile.

Benchmarks and the two Connector Tester runs are outstanding and run on separate hardware:

- `pwsh scripts/benchmark.ps1 -Filter "*SubjectUpdateBenchmark*" -LaunchCount 3`, plus serialized payload sizes before and after.
- `websocket-load`, as the new-protocol baseline.
- `websocket-structural`, added in this PR: server-side structural churn with transactions off. No profile in the repo exercised structural mutation before this one, and the pipeline it exercises is what this PR rewrites.

The chaos runs cannot yet serve as the gate for this defect class, and saying so is more useful than claiming otherwise. `StructuralMutator` only appends and removes nodes on a strict-tree model: no cross-parent moves, no re-adds of existing nodes, no reorders, no shared references. None of the phantom-subject defects fixed in this PR could surface under the current profiles, so a green run would certify less than it appears to. Extending the mutator to produce those shapes is a prerequisite for treating the runs as a gate, and is tracked as a follow-up below. What the runs do cover today is add-and-remove churn under connection faults, which is worth having but is not the same claim.

## What a reviewer should look at hardest

1. `SubjectUpdateBuilder.Initialize` seeding `completeSubjectIds` for partial updates, and `SubjectUpdateApplyContext.IsSubjectComplete` as the single point where a wire-level absence becomes permission to create graph state.
2. The applier hand-merge in `SubjectUpdateApplier` and `SubjectItemsUpdateApplier`: master's origin stamping and sent-value survival rule had to survive inside the reference branch's ID-resolving structure. Master's four origin-survival tests pass with unmodified assertions, which is the gate for that.
3. The two-call attribute pass with a start-index cursor in `ApplyUpdate`, and the ordering argument for why it runs both before and after the deferred retry pass.
4. `SubjectUpdateDiagnostics`: four increment sites against four documented meanings, and whether a rising `DroppedInboundSubjectUpdates` is actionable given it mixes subject entries and structural item references.
5. The batch scope's thread-affinity and owner rules in `LifecycleInterceptor`, and the deliberate double-fire of a last-detach edge that every `ILifecycleHandler` must now tolerate.

## Follow-up

Tracked separately, and surfaced again by the design analysis behind the next PR in this stack rather than by this one:

- Follow-up: #362, internal reconnects skip the write retry-queue reconcile. The design work for the transport PR rediscovered this independently before finding the issue, which is mild reassurance that both are describing something real. It is pre-existing on master and Connectors-wide, affecting the MQTT managed-client, OPC UA SDK and WebSocket reconnects alike, so it is the next PR's first item rather than new work.
- Follow-up: #352, model queue overflow as an explicit policy with an observable drop signal. This is where the one unbounded-growth hole in the sync analysis belongs. A stalled send leaves the buffered queue unbounded, because every source and both servers pass a null maxQueueDepth, and on the WebSocket client the stalled send also holds the connection lock that the reconnect needs, so the reconnect cannot break the stall. Bolting a send timeout onto one connector would be a local patch on a general problem.
- Follow-up: #462 stays open after #481. That PR added the bounded final drain, but the issue also covers a flush already in progress being aborted part-way through its batch, because the same linked token reaches the write handler, and its ask for a count of what was dropped is only partly met by the warning #481 logs.

Deliberately left, so the scope line is visible:

- Follow-up: `ApplyAttributeUpdates` does not recurse into an attribute's own attributes, though the factory emits them. Pre-existing; master has the same single-level shape.
- Follow-up: if the deferred retry pass roots a subject whose attribute references a further new subject whose entry that pass already consumed, the target is created unpopulated. Unreachable single-threaded; closing it means iterating the two passes to a fixpoint.
- Follow-up: the immutable-collection materializer in `DefaultSubjectFactory` dispatches through reflection per structural apply. Correct, but a compiled delegate would remove an allocation from the inbound path.
- Follow-up: `transformValueBeforeApply` does not run for newly created subjects. Documented under Contract; no production caller passes one today.

This PR also carries one unrelated fix, called out so it is not mistaken for protocol work: `ConnectorMetricsTests` completed a `TaskCompletionSource` on a thread-pool thread and then blocked that same thread, stranding the continuation in a work-stealing queue that thread never drained again, so the awaiting test timed out around one run in nine. Same family as #477. It is one statement plus a comment, kept here because this PR needs that assembly green; say the word and it moves to its own PR.

## Review history, and what it says about this change

Six defects of one class were found during review, each in a different code path, each silent, permanent and uncounted, and each found only by tracing that path. Recording them because the pattern matters more than any individual fix: the protocol's create-versus-drop decision depends on a set of subject ids surviving intact through building, serialization, projection and apply, and nothing failed loudly when it did not.

- The inbound drop tripwire counted the update root as dropped on every inbound update, so the only production signal for silent drops rose during healthy operation.
- Partial updates whose only structural change was a reorder or a remove marked nothing complete, so the field serialized as absent and the receiver read it as permission to create.
- Deferred attribute updates ran after the pass that consumes subject ids, so a subject reachable only through a subject-valued attribute was counted as dropped and then created unpopulated.
- The applier had no notion of a subject created earlier in the same apply, so a shared reference created a second instance that was never populated. This fired on every initial-state apply of a graph with shared references.
- A cross-parent move under the batch scope kept the old parent's fallback context and never gained the new one, so removing the old parent severed the subject's interception chain: no services, no change events, no replication.
- The server rebuilt every outbound payload field by field and dropped `completeSubjectIds`, handing every client the licence the field exists to withhold.

Three of these were inherited from #197, two were introduced by porting it, and one was written into the implementation plan. None would have been caught by the suites as they stood, which is the substantive argument for re-implementing rather than rebasing: a rebase preserves deletions silently, and #197 had deleted both the origin-survival tests and the read-only-types tests that would have failed.

The systemic fixes matter more than the individual ones. `UpdatePayload` now derives its fields through a copy constructor guarded by a reflection test, so a field added to `SubjectUpdate` cannot silently fail to reach the wire. `SubjectRegistryDiagnostics.DuplicateSubjectIdAttaches` makes fabrication observable in production, where previously nothing was.


RicoSuter:
Closed and reopened to re-stack: this now sits on top of the transport reliability pull request rather than beneath it, so that change can merge against master independently. GitHub does not allow changing the base of a pull request that is part of a stack, so recreating it was the only route. No code changed; the branch was rebased only, and it carries no review history to lose.

PR #484, opened 2026-08-20
fix: Reconcile and drain parked writes on a connector's own reconnect

RicoSuter:
A connector that reconnects inside its own loop never reconciled the writes it parked while it was down, and an idle client never drained them at all. Both fixes live in `SubjectSourceBase`, but they reach connectors differently, and the difference is worth stating rather than leaving to be discovered. The idle drain runs in the base class's own attempt loop, so every connector gets it. The reconnect reconcile is a pair of protected methods a connector calls around its own reconnect, and only the WebSocket client calls them here. MQTT and OPC UA reconnect inside their own loops too and have the same defect; wiring them is follow-up work, not something this pull request does.

`SubjectSourceBase.RunAsync` parks captured writes, loads the peer's state, reconciles the parked writes against it, and then runs the connected phase. All three clients reconnect inside a monitor loop nested under the listen lifetime, so a transport reconnect never iterates that loop and the reconcile never runs on it. Parked writes were replayed unreconciled, and only if a later change happened to trigger a flush, because the retry queue is drained solely by the change processor's write handler and the processor's flush returns early on an empty batch. In an idle client that is never.

The consequences are quiet. A write parked during an outage could be sent over a peer whose state had moved on, and an idle client could hold a committed write indefinitely while reporting itself synchronized and moving no counter.

Related to #362, which is the tracked form of the reconnect half. It stays open: this closes the WebSocket case and leaves MQTT and OPC UA to the same treatment.

## Why

The fix is a resume gate in `SubjectSourceBase`, plus a timer that drains the retry queue independently of the processor.

The gate suspends outbound delivery from the moment a drop is detected until the reconcile has judged the parked writes against the state the reconnect delivered. It has to cover that whole span rather than only the reconnect, because the gate check in the write handler and the socket becoming writable are not atomic: the backoff between them is the slack that keeps a flush which slipped past the check from landing on a live socket. That is recorded on `BeginResume` rather than left to be rediscovered.

The gate carries an epoch because two loops hold it. A connector's own reconnect loop runs concurrently with the attempt loop for the whole first connect window, so a completing resume clears the gate only while it still owns it. Without that, the attempt loop's completion cleared a gate the reconnect was holding and left a writable socket unguarded for the rest of the reconnect.

Two further mechanisms exist because judged and unjudged entries share one queue and the flush is edge-triggered. The reconcile's own restores are ordinary local commits, so a live processor can park them straight back while the gate is still held; a second pass judges those. And a batch parked in the window between the gate read and the enqueue is invisible to that pass, so the park path judges its own batch when it finds the gate cleared underneath it. Both are pinned to a distinct interleaving with its own regression test, and `CompleteResumeAsync` records why they are two things rather than one.

## Contract

- A write parked while a connector was disconnected is judged against the peer state that the reconnect delivers before anything sends it. It is sent, restored over the loaded value, or dropped as superseded, on a connector's own reconnect as well as on an attempt-level restart.
- A parked write leaves the queue without another write to carry it. The drain runs on the source's retry interval.
- Delivery stays suspended for the whole of a connection replacement. A write captured in that window is parked rather than sent, and is judged with the rest.
- A write that cannot be delivered because a stop lands while a connection is being replaced is counted and logged rather than parked into a queue that is about to be disposed. It is still lost; it is no longer lost silently.
- Suspension is not a delivery guarantee. The retry queue remains a bounded ring that evicts oldest first and counts what it discards, and a transactional commit reaches the source without passing through it.

## Breaking changes

None. The three members the gate adds are `protected` on `SubjectSourceBase`, so the public API snapshot changed, but nothing existing changed shape and no consumer has to act.

## Diff composition

|---|---:|---:|---:|---:|
| Production code | 3 | 292 | 15 | +277 |
| Tests and benchmarks | 7 | 923 | 5 | +918 |
| Documentation | 2 | 561 | 0 | +561 |
| **Total** | **12** | **1,776** | **20** | **+1,756** |

## Verification

- [x] Documentation updated
- [x] Unit tests
- [ ] Integration tests
- [ ] Benchmarks
- [ ] Load tests
- [ ] Chaos tests

Documentation: `docs/design/websocket-sync.md` is added here. It is the analysis this change implements, written in place of a model checker that could not be run, and it is explicit about being an analysis input rather than a proof.

Unit tests: Connectors 796, WebSocket 169, MQTT 104, solution builds with 0 warnings. Every behaviour above is covered by a test that was negative-verified, meaning the fix was removed and the test was confirmed to fail with the expected failure, not merely to go red.

Two are worth naming. The stranded-restore test blocks inside the reconcile's own restore commit so the re-park is observed with the gate provably still held; it fails 20 times out of 20 with the second pass removed. An earlier timing-based version of it failed only 17 times in 20 and was replaced for that reason. The idle-drain test makes no write at all while waiting, because a probe write triggers a flush that carries the parked write along and would make the test pass without the fix.

~~Integration tests~~, ~~Benchmarks~~, ~~Load tests~~ and ~~Chaos tests~~: not run here. This is teardown and reconnect behaviour in a shared base class with no protocol surface, and the chaos profiles exercise convergence, which by construction cannot see these losses: after each one both sides agree on a value the caller never wrote. The instrument that can see them is a write-durability oracle, which lands with the follow-up that needs it.

## Follow-up

- Follow-up: #352. The change processor's queue is unbounded for every source. A send that stalls without closing wedges the flush, and nothing drops, so the queue grows with the write rate. Recorded against invariant I6 in the design document, together with a second counterexample found here: only the transport connect is bounded during a reconnect, so a server that accepts a connection and never speaks holds the resume gate and the connection lock indefinitely.
- Follow-up: #462. The teardown drain's remaining half, where an in-flight flush is aborted part way through its batch.
- The acknowledgement gap stays open: a successful send is still treated as proof of application, so a write that reached the socket but not the peer's model is not re-asserted. That is the follow-up this branch is the base for, along with the write-durability oracle and the attributable inbound drop counter.

### DIFF
--- a/src/Namotion.Interceptor.WebSocket.Tests/Client/WebSocketClientLivenessTests.cs
+++ b/src/Namotion.Interceptor.WebSocket.Tests/Client/WebSocketClientLivenessTests.cs
@@ -641,9 +641,9 @@ public class WebSocketClientLivenessTests
         // the gate this attempt opened.
         using var portLease = await WebSocketTestPortPool.AcquireAsync();
         await using var server = await StartServerAsync(portLease.Port);
-        var throwOnce = new ThrowOnceInterceptor();
+        var loadFailed = new TaskCompletionSource(TaskCreationOptions.RunContinuationsAsynchronously);
         await using var source = CreateClientSource(
-            portLease.Port, reconnectDelay: TimeSpan.FromMilliseconds(50), writeInterceptor: throwOnce);
+            portLease.Port, reconnectDelay: TimeSpan.FromMilliseconds(50));
         await source.StartAsync(CancellationToken.None);
 
         var clientRoot = (TestRoot)source.RootSubject;
@@ -654,9 +654,18 @@ public class WebSocketClientLivenessTests
         {
             // Act - a real disconnect and reconnect, with the load failing on the reconnect's own
             // Welcome apply.
-            throwOnce.Arm();
+            // Thrown from the seam that fires once the reconnect holds a live connection and before it
+            // loads, which is inside the window the gate covers: the gate is up, the connect succeeded,
+            // and the failure lands in the same catch that has to clear it.
+            source.BeforeReconnectInitialStateLoad = () =>
+            {
+                source.BeforeReconnectInitialStateLoad = null;
+                loadFailed.TrySetResult();
+                throw new InvalidOperationException("Injected load failure.");
+            };
+
             await ((IFaultInjectable)source).InjectFaultAsync(FaultType.Disconnect, CancellationToken.None);
-            await throwOnce.Threw.WaitAsync(TimeSpan.FromSeconds(10));
+            await loadFailed.Task.WaitAsync(TimeSpan.FromSeconds(10));
 
             clientRoot.Name = "AfterFailedLoad";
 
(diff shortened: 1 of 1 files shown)


RicoSuter:
Correction, for anyone reading this later: this was **not** merged into master. GitHub closed it automatically during a re-stack, because its base branch was rebased to contain this branch's commits, which makes the head reachable from the base and reads as merged. Master was never changed by this pull request. The work is reopened against master as a new pull request.

PR #485, opened 2026-08-20
fix: Account for every change a connector cannot deliver at teardown

### LINKED ISSUE #462: ChangeQueueProcessor drops its buffered changes on shutdown without counting them

### LINKED ISSUE #483: Teardown flush: abandoned continuation runs on a disposed token source, and a flush throw can lose its dequeued batch
Found while checking how #481 interacts with the OPC UA write-integrity stack (#445, #446, #448). Both items are in master's own code, not the stack's, and neither blocks that stack.

## 1. The abandoned continuation outlives its cancellation token source

`ChangeQueueProcessor.FlushRemainingChangesAsync` bounds the teardown drain with `WaitAsync`, then leaves the drain running when the bound expires. The `CancellationTokenSource` it created is disposed by its `using` at that point, while the continuation is still using the token.

Two consequences:

- If the `Task.Run` delegate has not started when disposal happens, reading `Token` throws `ObjectDisposedException` into an unobserved task, and the batch dies with only the already-emitted timeout warning.
- The more serious one: no shipped source implements `ISupportsConcurrentWrites`, so every source waits on `writeLock.Semaphore.WaitAsync(cancellationToken)` in `SubjectSourceExtensions`, whose only catch is `OperationCanceledException`. If that wait goes async because a transaction write holds the lock, and `Register` on the disposed source throws, the exception escapes `WriteChangesInBatchesAsync` and crosses `WriteRetryQueue.FlushAsync` between its dequeue and its write.

`FlushAsync` has no exception protection between removing changes from `_pendingWrites` and handing them to the source, so up to 1024 already-dequeued writes are lost with no requeue, no drop count, and only a generic "Failed to write changes." error.

Suggested fixes, independent of each other:

- Hand the abandoned continuation `CancellationToken.None`, or keep the source alive until the continuation completes.
- Requeue the dequeued batch in `FlushAsync` when the write throws something other than a cancellation. The dequeue-then-write shape predates #481; what #481 adds is the first plausible non-cancellation throw at that call site.

## 2. A final periodic tick in flight when the stop cancels gets no teardown attempt

If the last periodic flush is mid-write when the stop cancels the linked token, its handler fails under the cancelled token and parks the batch in the retry queue. That leaves the processor's `_changes` empty, so `FlushRemainingChangesAsync` returns at its `IsEmpty` check and never makes the final attempt.

The rescue #481 exists to provide is skipped in what looks like a fairly common window.

## Two metric artifacts to expect at shutdown

Not defects, but they will read as delivery regressions in chaos kill cycles if nobody knows about them:

- When a teardown flush is abandoned at the bound and held writes were released to the pending queue head afterwards, the abandoned continuation's requeue can trip the ring trim, which evicts from the front and therefore takes exactly those just-released held writes, counting them through `AddDropped`. Shutdown-caused eviction recorded as counted loss.
- A parked-but-never-flushed final tick batch, per item 2 above.

RicoSuter:
Connector teardown now gives every accepted outbound change one explicit local outcome: confirmed, retained for retry, or counted as unconfirmed. Processor delivery and source retry use a single ownership handoff, shutdown remains bounded even when user code or cancellation callbacks block, and restarted connectors cannot inherit late diagnostics from an earlier run.

Closes #462
Closes #483
Builds on #481

## Why

The previous teardown had overlapping ownership between the processor, merger, source retry queue, and diagnostics. A batch removed from one buffer could be lost, retried twice, or counted twice depending on whether cancellation, disposal, a blocked handler, or a connector restart won the race. Repeated disposal and cancellation callback ordering could also allow delivery after disposal returned or prevent the shutdown deadline from starting.

This change replaces those paths with one admission owner and a synchronous processor-to-source handoff. Closing admission claims queued and active work exactly once before external diagnostics run, and a fixed internal deadline bounds the entire connector teardown path.

## Contract

- Every admitted outbound change is owned by exactly one of the active delivery, source retry, or terminal accounting paths.
- Processor survivors are handed synchronously to the source retry queue before processor retirement completes. A retired queue counts pending and future writes instead of retaining work that cannot be retried.
- `Dispose()` closes admission before any disposing caller returns. Work that finishes filtering after admission closes is counted and cannot invoke the handler.
- Work still blocked inside a consumer filter is not admitted. If the filter later accepts after admission closes, the accepted batch is counted before delivery is skipped.
- Immediate writes, buffered drains, retry handoff, diagnostics, and blocking cancellation callbacks share one fixed internal five-second shutdown bound. `HostOptions.ShutdownTimeout` still caps hosted shutdown as a whole.
- Work whose local completion is still unknown at the deadline is counted as unconfirmed. It may already have reached the remote endpoint or may finish later, but it cannot regain local ownership.
- Late reports from an abandoned processor are rejected by the old diagnostics epoch and cannot increment a restarted connector's counters.
- `QueueMetrics.CreateDropReporter()` is the only newly public connector API. It lets the MQTT, OPC UA, and WebSocket packages bind reports to the current diagnostics epoch without production `InternalsVisibleTo` access.
- A failed flush does not end delivery. The periodic flush catches per tick, so a throwing write handler, drop handler, logger or property filter costs that batch rather than every later one, and the failure is reported once per run rather than at the buffer time's rate.
- The teardown deadline arms when finalization begins, from cancellation or from a fault inside the pump, so a blocking handler cannot make `ProcessAsync` wait without bound.
- Delivery during teardown is at least once because a batch whose remote outcome is unknown may be retried.

## Breaking changes

- **API:** The unreleased `TeardownFlushTimeout` field, the `ChangeQueueProcessor` and `SubjectSourceBase` constructor arguments, and the six MQTT, OPC UA, and WebSocket client/server configuration properties were removed as part of the combined redesign contract. The public removal is already present on `master` through #502. Consumers of unreleased builds only need to remove those assignments or arguments.
- **Behavior:** `DropCount` and `Diagnostics.OutboundChanges.TotalDropped` include failed or terminally unconfirmed immediate, buffered, retry, and teardown writes, not only queue overflow.
- **Behavior:** Teardown can repeat a partially delivered batch because delivery is at least once.
- **Behavior:** Negative retry queue capacity is rejected instead of silently disabling buffering.
- **Contract:** Connector teardown uses the fixed internal five-second bound. Consumers configure only the host-wide shutdown budget through `HostOptions.ShutdownTimeout`.

## Performance

The final retry-queue fast-path fix was benchmarked against the immediately previous PR head, which isolates this review follow-up. `SubjectSourceBenchmark.WriteToSource` changed from 1.184 ms and 328.01 KB to 1.180 ms and 327.42 KB per operation. The timing confidence intervals overlap, while allocation fell by about 0.59 KB. The unrelated `ServiceOrderResolverBenchmark.LinearChain` noise reference moved from 1.003 us to 1.045 us with high variance and unchanged allocation, so no throughput movement is attributed to the follow-up.

The earlier scoped OPC UA load comparison processed 36 million mutations in each 15-minute arm. Server throughput changed by +0.03%, client throughput by +0.08%, allocation rate by approximately -0.45 MB/s, and post-GC heap changed from 813.4 MB to 809.0 MB. Tail latency remained in the same operating range.

## Diff composition

Current PR diff against latest `master`:

|---|---:|---:|---:|---:|
| Production code | 10 | 678 | 464 | +214 |
| Tests and benchmarks | 8 | 2'696 | 84 | +2'612 |
| Documentation | 5 | 15 | 14 | +1 |
| Repository guidance | 1 | 3 | 0 | +3 |
| **Total** | **24** | **3'392** | **562** | **+2'830** |

Production C# in the current PR is 678 additions and 464 removals, net +214. The final flush-recovery follow-up is 25 additions and 5 removals, net +20, against `94e5b57b`. The cleanup keeps the same behavior while compacting its comments and outer catch. The earlier net-negative target is no longer treated as a reason to remove useful XML documentation or narrow the explanation of the ownership contract.

The retained `AGENTS.md` changes document the agreed repository policies for long-running verification, test synchronization, documentation style, and pull request hygiene.

## Verification

- [x] Documentation updated: connector teardown and fixed-bound behavior are documented; the retry queue XML summaries are restored; public API snapshots pass.
- [x] Unit tests: the earlier full non-integration solution passed 3'353/3'353. On the cleanup head, the focused flush regressions passed 3/3, the processor class passed 49/49, and the full Connectors suite passed 785/785.
- [x] Integration tests: MQTT 103/103, OPC UA 352/352, and WebSocket 159/159.
- [x] Benchmarks: the targeted source-write comparison against the previous PR head found no throughput regression and reduced allocation by about 0.59 KB per operation; the unrelated reference showed larger timing drift with unchanged allocation.
- [x] Load tests: OPC UA baseline and final profiles ran for 15 minutes each with 36 million mutations per arm. A later current-head arm also passed with 36.0 million mutations, convergence in 2.1 seconds across 20'001 subjects and 140'007 properties, and both directions sustaining the offered 20k changes/sec. MQTT and WebSocket long profiles were excluded by the agreed reduced scope after their full project suites passed.
- [x] Chaos tests: OPC UA ran to the one-hour cap with 44/44 completed cycles passing, zero failed convergence cycles, no findings, and stable post-GC memory. A later 40-minute current-head run passed 29/29 cycles across all five rotating profiles with 151 injected events and flat post-GC memory. MQTT and WebSocket chaos profiles were excluded by the same agreed scope.

The deterministic ownership and deadline filter passed 39/39 on each of five consecutive runs before the final review follow-up. Regression tests now also cover restart after disposal during a live flush, handler-owned failure recovery, retry flush fencing behind a current write, empty retry-queue allocation, ordinary failure logging, retry-capacity validation, and the fixed teardown-bound contract.

One local full-solution attempt and the initial CI general-test job each hit an isolated five-second test-host scheduling timeout in existing source lifecycle tests. The exact reruns passed 1/1, the complete local rerun passed 3,331/3,331, and the CI retry passed. The initial MQTT CI job also hit a broken pipe while its client disconnected after broker retirement; the exact Release rerun and CI retry passed. No PR-specific scheduling workaround was added.

The long OPC UA evidence predates the final processor follow-ups and was not repeated under the agreed reduced scope. The cleanup head was verified with the focused flush regressions, the full 49-test processor class, Connectors 785/785, MQTT 103/103, OPC UA 352/352, and WebSocket 159/159.

An earlier independent review found one Important semaphore-fence regression in the first allocation fix. The pre-semaphore fast return was removed test-first, preserving both serialization and the allocation improvement.

A later deeper review retracted that round's merge-readiness conclusion and raised three more, all reproduced with failing tests before being fixed:

1. The five-second bound armed only on external cancellation, so a fault inside the pump plus a blocking handler left `ProcessAsync` waiting without bound. Finalization now signals its own start and arms the same deadline, rethrowing the captured fault at it.
2. A throwing drop callback on a failed buffered write ended the periodic flush while the pump kept filling an unbounded queue.
3. `CancellationToken.WaitHandle` throws when its source is already disposed, which left the core running with the lifecycle state stranded in `Processing`. Cancellation monitoring is now established before the core starts, and rolled back if the state transition fails.

Fixing (2) at the two callbacks where it was found proved to be the wrong shape: the drop handler was guarded and the logger two lines below it was not, and a test for the logger reproduced the identical failure. The flush now catches per tick, which closes the whole class at one site rather than one guard per call site. The retained regressions cover the distinct deadline, drop-handler, logger, and disposed-token paths.

The final external follow-up changed no production code. It strengthened the existing capacity-zero initialization test to assert no outbound reapply, exactly one accounted owned write, and zero retry depth; the exact test and the full Connectors suite passed.

The final processor follow-up preserves `_flushGate` and fixes internal-failure deadline activation, wait-registration setup ownership, throwing drop callbacks, idle merger cleanup, and immediate-buffer retention. It also applies the requested private naming and comment cleanups. A fresh independent concurrency review of the complete final diff found no Critical, Important, or Minor issues and concluded it is merge-ready.

The final cleanup changes no behavior. It compacts the flush-recovery implementation to the agreed net +20 production lines, removes a disposed-token test duplicated by `WhenCancellationWaitRegistrationFails_ThenAnotherProcessingLifetimeCanStart`, and removes trailing whitespace.


src/Namotion.Interceptor.Connectors/ChangeQueueProcessor.cs
```diff
@@ -13,11 +13,16 @@ namespace Namotion.Interceptor.Connectors;
 /// </summary>
 public class ChangeQueueProcessor : IDisposable
 {
-    /// <summary>
-    /// The teardown flush bound a processor uses when none is given. Public so a connector configuration
-    /// states this default rather than restating the value.
-    /// </summary>
-    public static readonly TimeSpan DefaultTeardownFlushTimeout = TimeSpan.FromSeconds(5);
+    // How long a stop waits for the teardown drain, and the binding constraint on every path, being
```
RicoSuter: Compact to 1-2 lines

src/Namotion.Interceptor.Connectors/ChangeQueueProcessor.cs
```diff
@@ -367,52 +353,82 @@ public async Task ProcessAsync(CancellationToken cancellationToken)
...
         // writing under it would fail every change, which is the loss this exists to prevent.
-        using var teardownTokenSource = new CancellationTokenSource(_teardownFlushTimeout);
+        var teardownTokenSource = new CancellationTokenSource(TeardownFlushBound);
+
+        // Read once, before the task starts, because the delegate below disposes the source as soon as
+        // the flush returns. Reading Token at the wait instead would race that disposal, and Token
+        // throws ObjectDisposedException once the source is gone.
+        var teardownToken = teardownTokenSource.Token;
+
+        // Off this thread, because the token only bounds a handler that observes it and the OPC UA
```
RicoSuter: Compact docs

src/Namotion.Interceptor.Connectors/ChangeQueueProcessor.cs
```diff
@@ -474,11 +479,24 @@ private async ValueTask TryFlushAsync(CancellationToken cancellationToken)
                 }
                 catch (OperationCanceledException) when (cancellationToken.IsCancellationRequested)
                 {
+                    // Cancelled means never confirmed, not failed, and nothing else recovers a batch that
+                    // left the subscription. The merger resolves by commit revision, not queue position.
+                    // Requeue the pre-merge list: the finally below returns the merger's pooled buffer,
+                    // so mergedChanges dangles the moment this unwinds.
+                    foreach (var change in _flushChanges)
```
RicoSuter: is this block correct and needed?

src/Namotion.Interceptor.Connectors/SubjectSourceBase.cs
```diff
@@ -321,6 +297,10 @@ protected sealed override async Task RunAsync(CancellationToken stoppingToken)
                         () => processor.QueueDepth, capacity: null);
 
                     await processor.ProcessAsync(stoppingToken).ConfigureAwait(false);
+
+                    // Inside the attempt's try: exiting this block disposes the listen lifetime and with
+                    // it the transport, so the same call in the finally below would find it dead.
+                    await FlushRetryQueueBeforeStoppingAsync().ConfigureAwait(false);
```
RicoSuter: is this even reachable?

src/Namotion.Interceptor.Connectors/SubjectSourceBase.cs
```diff
@@ -348,42 +328,17 @@ protected sealed override async Task RunAsync(CancellationToken stoppingToken)
...
+            // but never disposed, and a stopped source never gets another attempt at what is queued.
+            WriteRetryQueue.Retire();
+
             TransitionStateTo(SourceState.Stopped);
         }
     }
 
     [AsyncMethodBuilder(typeof(PoolingAsyncValueTaskMethodBuilder))]
     private async ValueTask WriteChangesViaRetryQueueAsync(ReadOnlyMemory<SubjectPropertyChange> changes, CancellationToken cancellationToken)
     {
```
RicoSuter: can we really just remove the whole

if (WriteRetryQueue is null)

is this now the same behavior and perf?

src/Namotion.Interceptor.Connectors/ChangeQueueProcessor.cs
RicoSuter: Needed, but it was not correct: fixed in the working tree.

Needed, because this is the whole "in-flight tick was lost" fix. `TryFlushAsync` moves the batch out of `_changes` before it writes, so a stop landing mid-write left `FlushRemainingChangesAsync` an empty queue and nothing else could recover the batch. Pinned by `WhenTheInFlightTickIsCancelled_ThenItsBatchIsStillDelivered`.

Not correct, because it requeued the pre-merge `_flushChanges` while `CountRemainingAfterDrain` counts raw queue entries. Measured, five writes to one property, same batch undelivered either way:

| teardown outcome | reported DropCount |
|---|---|
| write fails without cancelling (the catch below) | 1 |
| write cancelled at the bound (this catch) | 5 |

So the cancelled path over-reported by the collapse ratio, which at the default 8ms buffer over a hot property is much more than 5x. That contradicts the new `DropCount` doc and would make the counter useless for alerting.

Now requeues `mergedChanges.Span`. The span is still valid inside the catch, since the enclosing `finally` is what releases the pooled buffer, and `SubjectPropertyChange` is a struct so the enqueue copies. Re-merging the survivors yields the same survivors the raw batch would: `MergeWithNewer` gives the survivor the newest revision and the oldest old value, so it is already the revision extremum `IsSupersededBy` ranks against. Also cheaper, fewer enqueues.

Covered by `WhenACollapsedBatchIsCancelledAtTheBound_ThenOnlyTheSurvivorIsCounted`, proven by defect injection: restoring `_flushChanges` makes it fail with the count stuck at 5.

src/Namotion.Interceptor.Connectors/SubjectSourceBase.cs
RicoSuter: Yes, reachable, and it is the common stop path rather than an edge case.

`ProcessAsync` returns normally on a stop rather than throwing, because `PropertyChangeQueueSubscription.TryDequeue` returns `false` when the token is cancelled instead of letting the `OperationCanceledException` out (it catches its own `_signal.Wait`). So the dequeue loop exits, the `finally` runs the teardown drain, and control comes back here.

The two ways it could still throw past this line are both closed for a source: `WriteRetryQueue.FlushAsync` turns a cancelled semaphore wait into `return false`, and `WriteChangesInBatchesAsync` reports cancellation as a failed `WriteResult` rather than throwing. So `WriteChangesViaRetryQueueAsync` parks and returns normally even on the immediate path.

Pinned by `WhenTheLastWriteIsCancelledMidFlight_ThenTheParkedBatchIsHandedOverBeforeStopping`, which asserts the value is delivered and `OutboundRetries.TotalDropped` stays 0. Moving the call into the `finally` below makes that test fail with the batch rejected by a disposed transport, which is the defect injection you already noted in the PR body.

The one gap left is narrow: an unexpected throw in the connected phase skips the hand-over. The retry loop then goes round again and flushes on the next attempt, or `Retire()` counts it if the stop won the race, so nothing is lost silently either way.

src/Namotion.Interceptor.Connectors/SubjectSourceBase.cs
RicoSuter: Yes for behavior, with two deliberate exceptions already listed under Breaking changes. Not quite for perf, and the difference is real but confined.

Behavior. The old branch did three things: write directly, count the failure unless `IsExpectedShutdown` said the host was stopping, and (in `DrainOwnedWritesToRetryQueue`) empty the subscription without counting. The size-0 branch inside `WriteRetryQueue.Enqueue` already did the same counting and logging, it was just unreachable because the field was left null. So the logic was written three times and only one copy ran.

The two intended differences:
- stop-time write failures are now counted, because `IsExpectedShutdown` also read the run stopping token and suppressed them. That is the `#462` claim that turned out wrong.
- connect-window discards are now counted, where the old drain skipped the ownership filter entirely.

Both only affect `writeRetryQueueSize: 0`. A source with a queue behaves identically.

Perf. Two costs, both on the size-0 path only:
- `WriteChangesViaRetryQueueAsync` now calls `WriteRetryQueue.FlushAsync` first. That is one `Volatile.Read` and an `IsEmpty` return, since a size-0 queue is always empty. Negligible, and `Enqueue` short-circuits on `_maxQueueSize is 0` before taking the lock.
- `DrainOwnedWritesToRetryQueue` now runs the two ownership checks per change plus `CollapsePerProperty` and two array allocations, where the old code just spun `TryDequeueImmediate(out _)`. This is the one genuine regression, and it is what buys the counting above. It is paid only during the connect window of a source that explicitly turned its queue off.

For a source with a queue nothing changes: same call path, same allocations, and `Enqueue` got marginally cheaper because the metrics call moved outside the lock.

src/Namotion.Interceptor.Connectors/WriteRetryQueue.cs
```diff
@@ -25,18 +24,17 @@ internal sealed class WriteRetryQueue : IDisposable
...
+    // A negative value is the retired sentinel.
+    private int _ownedWriteCount;
+
     // Throttle flush-failure warnings to avoid log spam during extended disconnections
     private long _lastFlushWarningTimestamp;
     private bool _hasFlushWarnings;
 
-    /// <summary>
-    /// Gets a value indicating whether the write queue is empty.
-    /// </summary>
```
RicoSuter, line 34: dont remove public xml docs

src/Namotion.Interceptor.OpcUa.Tests/OpcUaRegistrationTests.cs
```diff
@@ -321,4 +321,5 @@ public void WhenTwoServersAreRegisteredForSameSubject_ThenOpcUaVariableKeysDiffe
         Assert.StartsWith("OpcUaVariable:", first.OpcUaVariableKey);
         Assert.StartsWith("OpcUaVariable:", second.OpcUaVariableKey);
     }
+
```
RicoSuter: revert file change

src/Namotion.Interceptor.Connectors/ChangeQueueProcessor.cs
```diff
@@ -205,199 +218,336 @@ private static ChangeDeliveryRule ValidateRule(ChangeDeliveryRule rule)
...
+            }
+            else
+            {
+                teardownCancellationTask = teardownTokenSource.CancelAsync();
+                Dispose();
+            }
+        }
+        finally
+        {
+            ObserveLateLifetimeAsync(
```
RicoSuter: ObserveLateLifetimeAsync does not return Task so we should not postfix with Async but name it explicitly eg *FireAndForget or i dont know what the convention is

src/Namotion.Interceptor.Connectors/ChangeQueueProcessor.cs
```diff
@@ -205,199 +218,336 @@ private static ChangeDeliveryRule ValidateRule(ChangeDeliveryRule rule)
...
+            throw new ObjectDisposedException(nameof(ChangeQueueProcessor));
+        }
 
-            while (_subscription.TryDequeue(out var change, linkedTokenSource.Token))
+        var processingTokenSource = new CancellationTokenSource();
+        var teardownTokenSource = new CancellationTokenSource();
+        var processingTask = Task.Run(
+            () => ProcessCoreAsync(processingTokenSource.Token, teardownTokenSource.Token),
+            CancellationToken.None);
+        // The wait handle is signalled before token callbacks run, so a later blocking callback cannot delay teardown.
```
RicoSuter:
```suggestion

        // The wait handle is signalled before token callbacks run, so a later blocking callback cannot delay teardown.
```

src/Namotion.Interceptor.Connectors/ChangeQueueProcessor.cs
```diff
@@ -205,199 +218,336 @@ private static ChangeDeliveryRule ValidateRule(ChangeDeliveryRule rule)
...
+        var cancellationSignal = new TaskCompletionSource(TaskCreationOptions.RunContinuationsAsynchronously);
+        var cancellationWait = ThreadPool.RegisterWaitForSingleObject(
+            cancellationToken.WaitHandle,
+            static (state, _) => ((TaskCompletionSource)state!).TrySetResult(),
+            cancellationSignal,
+            Timeout.InfiniteTimeSpan,
+            executeOnlyOnce: true);
+        var completedTask = await Task.WhenAny(processingTask, cancellationSignal.Task).ConfigureAwait(false);
```
RicoSuter:
```suggestion

        var completedTask = await Task.WhenAny(processingTask, cancellationSignal.Task).ConfigureAwait(false);
```

src/Namotion.Interceptor.Connectors/ChangeQueueProcessor.cs
```diff
@@ -449,22 +599,17 @@ private async ValueTask TryFlushAsync(CancellationToken cancellationToken)
...
-                }
-                catch (OperationCanceledException) when (cancellationToken.IsCancellationRequested)
-                {
-                    throw;
-                }
-                catch (Exception ex)
-                {
-                    _logger.LogError(ex, "Failed to write changes.");
-                }
+                await DeliverAsync(
```
RicoSuter: should we rename to WriteAsync or WriteChangesAsync?

src/Namotion.Interceptor.Connectors/ChangeQueueProcessor.cs
RicoSuter: Renamed to `ObserveLateLifetimeInBackground` in 94e5b57b so the name explicitly describes the detached continuation.

src/Namotion.Interceptor.Connectors/ChangeQueueProcessor.cs
RicoSuter: Applied in 94e5b57b. The blank line now separates cancellation-signal setup from the wait-handle ordering comment.

src/Namotion.Interceptor.Connectors/ChangeQueueProcessor.cs
RicoSuter: Applied in 94e5b57b. The blank line is present, and `Task.WhenAny` now also observes internal teardown start so the existing five-second bound covers non-cancellation failures.

src/Namotion.Interceptor.Connectors/ChangeQueueProcessor.cs
RicoSuter: Renamed to `WriteChangesAsync` in 94e5b57b.

PR #487, opened 2026-08-21
fix: Re-assert WebSocket writes that reached the socket but were never acknowledged

RicoSuter:
A WebSocket client treated a successful send as proof the server had applied the write. It is not. The connection can die between the send and the apply, nothing re-sent the write, and the next reconnect's complete-state load then overwrote the client's own value with the server's older one. The write was gone from both sides with nothing logged, and no convergence check could see it, because both sides agreed.

The client now keeps a per-connection set of writes that reached the socket but are not known to have been applied, and re-parks that set when a new connection is established, where the reconcile added in the base pull request decides whether each is sent, restored or dropped. The server counts what it applied on each connection and reports it on the existing heartbeat so entries retire, and states at connect whether it acknowledges at all.

Related to #362, which stays open: this branch does not wire MQTT or OPC UA.

## Why

Three constraints shaped this and are worth stating, because each was found the hard way.

The set is recorded at the send site under the connection lock rather than in the retry-queue wrapper. A transactional commit reaches the source directly and never passes through that wrapper, so a set kept there would have missed every transactional write, in the very profile this adds as its own gate.

Nothing was added to the update message. The server counts the updates it applied on a connection and the client counts the ones it sent, and the two line up because the transport is ordered within a connection and the client sends exactly one update per write. One additive field on the heartbeat carries the count back.

The mechanism is gated on the server actually acknowledging, not on it promising to. Heartbeats can be disabled, and their interval has no upper bound while the client's receive timeout lives in another process, so a client that retains writes without ever being acknowledged re-asserts everything it has written at every reconnect. The gate keys on an applied-through value having been observed, and on the connection having ended by its own receive timeout, because a connection that simply ended before its first heartbeat has observed nothing on a perfectly healthy server.

## Contract

- A write that reached the socket is re-asserted after a reconnect rather than assumed applied, unless the server has reported it applied.
- The server reports, per connection, how many of that connection's updates it applied, and stops advancing that count for the rest of the connection once an apply throws or drops any part of an update it could identify.
- The server states at connect whether it acknowledges. A client told it will not be acknowledged does not retain writes at all, which is the behaviour shipped today.
- An acknowledgement is not a guarantee that every property applied. A property the receiving subject does not declare is counted by a tripwire and does not stall the acknowledgement, because a schema mismatch is permanent and stalling on it would wedge the connection.
- **A reconnecting client can still re-assert a write over a newer value another client committed.** The reconcile cannot distinguish a write the server never applied from one another writer superseded, and the applied-through report narrows that window to the interval since the last heartbeat rather than closing it. This is the headline behavioural trade of this change and it is documented in the connector guide.

## Breaking changes

**API, binary only.** `SubjectUpdateExtensions.ApplySubjectUpdate` now returns `bool` rather than `void`, and takes an optional logger. Both are needed so a server can tell whether an apply silently dropped part of an update and withhold its acknowledgement, and so a drop can name the connection it came from.

Source compatible: every existing call is a statement that ignores the result and still compiles. Not binary compatible: the return type is part of the signature, so an assembly compiled against the previous version fails at runtime with a missing-method error rather than at build. **Migration is to recompile.** This matters more than usual here because this repository loads pre-compiled plugin assemblies, where the failure surfaces at runtime.

Accepted deliberately rather than worked around with a parallel method, because the base pull request in this stack already removes public types from the same assembly, so every consumer of this release must recompile regardless.

**Behavior.** A client connected to a server that does not acknowledge no longer retains unacknowledged writes. That is the same behaviour as before this stack, reached explicitly rather than by absence.

## Diff composition

|---|---:|---:|---:|---:|
| Namotion.Interceptor.Connectors | 6 | 192 | 19 | +173 |
| Namotion.Interceptor.ConnectorTester | 12 | 460 | 33 | +427 |
| Namotion.Interceptor.WebSocket | 6 | 518 | 23 | +495 |
| Tests and benchmarks | 22 | 1,705 | 76 | +1,629 |
| Documentation | 4 | 126 | 66 | +60 |
| **Total** | **50** | **3,001** | **217** | **+2,784** |

The Connector Tester figure is test infrastructure that happens to live in a non-test project. Shipped library code is +668.

## Verification

- [x] Documentation updated
- [x] Unit tests
- [ ] Integration tests
- [ ] Benchmarks
- [ ] Load tests
- [ ] Chaos tests

Unit tests: WebSocket 186, Connectors 801, Connector Tester 151, solution builds with 0 warnings. Every behaviour above is covered by a test that was negative-verified, meaning the mechanism was removed and the test confirmed to fail with the expected failure rather than merely to go red.

Documentation: the design document's case verdicts and roster are reconciled with what the code now does, the connector guide gains a write-durability section including the residual above, and the Connector Tester guide covers the oracle and the two new profiles.

~~Integration tests~~, ~~Benchmarks~~, ~~Load tests~~ and ~~Chaos tests~~: not run here. The chaos profiles cannot see the losses this fixes, because after each one both sides converge on a value nobody wrote. That is why this branch also adds a write-durability oracle to the Connector Tester, which checks that each participant's own model still holds the last value it wrote. Two profiles enable it, and both were run locally for a small number of cycles; the full runs are pending on separate hardware.

The oracle is sound only where each property has a single writer, so the profiles that enable it have no overlapping writers and it cannot see the multi-client cases at all. That limitation is documented rather than implied.

## Follow-up

- Follow-up: #362. Only the WebSocket client wires the reconnect reconcile. MQTT and OPC UA reconnect inside their own loops and have the same defect.
- Follow-up: #352 and #462, both carried from the base pull request.
- The durability oracle found a real lost write in an OPC UA chaos run on its first use, after twenty-one clean cycles, correlated with a session reconnect overwriting a client's in-flight write. That is the same defect class this branch fixes for WebSocket, in a connector this branch does not touch. It deserves its own issue.
- One drop site's attribution path was not confirmed to share the logging the other three use, and is marked unconfirmed in the design document rather than assumed.


RicoSuter:
Superseded by a re-stack: this work now sits on top of the stable-ID protocol pull request rather than beneath it, so the transport reliability change can merge independently. Reopened with the corrected base; no code changed, the branch was rebased only.

PR #488, opened 2026-08-21
fix: Retry parked writes while idle and after a connector reconnects

RicoSuter:
A source's write retry queue was retried only ahead of the next outbound change. An idle source therefore kept a failed write indefinitely while reporting `Synchronized`, and after a reconnect handled inside a connector (the OPC UA health loop, the MQTT and WebSocket monitors) parked writes waited for an unrelated write.

- **Idle flush:** `SubjectSourceBase` now retries parked writes every `retryTime` while the source is `Synchronized`, and each time it becomes `Synchronized`, so they go out right after the load that ends an outage. Every connector gets it, with no new API.
- **Read-back rule:** `ISubjectSource.WriteChangesAsync` now states that after the peer accepts a write, the source must bring the peer's value back to the model through its inbound path. That is what makes replaying parked writes as they are after a connector-internal reconnect converge. The WebSocket, OPC UA and MQTT clients already comply.

Related to #362: connector-internal reconnects still skip the reconcile, but their parked writes now go out and converge.

## Why

A connector-internal reconnect reloads the peer's state, which moves the model off a parked write, and nothing sent that write until some other property changed. With the idle flush the write is replayed right after the reload, the peer applies it, and the read-back brings the model to it within about a round trip. Measured in the WebSocket integration test: the write is back 13 to 22 ms after `Synchronized`, where before it waited for an unrelated write.

## Contract

- A parked write is retried every `retryTime` while the source is `Synchronized` and each time it becomes `Synchronized`, without a further write. Outside 1 ms to about 24 days the interval is off and only the retry on becoming `Synchronized` runs.
- A source must read accepted writes back, as stated on `ISubjectSource.WriteChangesAsync`. A write reported as failed in the `WriteResult` needs no read-back.
- After a connector-internal reconnect the model shows the peer's older value until the replayed write is read back. An OPC UA node without read-back (not monitorable with polling fallback disabled, or a deadband or data change trigger that suppresses the value) keeps it until the next load.

## Breaking changes

API: none.

Behavior:
- A write the peer keeps rejecting, for example an OPC UA `BadNotWritable`, is retried with a warning every `retryTime` instead of only ahead of the next change.

Contract:
- A custom `ISubjectSource` must read accepted writes back. A source that receives the peer's changes by subscription, broadcast or polling already does.

## Diff composition

|---|---:|---:|---:|---:|
| Production code | 2 | 84 | 1 | +83 |
| Tests and benchmarks | 3 | 206 | 38 | +168 |
| Documentation | 1 | 10 | 8 | +2 |
| **Total** | **6** | **300** | **47** | **+253** |

Bottom of the stack: #489 builds on this and #490 on that. It merges on its own.

## Verification

- [x] Documentation updated
- [x] Unit tests
- [x] Integration tests
- [ ] ~~Benchmarks~~: no change on the per-write path; the idle flush runs every `retryTime`.
- [ ] Load tests
- [ ] Chaos tests

Unit tests: all 34 assemblies green, 4,050 tests, including Connectors 817 and WebSocket 105. WebSocket including integration: 161 of 161. The solution builds with 0 warnings. Each new test fails with the mechanism it pins removed: the interval, the wake on `Synchronized`, and the 1 ms floor.

Outstanding before merge: Connector Tester chaos runs for WebSocket, OPC UA and MQTT, because the idle flush changes when all three send parked writes after their own reconnects.

## Limitations and follow-up

- Follow-up: #362. Connector-internal reconnects replay parked writes instead of reconciling them.
- Follow-up: #352. Only the transport connect is bounded during a reconnect.
- Follow-up: #462. The teardown drain aborted part way through its batch.
- A WebSocket reconnect whose connect succeeds but whose load fails never retries the load, so the source stays `Synchronizing` and the idle flush does not run until the connection drops again. This is already the case on master and is left for its own change.
- A stop can take twice the internal teardown bound when a write handler ignores cancellation, because the idle flush has its own bounded wait.
- Follow-up: the base class's own reconnect reconcile could replay parked writes the same way and rely on the read-back rule, which would remove the local restore step and about 150 lines.
- Follow-up: a write that reached the socket but was never applied is still lost on WebSocket, whose writes are not acknowledged. #490 addresses that.


RicoSuter:
### The idle flush can reorder a write, so the remark on `FlushRetryQueuePeriodicallyAsync` is not accurate

Found while designing the same idle-drain fix on a separate branch, so this is a note on the mechanism rather than a request to change the approach. The timer is the right shape and the cadence choice matches.

`SubjectSourceBase.cs:489-490` states:

> `WriteRetryQueue.FlushAsync` is serialized by its own gate, so this racing the write handler costs a wait rather than a reorder.

The gate does not cover the window that matters, because `FlushAsync` returns before taking it. `WriteRetryQueue.cs:104` early-returns on `IsEmpty`, and the queue reads empty for the whole time the timer's flush holds the semaphore with the batch already moved into the scratch buffer: `WriteRetryQueue.cs:155-156` removes the entries and republishes `_count` **before** the write at `:166`.

So:

1. The timer takes `_flushSemaphore` and dequeues parked change W. `_count` drops to 0.
2. The write handler calls `FlushAsync` at `SubjectSourceBase.cs:453`, sees `IsEmpty`, and returns `true` without ever touching the semaphore.
3. `succeeded` is true, so it falls through to the direct write at `:463` with newer change X for the same property.
4. Both calls serialize on the per-source lock in `SubjectSourceExtensions.WriteChangesInBatchesAsync`, so they do not overlap. They simply arrive in whichever order the lock grants. If the handler wins, the source receives X and then W.

The source settles on the older value, and nothing is left to correct it: W was the last thing in the queue, and the property's marker has already moved past X, so no later change re-delivers it. The per-source lock gives mutual exclusion but no ordering, which is why it does not help here.

The window is small but it is not rare in the situation this PR targets: an idle source with one parked write is exactly the state where the timer fires and a single new write arrives.

`_resumeInProgress` at `:503` does not cover it either, since this is the ordinary connected path with no resume in flight.

**Suggested fix.** Give `WriteRetryQueue` one operation that flushes the queue and then writes the supplied changes under a single `_flushSemaphore` acquisition, and route both callers through it, with the timer passing an empty batch. The handler's flush-then-check-then-write protocol at `:451-463` collapses into one call, the timer gets no write path of its own, and ordering follows from the acquisition rather than from a caller-side rule. Keeping the existing behaviour that a failed flush queues the new changes behind the parked ones is what stops a newer value overtaking an older one.

Two details worth carrying over if you take that shape:

- The empty-queue case can no longer skip the semaphore, so it should stay otherwise free. `WriteRetryQueue.cs:130-134` currently grows the scratch buffer by doubling on every entry until it reaches 1024, which allocates on roughly the first five calls of a source's life whether or not anything is dequeued. Moving that growth into the dequeue path keeps an empty pass allocation-free.
- A source implementing `ISupportsConcurrentWrites` would then be serialized by `_flushSemaphore` despite opting out of the per-source lock. No production source implements it, but the interface's documentation should say so.

**Test that pins it**, since a test asserting the two writes do not overlap passes either way: gate the fake source so the timer's flush has dequeued W and is blocked in the write, release a handler write of X for the same property, then assert the source's last received value is X rather than W.

PR #489, opened 2026-08-21
feat: Stable-ID subject update protocol and ID-resolving pipeline

RicoSuter:
Replaces the index-based subject update protocol with a stable-ID one. Subjects are addressed by their registry ID, a partial update carries only the subjects it changes and no path from the root, and a collection or dictionary update states its whole membership by ID instead of positional insert, remove and move operations. A receiver resolves every reference by ID, so shared subjects, cycles, moves between parents and reorders round-trip with their identity instead of arriving as copies.

- **Wire:** the WebSocket protocol moves to version 2, a hard break. The same shape reaches AspNetCore HTTP payloads. The TypeScript client needs a version 2 applier before this ships.
- **Producer:** a subject new to the receiver arrives complete whatever the order of changes in the batch, a change is sent only when its subject is reachable through included properties, and batched writes never split a move or a nested new subject across two messages.
- **Receiver:** no subject is created without a payload, a declared-type mismatch or a missing payload fails only that property, and a property without a setter applies the payload to the instance it holds.
- **Tracking:** `LifecycleInterceptor.CreateBatchScope` keeps a subject moved between properties within one apply attached and registered, and context inheritance follows the parent a subject still has.
- **Tooling:** a `websocket-structural` Connector Tester profile produces cross-parent moves, re-adds, reorders and shared references.

Related to #197, whose protocol and pipeline work this stack carries onto current master.
Related to #537, which fixes the same integrity problems within the index-based protocol.

## Why

The index-based protocol mints IDs per update, so a receiver can match subjects only by position. A subject shared by two properties arrives as two copies, a move between parents arrives as a removal plus a new copy, an insert and a move in one collection corrupt each other, a change below a loop of first parents cannot be addressed, and exclusion can only follow first-parent paths. Those are properties of the addressing, not bugs a producer can fix, which is why a stable-ID protocol is the direction rather than more wire markers.

A randomized round-trip harness makes the difference measurable. It mutates a source graph with twelve scenario mixes (trees, shared subjects, back references, reassignments, reorders, exclusion, shuffled change order), builds partial updates from each batch, applies them to a mirror over JSON and compares the two graphs after every round.

| Outcome | This PR | #537 |
|---|---:|---:|
| Rounds | 71,497 | 62,272 |
| Value divergence | 0 | 1,996 |
| Identity split (shared subject arrives as copies) | 0 | 5,676 |
| Default-valued subject created | 0 | 54 |
| Apply failure | 0 | 61 |
| Plain-tree scenarios, any problem | 0 | 0 |

Most of #537's divergences are the limitations it documents: 1,368 of them sit below a loop of first parents. The round counts differ because a seed ends early when the source graph itself loses a registration.

The remaining harness noise, 26 source-side and 1 receiver-side unregistered subject, traces to a lifecycle defect that is also on master: a context-attached root is detached with its whole graph when its last property reference is removed.

### Wire

- Every subject is addressed by its stable registry ID. `root` maps the sender's root ID onto the receiver's root and may be absent.
- A partial update carries only subjects with changes. It never carries a path from the root.
- A collection or dictionary update states the whole resulting membership as `items`, each with `id` and, for a dictionary, `key`. Collection order is array order.
- A partial update always carries `completeSubjectIds`, empty when nothing in it is complete. An absent set means every subject is complete, which only a complete update uses.

### Producer

- A change is sent only when its subject is reachable from the root through included properties.
- A subject new to the receiver arrives complete, including every member of its collections and dictionaries, whatever the order of changes in the batch.
- Repeated changes to one property are merged by commit revision.
- Computed subject projections are not published.
- A subject a property holds without Registry metadata, during the attach race or without a registry, is serialized from its own metadata. Processors and transforms do not apply to it, and its attribute properties appear as plain properties. A subject that has left the graph is left out on its own, never with the rest of its property.
- Batched source writes and server broadcasts merge a repeated property by commit revision and put every subject-holding change into the first slice, which may exceed the batch size. Value changes are sliced by size, and each write is one message.

### Receiver

- An ID names one subject for the whole apply. A reference back to the root resolves to the local root.
- A subject is created only when its ID is complete and its payload is in the update. It is populated before it enters the graph, so its initial values bypass the write-interceptor chain: no validation, no derived recalculation, no change events, and no `transformValueBeforeApply`.
- A resolved subject that does not fit the property's declared type, or an ID without a payload, fails that property. The rest of the update applies.
- A property without a setter applies the payload to the instance it holds and re-keys it to the sender's ID, unless the update names its current ID elsewhere. Structure it cannot store is reported once per update in a warning.
- A settable position that receives an ID it does not know gets a new instance. After a sender restart, which assigns new IDs, the first complete update therefore replaces every subject at a settable position.
- A relay's downstream peers keep the relay's previous ID for a subject re-keyed at a position without a setter until they reconnect.
- Every write carries the update's `ChangeOrigin`, so echo suppression holds.
- `SubjectUpdateDiagnostics` counts what the pipeline drops. A property the receiver does not declare, and subjects named only through it, through structure the receiver cannot store, or through a computed projection, are not drops. A later partial update to a member the receiver could not store does count, because the receiver keeps no memory across updates.

### Tracking

- `CreateBatchScope` defers the last detaches made on the calling thread until that thread's outermost scope closes. Detaches on other threads are not deferred. The scope can be disposed on any thread, and disposing it twice is safe.
- Under a scope a deferred last detach is reported twice: the reference removal at once, and the detach with `IsContextDetach` when the scope closes. Lifecycle handlers must tolerate that.
- Context inheritance follows a parent the subject still has when the parent it inherited from leaves, and falls back to the chain's root when none is left. A subject referenced from two independent graphs inherits only from the graph it joined first and loses that inheritance when it leaves that graph, even while the other still references it. Sharing a subject across independent graphs is not supported.

## Breaking changes

API:

- `Namotion.Interceptor.Connectors`: `SubjectCollectionOperation`, `SubjectCollectionOperationType`, `SubjectPropertyUpdate.Operations` and `SubjectPropertyUpdate.Count` are removed. Send the resulting membership in `Items` instead.
- `SubjectPropertyItemUpdate.Index` (`required object`) becomes `Id` (`required string`), the item's subject ID. The previous `Id` (`string?`) becomes `Key` (`string?`), the dictionary key.
- `SubjectUpdate.Root` becomes `string?`. Added: `SubjectUpdate.CompleteSubjectIds`, a protected copy constructor on `SubjectUpdate`, and the `SubjectUpdateDiagnostics` static class.
- `Namotion.Interceptor.Tracking`: added `LifecycleInterceptor.CreateBatchScope`.
- `Namotion.Interceptor.Registry`: added the `SubjectRegistryDiagnostics` static class.
- `Namotion.Interceptor.WebSocket`: `WebSocketProtocol.Version` is 2, and `UpdatePayload` gains a constructor from a `SubjectUpdate`.

Behavior:

- The WebSocket wire protocol goes from version 1 to 2 with no dual-shape support. A version 1 peer is rejected at the handshake, so both ends must be upgraded together, including the external TypeScript client.
- AspNetCore HTTP payloads serialize the same types, so structural properties emit `id` and `key` instead of `index`, and `root` may be absent.
- A client connected to a restarted server receives new instances at every settable position on the first complete update.
- Every source that sets `WriteBatchSize` now merges a repeated property by commit revision and writes every subject-holding change in its first slice. A failed write reports the merged change.
- Partial updates drop changes to subjects not reachable through included properties, even with no processors configured.
- Context inheritance no longer keeps a fallback to a parent the subject left.

Contract:

- A source or transaction writer must build its payload from the changes it is given, never by reading subject properties. This is stated on `ISubjectSource.WriteChangesAsync`. See #476.
- An `ILifecycleHandler` must tolerate the double-reported last detach described under Tracking.

## Performance

Not measured with the benchmark suite yet; `SubjectUpdateBenchmark` is outstanding and the `performance: requires-benchmark` label gates the merge. A rough probe on a 4-change partial build, which is on the connector hot path, went from 1.90 to about 2.48 µs per build, the cost of the reachability check and the per-property merge. Allocations went down.

| Probe | Before | After | Delta | Allocated |
|---|---:|---:|---:|---|
| Partial build, 4 changes | 1.90 µs | 2.48 µs | +31% | 2,720 B to 2,632 B |
| Partial apply | | | | 64 B to 40 B |
| Complete apply | | | | 5,520 B to 5,184 B |

Payload size moves both ways. Inserting one item into a 1,000-item collection costs 32,526 bytes, against 499 bytes in the index-based protocol, because the complete membership is sent with 22-character IDs. A value change inside an item costs the same (about 186 bytes), deeper nesting favours this protocol because no path is sent, and a complete update of 1,000 items shrinks from 347 KB to 317 KB.

## Diff composition

|---|---:|---:|---:|---:|
| Namotion.Interceptor | 1 | 74 | 0 | +74 |
| Namotion.Interceptor.Connectors | 20 | 1'674 | 1'196 | +478 |
| Namotion.Interceptor.ConnectorTester | 8 | 690 | 35 | +655 |
| Namotion.Interceptor.Registry | 4 | 89 | 0 | +89 |
| Namotion.Interceptor.Tracking | 4 | 433 | 15 | +418 |
| Namotion.Interceptor.WebSocket | 3 | 33 | 10 | +23 |
| Production code | 1 | 3 | 0 | +3 |
| Tests and benchmarks | 120 | 10'057 | 2'307 | +7'750 |
| Documentation | 5 | 441 | 213 | +228 |
| Project and CI files | 2 | 2 | 0 | +2 |
| **Total** | **168** | **13'496** | **3'776** | **+9'720** |

The Connector Tester figure is test infrastructure in a non-test project.

## Verification

- [x] Documentation updated
- [x] Unit tests
- [x] Integration tests
- [ ] Benchmarks
- [ ] Load tests
- [ ] Chaos tests

Unit tests: all 34 assemblies green, 4,304 tests, including Connectors 1,022, Tracking 606 and Registry 163. WebSocket including integration: 168 of 168. The solution builds with 0 warnings.

Documentation: `docs/connectors-subject-updates.md` describes the stable-ID protocol and its contract, `docs/connectors-websocket.md` version 2, `docs/design/tracking-lifecycle.md` the batch scope and context inheritance, `docs/connectors.md` batched writes and `docs/connector-tester.md` the new profile.

Every fix is covered by a test that fails with the fix reverted. The randomized round-trip harness behind the table under Why was run locally over 200 seeds per scenario and is not part of the suite.

Benchmarks, `websocket-load` and `websocket-structural` are outstanding and run on separate hardware:

- `pwsh scripts/benchmark.ps1 -Filter "*SubjectUpdateBenchmark*" -LaunchCount 3`.
- `websocket-load` as the new-protocol baseline, and `websocket-structural` for structural churn with transactions off.

## What a reviewer should look at hardest

1. `SubjectUpdateApplyContext`: when a receiver creates, adopts or re-keys a subject, and the refusal when the update still names the held subject's current ID.
2. `ContextInheritanceHandler`: moving the inherited context to a remaining parent, and the fallback to the chain's root.
3. `LifecycleBatch`: per-thread deferral under the interceptor lock.
4. `SubjectChangeSlices` on the generic source write path, which every connector with `WriteBatchSize` now goes through.
5. `SubjectUpdateBuilder`: the reachability check and change merging on the partial-update hot path.

## Follow-up

- Follow-up: a context-attached root is detached with its whole graph when its last property reference is removed. Pre-existing on master, and the source of the remaining unregistered counts in the harness.
- Follow-up: the complete membership makes structural changes to large collections expensive on the wire. Whether large collections need an incremental form is a design decision for after the benchmark.
- Follow-up: the immutable-collection materializer in `DefaultSubjectFactory` dispatches through reflection, a Native AOT gap.
- Follow-up: `transformValueBeforeApply` does not run for subjects an update creates.
- Follow-up: a subject shared by two independent lifecycle graphs inherits only from the first. Leaving that graph drops the inheritance while the other graph still holds it, and under a batch scope a detach order exists that keeps a fallback into the graph it left. Master is wrong here in a different way; a correct design needs cross-graph coordination that this PR does not attempt.


src/Namotion.Interceptor.Connectors/Updates/Internal/SubjectUpdateFactory.cs
```diff
@@ -86,28 +106,56 @@ internal static void ProcessSubjectComplete(
             if (!IsPropertyIncluded(property, builder.Processors))
                 continue;
 
-            var propertyUpdate = CreatePropertyUpdate(property, builder);
-            properties[property.Name] = propertyUpdate;
+            if (properties.TryGetValue(property.Name, out var existingUpdate))
```
RicoSuter, line 109: [P1] Complete referenced child payloads when promoting an existing structural entry

There appears to be an ordering-dependent data-loss case when `ProcessSubjectComplete` finds a collection/dictionary entry already produced by a change earlier in the same batch. This branch preserves the entry and fills its attributes, but does not complete the subjects referenced by its membership.

Example to cover with a producer-to-receiver regression test:

1. The supplied batch processes `parent.Children: [A] -> [A, B]` before `root.Father: null -> parent`. The receiver does not yet know parent, A, or B.
2. `BuildCollectionUpdate` emits membership `[idA, idB]`, but includes complete state only for B because A occurs in the old collection.
3. Processing the parent assignment calls `ProcessSubjectComplete(parent)`. This branch keeps the existing Children entry without traversing A, so A still has no complete payload and is absent from `completeSubjectIds`.
4. `ResolveOrCreateSubject` cannot resolve or create A and skips it, leaving the receiver with `[B]` instead of `[A, B]`.

The dictionary path appears to have the same gap for retained key/value entries. Stable-ID complete membership removes the positional-baseline problem, but every child unknown to a newly created receiver still needs creatable state. When promoting the parent to complete, preserve the captured membership and explicit child changes while ensuring its referenced children have the required complete payloads. Test both batch orders for lists and dictionaries with a fresh receiver.

This finding is based on tracing the code at this revision; I have not run the proposed reproducer against this PR.

RicoSuter: Fixed in 92af250c2. Completing a subject now states every published property in full, including members that were already in the container, so completeness no longer depends on which change in the batch touched the container first. `SubjectUpdateBatchTests.WhenAContainerGainsAMemberAndItsOwnerIsAssignedInOneBatch_ThenAFreshReceiverGetsEveryMember` covers both batch orders for a list and a dictionary against a fresh receiver.

PR #490, opened 2026-08-21
fix: Reconcile and re-assert WebSocket writes across reconnects

RicoSuter:
Two WebSocket client changes for writes around a reconnect the client handles itself:

- **Resume gate**, moved here from #488: from drop detection until the reconnect's load, the client parks outbound writes, and after the load the reconcile judges each parked write against the loaded state (sent, restored locally, or dropped as superseded) instead of replaying it as it is. The client's own write is back in its model right after the load rather than after the server reads it back.
- **Write acknowledgement**: a successful send is no longer treated as applied. The client keeps the writes each connection sent that the server has not acknowledged and re-parks them for that reconcile after a reconnect. The server reports on the heartbeat how many of the connection's updates it applied, and states at connect whether it acknowledges.

Without this pull request, #488's idle flush and read-back rule already make parked writes converge after a reconnect. What this adds is the immediate local restore, and re-sending a write the connection lost in flight, which otherwise disappears on both sides with nothing logged.

Related to #362: this covers the WebSocket client only.
Related to #591, the Connector Tester oracle that detects a write lost on both sides.

## Why

The connection can die between a send and the server's apply. Nothing re-sent the write, and the reconnect's load then overwrote the client's value with the server's older one, so both sides agreed on a value nobody wrote last and no convergence check could see it. The in-flight set is recorded at the send site under the connection lock, so transactional commits, which bypass the retry queue, are covered too. The acknowledgement rides on the existing heartbeat, and the client only retains writes once it has seen the server acknowledge, because heartbeats can be disabled.

## Contract

- A write that reached the socket is re-asserted after a reconnect unless the server has reported it applied.
- From drop detection until the reconcile after the reload, outbound writes are parked rather than sent.
- The server reports, per connection, how many of that connection's updates it applied, and stops advancing that count for the rest of the connection once an apply throws or drops any part of an update it could identify.
- The server states at connect whether it acknowledges. A client told it will not be acknowledged does not retain writes.
- An acknowledgement does not guarantee every property applied. A property the receiver does not declare, and subjects reachable only through it, are counted and do not stall the acknowledgement.
- A reconnecting client can still re-assert a write over a newer value another client committed, within the interval since the last heartbeat.

## Breaking changes

API:
- `SubjectUpdateExtensions.ApplySubjectUpdate` returns `bool` rather than `void` and takes an optional logger. Source compatible, not binary compatible: recompile, which matters for pre-compiled plugin assemblies.
- `SubjectSourceBase` gains three protected members: `BeginResume`, `TryEndResume` and `CompleteResumeAsync`.

Behavior:
- A client connected to a server that does not acknowledge retains no unacknowledged writes, as before this stack.
- Dropped inbound subjects are logged as one warning per apply naming the origin and the subject IDs.

## Diff composition

|---|---:|---:|---:|---:|
| Namotion.Interceptor.Connectors | 6 | 367 | 48 | +319 |
| Namotion.Interceptor.WebSocket | 6 | 545 | 29 | +516 |
| Tests and benchmarks | 16 | 1'892 | 166 | +1'726 |
| Documentation | 4 | 610 | 8 | +602 |
| **Total** | **32** | **3'414** | **251** | **+3'163** |

Top of the stack, on #489.

## Verification

- [x] Documentation updated
- [x] Unit tests
- [x] Integration tests
- [ ] Benchmarks
- [ ] Load tests
- [ ] Chaos tests

Unit tests: all 34 assemblies green, 4,325 tests, including Connectors 1,039 and WebSocket 115. WebSocket including integration: 188 of 188. The solution builds with 0 warnings. The gate tests fail with the dispatcher sending unconditionally, and a test fails when the idle flush ignores the gate.

Documentation: `docs/design/websocket-sync.md` lands the design this stack was built from, and the connector guide gains a write-durability section including the re-assert trade above. Chaos and durability runs are outstanding and use the oracle from #591.

## Limitations and follow-up

- Follow-up: #362. MQTT and OPC UA reconnect inside their own loops and replay parked writes as they are.
- Follow-up: #352 and #462, carried from #488.
- The retry-queue flush that runs when the change processor tears down does not check the resume gate, so a source stopping while a reconnect holds the gate can send parked writes that were never judged.
- The oracle found a lost write in an OPC UA chaos run, correlated with a session reconnect overwriting a client's in-flight write. It deserves its own issue.
- A transactional durability profile for the Connector Tester, with failed-commit handling, is follow-up work.


RicoSuter:
Follow-up work from this stack is tracked in #492, which carries the structural ownership and transactional routing design, the decisions that work has to take rather than inherit, and the acceptance matrix. It also records the things learned while building this stack that change its scope, in particular that the write-durability oracle added here cannot see the multi-client cases, and that removing the structural ownership filter widens the cross-client re-assertion residual documented in the description above rather than leaving it unchanged.

The lost write the oracle found in an OPC UA chaos run is #493.

PR #494, opened 2026-08-21
refactor: Simplify lifecycle ownership to one context

RicoSuter:
A subject now belongs to one exact context, replacing fallback-context resolution with ownership determined by reachability from anchors over occurrence-aware edges. The rewrite unifies lifecycle descent, parent tracking and context inheritance, releases unreachable cycles, and coordinates structural writes with attachment transitions.

This is a breaking API and behavior change. Structural assignment inside lifecycle callbacks remains unsupported in this PR. Several consistency defects and incomplete rejection paths remain open; the limitations below are part of the merge assessment, not guarantees that invalid operations always leave state unchanged.

Related to #501. Follow-up: #524. Callback-support and additional correctness changes are under separate review in #527, which targets this branch.

## Why

The fallback model let a subject resolve services through several contexts without a single ownership answer. The new model gives each subject one attachment and makes graph reachability decide whether that attachment survives. It also coordinates field writes, ownership claims and graph reconciliation so competing contexts cannot normally claim the same subject.

## Contract

- With lifecycle enabled, ownership follows committed structural edges from explicit or provisional anchors. Without lifecycle, attachment is root-only. Sharing one subject between contexts is forbidden; remove its supporting edges and anchor before attaching it elsewhere.
- A generated structural setter commits its backing field before lifecycle reconciliation. Ordinary replacement already delivers old-child detach callbacks before new-child attach callbacks. A detach callback can therefore read the newly assigned field value. This ordering does not imply atomic visibility of the entire graph during callbacks.
- Structural validation examines the proposed component and then the authoritative stored value. Normalizing/subset setters are supported. A terminal introducing an unproposed foreign subject can be rejected after it stores the field; the framework does not restore arbitrary terminal side effects. Pre-store validation rejection and exceptions after publication have different guarantees.
- At most one context's topology transaction may be active on a thread. Nested topology work in a second context throws. Same-context structural reentry from a downstream interceptor, terminal or getter at callback depth zero is permitted, with the discovery/reconciliation gaps listed below.
- During lifecycle callback scope, structural assignments and explicit attach/detach are intended to throw; scalar writes and same-context dynamic property admission are supported. The guard is not yet a universal first-use check: an unattached generated setter or an equality short-circuit can bypass it. The cross-thread wait policy is a timeout heuristic, not deterministic contract enforcement.
[...]

PR #501, opened 2026-08-31
refactor: finalize race-safe single-context lifecycle protocol

RicoSuter:
This draft is an alternative implementation of #494. It keeps the same single-context ownership direction and intentional removals, but replaces the lifecycle and structural-write protocol to close the remaining deadlocks, publication gaps, stale projections, and derived-cascade races. PR #494 remains open and unchanged so both implementations can be reviewed side by side.

The complete alternative is compared against `master` in this PR. Reviewers interested only in the redesign can use the [comparison against PR #494](https://github.com/RicoSuter/Namotion.Interceptor/compare/rewrite/single-context-lifecycle...rewrite/single-context-lifecycle-v2).

Related to #494

## Why

The reviewed PR established the desired product contract but its implementation still admitted several concurrency failures: the whole-chain structural gate could deadlock with a downstream interceptor waiting for another writer, the value proposed before the chain could differ from the value actually committed, storage and topology publication could separate, remove-before-add could expose a claim gap, attachment capture and callback delivery could publish stale state, and a temporarily fenced derived target could fail an already committed source write.

This candidate organizes correctness around five blocks: one atomic attachment authority per subject, coordinated structural storage, one occurrence-aware ownership graph, leases and reservations as temporary reachability roots, and immutable revisioned callback journals drained outside framework locks. The [design specification](docs/superpowers/specs/2026-08-30-lifecycle-write-protocol-redesign-design.md) and [implementation plan](docs/superpowers/plans/2026-08-30-lifecycle-write-protocol-finalization.md) contain the detailed invariants, race analyses, and verification evidence.

## Contract

- A subject is attached to exactly one context or none. Ownership is reachability from explicit or provisional anchors over committed occurrence-aware edges.
- Interceptors may rewrite `PropertyWriteContext<TProperty>.NewValue` while forwarding the received context by `ref` toward one terminal invocation. Terminal entry freezes that final value. Mutation after `next` remains interceptor unwind state and cannot change committed storage or built-in publication.
- A structural terminal rereads the exact predecessor and faithfully stores the frozen value while its ownership-graph publication commits through the same short terminal-lock plus topology-gate boundary.
- Active structural leases and same-context reservations are temporary roots. Concurrent removal can commit without detaching their protected closure, and final release performs one deferred reachability sweep.
- Lifecycle journals contain immutable revisioned projections and drain after framework locks are released. Journals from different threads may overlap, while stale same-entity delivery cannot overwrite newer Registry state.
- A derived target temporarily fenced after its source commits coalesces one guarded continuation. Exact attachment and recalculation sequence checks prevent stale retry from duplicating newer publication.
- Generated structural access remains coordinated before first attachment while preserving the historical prepublication timestamp of `0`. Scalar generated access retains its normal routing shape.
- Direct in-place mutation of an assigned mutable collection remains outside the intercepted-write contract. An intercepted assignment or lifecycle capture is required to publish a new immutable structural snapshot.

## Breaking changes

**API relative to `master`.**

- `IInterceptorSubjectContext.AddFallbackContext`, `RemoveFallbackContext`, and `HasFallbackContext` are removed. Attach the subject to the one context whose services it uses.
- `IInterceptorSubject.Context` becomes `IInterceptorSubject.Executor`, and `IInterceptorExecutor` no longer implements `IInterceptorSubjectContext`. Use `TryGetContext()` or `GetContext()` for attachment and the executor for interception.
- Public subject `SyncRoot`, `SetPropertyValueWithInterception`, `ContextInheritanceHandler`, and `ParentTrackingHandler` are removed.
- `InterceptorSubjectContext` is sealed. Explicit attach/detach extensions, `SubjectAttachmentAnchorKind`, singleton context service markers, and `SubjectPropertyRegistration` support the replacement model.

**API differences from PR #494.**

- The PR-only `ILifecycleInterceptor.EnterStructuralWriteGate` and `ExitStructuralWriteGate` methods are removed.
- `InterceptorExecutor.GetGeneratedPropertyValue` and `SetGeneratedPropertyValue` expose the coordinated generated and advanced hand-written structural storage seam.
- `LifecycleConflictException` reports prompt retryable lifecycle admission conflicts.
- Lifecycle callback payloads add immutable context, revision, property, parent, and child projections. `IPropertyLifecycleHandler` gains the revisioned `RefreshCollectionProperty(SubjectPropertyLifecycleChange)` overload used by Registry reorder and rekey publication.

**Behavior and implementer contract.**

- A structural write or attachment transition racing a non-stable attachment phase or exclusive reservation can fail promptly with `LifecycleConflictException` instead of waiting around arbitrary user code.
- Attached generated, Dynamic, or hand-written structural storage must supply a synchronous, nonblocking, non-reentrant raw reader and a faithful exception-free raw writer. Normalization belongs before the terminal, such as in `On<Property>Changing` or an interceptor.
- Lifecycle callbacks remain synchronous for their originating operation but are not globally serialized across threads.
- Constructor-created generated structural state retains coordinated revisions without inventing timestamps before first attachment.

## Diff composition

|---|---:|---:|---:|---:|
| Production code | 112 | 7'416 | 2'703 | +4'713 |
| Tests and benchmarks | 197 | 20'075 | 4'537 | +15'538 |
| Documentation | 16 | 963 | 289 | +674 |
| Scripts and CI | 1 | 15 | 1 | +14 |
| Project and CI files | 3 | 9 | 0 | +9 |
| **Total** | **329** | **28'478** | **7'530** | **+20'948** |

Relative to PR #494, the five lifecycle production projects are `+4,977/-3,384`, net `+1,593`. Relative to current `master`, those projects are net `+4,695`. The candidate is conceptually simpler than the PR because it retains one graph, journal, terminal, and admission protocol, but it is not a source-line reduction. Independent deletion review found no sizeable remaining removal that did not also discard a distinct correctness invariant.

## Verification

Merged `master` (`3d397bca1`) so this branch and #494 share the same base. Two of master's #485 teardown tests needed their fixtures adapted: this branch consumes per-subject commit revisions during construction, so the tests' hardcoded revisions rank stale and every change is dropped. The fixtures were adapted rather than the production code, but whether that construction-time consumption is itself the defect is an open question for this branch.

- [ ] Public lifecycle documentation updated. The internal design and execution documents are complete; public docs are deferred while this remains a comparison candidate.
- [ ] Unit tests: 4,217 of 4,218 pass after merging `master`. `DynamicPropertyWithWriteInterceptorTests.WhenAddPropertyCalledConcurrently_ThenAllPropertiesAreRegistered` fails, reproduced 3 of 3 in isolation at the pre-merge head `f01a27c7d`, so it predates the merge rather than being caused by it. A `LifecycleConflictException` escapes `BeginNonStructuralWrite` through `PublishInitialValue` during concurrent `AddProperty`, so the retryable-conflict protocol reaches a caller that does not retry. Left unfixed: it needs a decision on this branch's design rather than a test change.
- [x] Focused suites: Core 196/196, Tracking 761/761, Registry 193/193, Generator 273/273, Dynamic 11/11, Connectors 738/738, and ConnectorTester 117/117.
- [x] Debug and Release solution builds completed with zero warnings and zero errors.
- [x] Release package creation completed after the Release build, with only expected warnings from projects that disable packaging.
- [ ] Integration tests: not run because connector implementations and HomeBlaze UI were not changed.
- [ ] Benchmarks: not run. Performance comparison remains a deliberate follow-up after the design and size decision.
- [ ] Load tests: not run; no connector implementation changed.
- [ ] Chaos tests: not run; no connector implementation changed.


RicoSuter:
## Evaluation against #494

Both branches were measured rather than compared on their descriptions: the eleven defect classes pinned by #494's acceptance suite were ported onto this branch and run, and both protocols were attacked with probes executed on each branch. The acceptance suite is a fair instrument here because it predates both implementations, so neither can have been shaped to fit it.

Recording the result so the decision is on paper.

### What this branch does better, confirmed by attack rather than taken from the description

- **Callbacks genuinely run outside all framework locks.** The cross-context deadlock class is impossible by construction here, where #494 backstops it with a per-thread topology rule and keeps user code inside the gate. This is the one architectural difference that is fundamental rather than incremental.
- **Journal ordering held.** Same-entity inversion is impossible by construction, and the Registry's per-entity revision guards held against stale-delivery attacks.
- **The retention design is not circular.** The lease and deferred-sweep model was probed with the exact reparent-cascade shape that falsified an earlier retained-set design during #494's review, and it handles it correctly: retention walks committed edges to protector roots rather than depending on the discovery skip.
- **The field-not-restored boundary is closed by construction**, because the terminal freezes the final value.
- Late `WithLifecycle` registration is rejected here too, which the description does not mention.

### What blocks adoption today

**1. Legal writes throw under ordinary concurrency.** Six race shapes were probed on both branches. All six commit on #494 and all six raise `LifecycleConflictException` here. The sharpest is plain scalar traffic: reparenting a subject while a background thread writes one of its scalar properties throws on the first attempt, because the edge transition collides with the scalar-write admission counter. The rejection window is bounded by user-callback drain time rather than being short, and it fires on bystander subjects whose attachment does not change. No connector, hosting or sample code catches the exception; the only retry loop in the tree is this branch's own test helper, and `IsTransientCapture`, the discriminator a consumer would need to retry correctly, is internal.

**2. The unfaithful-terminal defect reproduces, silently.** A raw writer that substitutes a foreign-owned subject commits with no exception: the graph attaches the proposed subject while the field and getter expose the foreign one, owned by another context. A normalizing writer that drops a subject leaves it attached with a phantom parent edge until the property's next write. The terminal contract was redefined to require a faithful raw writer, and that requirement is not enforced. #494 detects and rejects this shape.

**3. Hand-written structural subjects on the legacy `SetPropertyValue` path now throw** on every attached structural write, which is an unlisted API break.

**4. An acceptance test in the derived-convergence area is flaky**, failing roughly half of isolated runs at the rendezvous deadline regardless of how long that deadline is. Behind it sits one unresolved question: whether a projection can publish a stale value under the path that accepts a fenced value without re-running.

### Size and performance

Production is +4,712 against master, versus +3,226 for #494, and this branch's own description records that it fails its numeric size gates. It is unbenchmarked by its own admission. Reading the code, the mechanisms a benchmark would need to price are: two attachment-monitor acquisitions on every scalar write, a `SyncRoot` acquisition on every generated structural read, per-structural-write snapshot and dictionary allocation plus journal construction, and a process-wide static lock taken per projection revision. #494 has a full three-arm gate with the CPU pinned and every regression's mechanism named.

### Suggested sequencing

Neither branch subsumes the other, so treating them as competitors undersells this one.

- **#494 merges as the contract change.** It is the smaller sound implementation of the same public contract, reviewed, benchmarked and green, so consumers absorb one breaking release.
- **This branch becomes the source for a later internals-only migration**, specifically callbacks-outside-locks, which changes no public contract and is therefore invisible to consumers. It is viable once the conflict-liveness problem is resolved (a bounded gate-held wait on scalar-write draining is lock-order-safe, since scalar writes never take the gate, or a public transient discriminator so a retry policy can be written), terminal-store enforcement is restored, and it passes the same benchmark gate.

The acceptance suite is what makes that second step judgeable mechanically rather than by argument, since it pins the behaviour independently of either implementation.

RicoSuter:
The acceptance suite from #494 is now landed on a branch off this one, for the record: [`eval/501-baseline`](https://github.com/RicoSuter/Namotion.Interceptor/compare/rewrite/single-context-lifecycle-v2...eval/501-baseline). Three commits, **ten test files, zero production files**.

It pins the eleven defect classes #494 reproduced, written before either implementation was finished, so it judges both without favouring either. Current result on this branch: **4 pass, 11 fail**, every failure deterministic across three byte-identical runs.

The failures are the work list rather than a verdict, and one root cause covers the two most serious:

**`InterceptorExecutor.cs:924` enforces the faithful-raw-writer rule only when no raw reader is supplied.** Supply a raw reader, as every generated subject does, and the writer is trusted and never verified. It is a shape check rather than a validation. That single line is why a raw writer substituting a foreign-owned subject commits with no exception, leaving the graph attached to the proposed subject while the field exposes the foreign one, and why `docs/design/tracking-lifecycle.md`'s claim that "the graph is left untouched" on a rejected write does not hold here.

Three things about the suite itself worth flagging, all recorded rather than quietly fixed:

- **The register on this branch was carried from #494 byte-identical and never adapted**, so several entries had become false. One claims a repro parks in the authoritative getter when it now parks in the raw store, which is the exact placement #494 measured and rejected because the reader blocks instead of racing. It is rewritten to cover all eleven classes.
- **Two instruments were lost in the rewrite**: the substituting device can no longer substitute, and two other fixtures are dead code that would be refused before their terminal ran.
- **Two repros were re-armed rather than landed verbatim**, because their phase predicates no longer fire here and verbatim they would have passed on a missing window rather than a defect. Both substitutions are recorded in the register.

One pre-existing failure on the untouched branch is excluded from the count, and `NormalizingSetterDerivedRaceTests.WhenDirectAliasObservesTwoConsecutiveReservedValues_ThenItConvergesToTheLatest` flakes 11 of 18 isolated runs, which is left in place and recorded.

Issue #508, opened 2026-09-02
Connector apply path cannot write a struct collection property


RicoSuter:
## What happens

Applying a `SubjectUpdate` to a subject whose collection property is declared as a struct collection (`ImmutableArray<T>`, `ArraySegment<T>`) throws as soon as the update actually changes the collection's structure:

```
System.InvalidCastException: Unable to cast object of type 'System.Collections.Generic.List`1[Node]'
to type 'System.Collections.Immutable.ImmutableArray`1[Node]'.
   at RegisteredSubjectProperty.SetValue(Object value)
   at SubjectUpdateApplyContext.SetPropertyValue(...)
   at SubjectItemsUpdateApplier.ApplyCollectionUpdate(...)
```

## Why

`SubjectItemsUpdateApplier.ApplyCollectionUpdate` builds a working list and, when the structure changed, writes it back through `ISubjectFactory.CreateSubjectCollection(property.Type, workingItems)`.

`DefaultSubjectFactory.CreateSubjectCollection` has exactly two shapes: arrays, and `List<>` of `propertyType.GenericTypeArguments[0]`. `ImmutableArray<T>` is not an array, so it takes the second branch and always produces a `List<T>`, which the generated setter cannot accept. The same applies to `ArraySegment<T>` and to any other collection type that is neither an array nor assignable from `List<T>`.

This needs a real conversion mechanism in the factory (materialize into the declared property type), not a special case per type.

## Reproduction

```csharp
[InterceptorSubject]
public partial class Node
{
    public partial string? Name { get; set; }
    public partial ImmutableArray<Node> Items { get; set; }
}

var context = InterceptorSubjectContext.Create().WithRegistry();
var source = new Node(context) { Items = [new Node { Name = "Item1" }, new Node { Name = "Item2" }] };
var target = new Node(context) { Items = [new Node { Name = "Other" }] };

target.ApplySubjectUpdate(SubjectUpdate.CreateCompleteUpdate(source, []), null, ChangeOrigin.Local);
```

Note that the item counts have to differ. With equal counts the applier takes the "update existing item" path, leaves `structureChanged` false and never writes, so the bug stays hidden.

## Scope

Independent of the default-struct-collection defect: the reproduction above uses initialised arrays throughout. Confirmed present on `master`.

RicoSuter:
Second instance of the same gap, found while testing a struct dictionary property.

[...]

PR #509, opened 2026-09-02
fix: read a default struct collection as empty instead of throwing

RicoSuter:
`ImmutableArray<T>` and `ArraySegment<T>` are structs whose `default` holds a null inner array, which the BCL treats as distinct from empty on purpose: enumerating or measuring one throws. A property of either type therefore starts life unusable, because a C# field defaults to `default`, and `subject.Cars = default;` is legal and puts it straight back there.

A default instance of those two types now reads as empty everywhere the framework enumerates a property value. It releases children exactly as an empty collection does, and does not throw.

Stacked on #507. That PR makes `Nullable<T>` structural properties classify by their underlying type, which widens this defect's exposure: on `master` a nullable immutable array is not structural, so a default assignment is a silent no-op, and after #507 it becomes structural and the same assignment throws. **The two should merge together.**

## Why

The day-one case needs no unusual code at all:

```csharp
[InterceptorSubject]
public partial class Garage
{
    public partial ImmutableArray<Car> Cars { get; set; }   // never assigned
}

new Garage(context);   // throws inside the lifecycle attach lock
```

And the exposure is wider than the type choice suggests, because an interface-declared property accepts a boxed default struct:

```csharp
public partial IReadOnlyList<Car> Cars { get; set; }   // ordinary declaration
garage.Cars = default(ImmutableArray<Car>);            // legal, throws today
```

Seven readers were found by instrumenting every code path that enumerates a property value and driving them from an uninitialised property. Five throw, one passes a boxed default through to a caller that then throws on `.Count`, and one takes down a HomeBlaze display panel. Notably `ValueToJsonNode` enumerates any `IEnumerable`, so a default `ImmutableArray<int>` reaches it even though it is not structural at all.

Every immutable *class* collection is unaffected, since their default is null and the existing null paths handle it.

## How the test works

`SubjectLookup.IsDefaultStructCollection(object?)` covers exactly `ImmutableArray<T>` and `ArraySegment<T>`, asking each through its own API (`IsDefault`, and a null `Array`) via a delegate closed over the element type once per runtime type. Everything else returns false, which is precisely how readers behaved before the guard existed and is therefore safe.

It deliberately does **not** ask a value whether it equals its own default, because that dispatches into the struct's `Equals`, which is consumer code. A struct collection with ordinary value semantics compares its contents, so on a default instance it dereferences the null inner array and throws: an equality-based guard reproduces the failure it exists to prevent. That matters beyond correctness, because the lifecycle reader calls this while holding its attach lock.

The narrow scope is the point. A consumer struct collection with the same property is not covered and should defend in its own accessors, which costs a `?? []` on the backing field. Inline arrays, fixed buffers, pointers, ref structs and negative zero are not cases to get right, they are simply out of scope.

## Contract

- A default `ImmutableArray<T>` or `ArraySegment<T>` reads as empty: it establishes no ownership edges, and assigning one releases whatever the property held.
- An initialised empty collection remains distinguishable from a never-set one, which is the distinction a nullable struct collection property exists to express.

## What this does not cover

**Dictionaries are out of scope in both directions.** Neither type in scope can be classified as a dictionary, so a default struct dictionary is not handled here and nothing about dictionary properties changes. A concrete declared dictionary type is untouched by this PR: it was already produce-only on the apply path and still is, for reasons unrelated to default instances (#508).

The seven sites are the complete set of readers that walk a property value looking for **subjects**. They are not the complete set of readers of a struct collection: a scalar-typed one, for example `ImmutableArray<string>`, still reaches connector serializers as a raw value through `SubjectUpdateFactory`'s value path and the per-connector converters. That is pre-existing and out of scope here.

## Also fixed

Two defects of the same shape, found while reviewing the guard sites:

- The JSON writer threw on a default struct collection both when enumerating it and when serializing it as a plain value. It now emits what an initialised empty one emits.
- The JSON path resolver threw `ArgumentOutOfRangeException` for an index past the end of a collection.

## Breaking changes

None. `SubjectLookup.IsDefaultStructCollection(object?)` is added, so the public API snapshot moves by one line. No signature changes, and no shape that previously worked behaves differently: every affected shape previously threw.

## Performance

The guard sits behind the existing `CanContainSubjects` gate that scalar writes already return at, so scalar writes pay nothing. In the lifecycle reader it sits in a `ValueType` arm placed after the null and subject arms, so a subject reference, the dominant structural write, keeps its exact previous cost. Reference-typed collections pay one `isinst`; only boxed structs reach the predicate. The connector and lookup entry points keep their inlined reference-type fast paths and route boxed structs to their existing slow paths.

Not benchmarked. The affected classes are the lifecycle and registry ones if that is wanted before merge.

## Verification

- [x] Documentation updated
- [x] Unit tests
- [ ] ~~Integration tests~~ no connector behaviour changed; the connector guards convert a throw into an empty read
- [ ] ~~Benchmarks~~ see above
- [ ] ~~Load tests~~ no connector implementation changed
- [ ] ~~Chaos tests~~ no connector implementation changed

Twenty tests across Tracking, Connectors, Registry and HomeBlaze; full suite 3330 passing across 26 assemblies. Every guard site was demonstrated rather than assumed: with the value test forced to return false, fifteen tests fail and every one of the seven sites is among them. The two JSON fixes were each reverted individually and confirmed to fail their own test.

Net production change is +137 / -8 lines, 73 of the additions being code rather than comments.

### DIFF
--- a/src/HomeBlaze/HomeBlaze.Host.Services.Tests/Display/StateUnitExtensionsTests.cs
+++ b/src/HomeBlaze/HomeBlaze.Host.Services.Tests/Display/StateUnitExtensionsTests.cs
@@ -1,3 +1,4 @@
+using System.Collections.Immutable;
 using HomeBlaze.Abstractions.Attributes;
 using HomeBlaze.Host.Services.Display;
 using HomeBlaze.Services.Lifecycle;
@@ -121,6 +122,22 @@ public class StateUnitExtensionsTests
 
         // Assert
         Assert.Equal(expected, result);
+    }
+    [Fact]
+    public void WhenDisplayValueIsADefaultStructCollection_ThenItRendersAsEmpty()
+    {
+        // Arrange: a property that was never assigned holds a struct collection whose inner array
+        // is null, so joining it throws instead of producing the empty string an empty one gives.
+        var context = CreateContext();
+        var subject = new DisplayTestSubject(context);
+        var registered = subject.TryGetRegisteredSubject()!;
+        var property = registered.TryGetProperty(nameof(DisplayTestSubject.Rate))!;
+
+        // Act
+        var result = property.GetPropertyDisplayValue(default(ImmutableArray<string>));
+
+        // Assert
+        Assert.Equal(string.Empty, result);
     }
 }
 
--- a/src/HomeBlaze/HomeBlaze.Host.Services/Display/StateUnitExtensions.cs
+++ b/src/HomeBlaze/HomeBlaze.Host.Services/Display/StateUnitExtensions.cs
@@ -3,6 +3,7 @@ using HomeBlaze.Abstractions.Attributes;
 using HomeBlaze.Abstractions.Metadata;
 using HomeBlaze.Services;
 using Namotion.Interceptor.Registry.Abstractions;
+using Namotion.Interceptor.Tracking;
 
 namespace HomeBlaze.Host.Services.Display;
 
@@ -103,7 +104,11 @@ public static class StateUnitExtensions
             DateTime dt => dt.ToString("g"),
             DateTimeOffset dto => $"{dto.ToLocalTime().ToString("g")} {dto.ToLocalTime():zzz}",
             Enum e => e.ToString(),
-            IEnumerable<string> strings => string.Join("\n", strings),
+            // A default struct collection throws when joined; it renders as the empty join an
+            // initialized empty collection produces.
+            IEnumerable<string> strings => SubjectPropertyTypeExtensions.ReadsAsEmpty(value)
+                ? string.Empty
+                : string.Join("\n", strings),
             _ => value.ToString() ?? ""
         };
     }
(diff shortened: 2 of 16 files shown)


RicoSuter:
Closing this. Measurement on the base branch shows it is not worth its cost.

The behaviour it set out to fix is already fail fast. Both routes that could put a default struct collection into an attached subject throw immediately, deterministically, with the BCL's own message:

```
CONSTRUCT (lifecycle):  InvalidOperationException
CONSTRUCT (registry):   InvalidOperationException
ASSIGN default later:   InvalidOperationException
  "This operation cannot be performed on a default instance of ImmutableArray<T>.
   Consider initializing the array, or checking the ImmutableArray<T>.IsDefault property."
```

There is no silent corruption, no race, and no wrong answer. The author hits it on the first run of their own code, and the message names the one character fix. This PR would have converted that into silent tolerance for 129 lines of production code and 20 tests.

The one case that could have justified it does not exist. The concern was that #507 makes `Nullable<T>` structural properties get scanned, so a previously working model could start throwing. Measured: a nullable holding null constructs fine, and null is what the nullable form uses for "not set". Only an explicit `default(ImmutableArray<T>)` inner value throws, which the non-nullable form already does on master. #507 makes the two forms consistent rather than introducing a regression.

Scope for anyone reading later: this was narrowed from a general field-walking mechanism (unsound, silently dropped children for inline arrays and fixed buffers) to an allowlist of the two BCL types, and then to nothing. The struct dictionary apply-path failure is tracked separately in #508.

#507 merges on its own, with the fail fast behaviour documented in its description as intended.

Issue #516, opened 2026-09-03
feat: Make the libraries Native AOT compatible


RicoSuter:
Native AOT is not a supported target today and nothing declares it: no `PublishAot`, `IsAotCompatible` or `IsTrimmable` anywhere in the tree. This tracks getting there, and records what stands in the way so the question does not have to be re-derived in each pull request that brushes against it.

The immediate reason to write it down: reviewing #514 turned on whether putting `Expression.Compile()` on the common path of a keyed lookup was acceptable. It was avoidable there, cheaply, and avoiding it shaped the final design. Without a stated position that decision has to be argued from scratch every time.

## What blocks it

`Expression.Compile()`, `MakeGenericMethod` and `MakeGenericType` over value types, and `Activator.CreateInstance` on a type the trimmer cannot see are all annotated `RequiresDynamicCode` or `RequiresUnreferencedCode`. Under Native AOT they throw or degrade to interpretation, and they raise IL3050 or IL2026 under AOT analysis.

Known sites in production code:

| File | Construct |
|---|---|
| `Namotion.Interceptor.Connectors/DefaultSubjectFactory.cs` | `Activator.CreateInstance`, `MakeGenericType` |
| `Namotion.Interceptor.Tracking/SubjectLookup.cs` | `Expression.Compile` |
| `Namotion.Interceptor.Registry/SubjectRegistryExtensions.cs` | `Expression.Compile` |
| `Namotion.Interceptor.Tracking/Change/Performance/InlineValueStorage.cs` | `MakeGenericMethod` |
| `Namotion.Interceptor.Dynamic/DynamicSubjectFactory.cs` | `Activator.CreateInstance` |
| `Namotion.Interceptor.Mqtt/JsonMqttValueConverter.cs` | reflection-based conversion |

The source generator is the reason this is tractable rather than hopeless: the declared shape of every `[InterceptorSubject]` type is known at compile time, so most of what is resolved reflectively at runtime could be emitted instead. #515 is one instance of that, for keyed dictionary access.

`Namotion.Interceptor.Dynamic` is the part that may never be fully AOT compatible, since defining properties at runtime is what it is for. Deciding whether it is out of scope, or supported through a generator-backed path, is part of this issue rather than a prerequisite.

## Suggested steps

1. Set `IsAotCompatible` on the libraries that are close today and let the analyzer report the truth, rather than guessing from greps like the table above.
[...]

PR #517, opened 2026-09-03
docs: state the Native AOT position for new code

RicoSuter:
States the Native AOT position for new code in `AGENTS.md`, so it does not have to be argued in each pull request that brushes against it.

Nothing changes mechanically: no project declares `PublishAot`, `IsAotCompatible` or `IsTrimmable`, and no analyzer is enabled. This records a direction and a default, not a gate.

Related to #516, which tracks getting to full compatibility and lists the sites that stand in the way.

## Why

Reviewing #514 turned on whether putting `Expression.Compile()` on the common path of a keyed dictionary lookup was acceptable. It was avoidable there at little cost, and avoiding it shaped the final design: ordinary dictionaries keep a path that needs no runtime code generation, and only a type observed to misbehave reaches the compiled delegate.

That was the right outcome, but it was reached by argument rather than by policy. The next change to touch this ground would have to have the same conversation from scratch.

## Contract

- New code avoids `Expression.Compile()`, `MakeGenericMethod` and `MakeGenericType` over value types, and `Activator.CreateInstance` on types the trimmer cannot see, where a static alternative exists at comparable cost.
- Existing sites are worth closing when a change already touches them, and worth leaving alone otherwise.
- The source generator is the preferred escape hatch, since the declared shape of an `[InterceptorSubject]` type is known at compile time.
- This is guidance, not a rule. A static alternative that costs a measurable amount on a hot path, or a large amount of complexity, is not automatically the right trade; the pull request should say which way it went and why.

## Breaking changes

None. Documentation only.

## Diff composition

|---|---:|---:|---:|---:|
| Documentation | 1 | 10 | 0 | +10 |
| **Total** | **1** | **10** | **0** | **+10** |

## Verification

- [x] Documentation updated
- [ ] ~~Unit tests~~ documentation only
- [ ] ~~Integration tests~~ documentation only
- [ ] ~~Benchmarks~~ documentation only
- [ ] ~~Load tests~~ documentation only
- [ ] ~~Chaos tests~~ documentation only

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -120,6 +120,16 @@ The library has specialized support for:
 - **GraphQL**: Real-time subscription support
 - **Blazor**: UI data binding components
 
+## Native AOT
+
+Native AOT is not supported yet and nothing declares it, but it is the direction, so do not add new obstacles to it. Full compatibility is tracked in issue #516.
+
+- **New code avoids dynamic-code constructs** where a static alternative exists at comparable cost. `Expression.Compile()`, `MakeGenericMethod` and `MakeGenericType` over value types, and `Activator.CreateInstance` on a type the trimmer cannot see are all annotated `RequiresDynamicCode` or `RequiresUnreferencedCode`: under Native AOT they throw or fall back to interpretation.
+- **Existing sites are worth closing when a change already touches them**, and worth leaving alone otherwise. Several exist today, listed in #516.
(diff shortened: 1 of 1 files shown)


RicoSuter:
Folding the AGENTS.md section into #514 instead, since that is the change the policy came out of.

Issue #524, opened 2026-09-04
Read-only dictionary implementing non-generic ICollection is silently dropped


RicoSuter:
A subject held in a read-only dictionary that also implements non-generic `ICollection` is silently dropped from the object graph. The children never attach: no lifecycle callbacks, no registry entries, no parent edges, no context inheritance, and connectors never publish them. No exception is raised and nothing appears in a log, so the property assignment looks like it succeeded.

## Repro

```csharp
public sealed class CarRegistry : IReadOnlyDictionary<string, Car>, ICollection
{
    private readonly Dictionary<string, Car> _cars = new();

    // Non-generic ICollection is what routes this to the wrong arm.
    public int Count => _cars.Count;
    public bool IsSynchronized => false;
    public object SyncRoot => this;
    public void CopyTo(Array array, int index) => ((ICollection)_cars).CopyTo(array, index);

    public IEnumerator<KeyValuePair<string, Car>> GetEnumerator() => _cars.GetEnumerator();
    IEnumerator IEnumerable.GetEnumerator() => GetEnumerator();
    // ... remaining IReadOnlyDictionary members
}
```

Assigned to a structural property, behind either a broadly declared (`object?`) or a precisely declared `IReadOnlyDictionary<string, Car>?` property. Measured: `IDictionary=False`, `ICollection=True`, `IsSubjectDictionaryType=True`, and the result is 0 children and 0 parents in both cases.

## Mechanism

`LifecycleInterceptor.FindSubjectsInProperty` dispatches in this order:

```
null -> IInterceptorSubject -> IDictionary -> ICollection -> string -> IEnumerable
```

The type implements non-generic `ICollection` but not non-generic `IDictionary`, so it matches the `ICollection` arm, which walks the value positionally:

```csharp
case ICollection collection:
    foreach (var item in collection)
        if (item is IInterceptorSubject subjectItem)   // a boxed KeyValuePair, so always false
```

Every item is a boxed `KeyValuePair<string, Car>`, which fails the subject test, so nothing is collected. The classifier says the type is a dictionary while the handler dispatches it ordinally, which is the same classifier and handler disagreement fixed for the `IEnumerable` arm in #520, left open one arm earlier.

## Scope

[...]

PR #527, opened 2026-09-08
fix: support lifecycle callback mutations and resolve PR #494 regressions

RicoSuter:
This review PR targets #494 (`rewrite/single-context-lifecycle`) and contains the callback-support implementation plus the correctness fixes found while validating it. Replacing a child supports finite same-context structural writes from lifecycle callbacks, delivers detach before attach, and keeps ownership, Registry projections and committed-write completion consistent under nested operations and failures.

It also removes unnecessary derived-projection restrictions, fixes connector update loss, and coordinates hosted startup with configuration. This remains a draft while sustained connector validation and benchmarks on the final code are outstanding. The original PR's migration from master is outside this delta.

Related to #494, #501 and #524. Master fixes #525, #526 and #528–531 are now inherited through refreshed #494; the extracted MQTT, method-discovery and reference-identity fixes are no longer separate changes in this delta.

## Changes

- **Lifecycle capture and recovery:** capture desired structural occurrences at storage commit and retain installed occurrences while seeding or reconciliation is incomplete. Release and old-side reconciliation use those captures without re-enumerating consumer collections. Fix interrupted seeding, nested getter writes, failed-attach cleanup of independently supported descendants, stale rollback after detach/reattach, and gate-wait transaction identity.
- **Registry:** reconcile duplicate occurrences, ordinal reshapes and dictionary rekeys using reference identity and indexed occurrence matching. Prevent obsolete or duplicate parent/child entries after nested changes.
- **Committed-write completion:** after a truthful storage commit, downstream failure still permits ownership reconciliation, property publication and derived recalculation. Attempt remaining dependents after one fails; preserve a single original error and aggregate additional completion errors after it.
- **Contract simplification:** computed derived getters may return detached or foreign-context subjects without owning them. Remove the orphan-result restriction and its deferred verdict/retry machinery. Metadata enumeration may perform same-context topology changes and is revalidated afterward. Re-registering the same singleton instance is idempotent; conflicting instances still reject. Direct late lifecycle registration rejects through every public registration path.
- **Concurrency and routing:** declared metadata controls structural routing, preventing narrowed generic arguments from bypassing it and boxed scalars from unnecessarily acquiring a topology gate. Framework synchronization protects ownership, Registry projections, queues and commit bookkeeping. Whole-accessor synchronization and concurrent non-atomic value access remain consumer responsibilities.
- **Wire updates:** reuse the existing change merger for repeated property changes so public partial-update construction does not lose collection insertions. Clear stale IDs on final null assignments. Validate final emitted references after processors, allowing overwritten intermediate values. Report missing referenced payloads instead of clearing an object or dropping an item silently.
- **Collections:** preserve keys for supported read-only dictionaries implementing non-generic collections; infer factory element types from actual implemented interfaces and reject incompatible default containers. Dictionary declarations require keyed enumeration; a declared typed subject sequence view remains valid for positional encoding.
- **Constructors and HomeBlaze:** correct constructor declaration-order and primary-constructor handling across partial declarations, with accessibility and supported method-target attributes. Reuse master's RootLoaded readiness signal and layout gate, with readiness published after the startup scope completes. The startup regression verifies the loading gate, connected storage and loaded file-backed content. Cancel RootLoaded even when .NET 10 cancels queued startup before ExecuteAsync can run.
- **Hosted startup:** replace the arbitrary 50 ms start delay with explicit startup scopes. Configuration and root publication finish before captured services start; incomplete scopes cancel captured starts. Open scopes allow unrelated ready service actions to continue. Detach and shutdown cancel pending generations and release their completion holds. Shutdown closes admission before cancellation callbacks so reattachment cannot leak a new hold, while shutdown retains responsibility for stopping attached services.

- **Review cleanup:** repair the boxed getter regression test so it executes and asserts its nested write. Share graph/attachment assertions and remove weaker duplicate cases while retaining cyclic, failure, and configuration coverage. Remove unused incoming-edge parameters, a forwarding method, and duplicate collection enumeration.

## Contract

- Finite same-context structural reentry is supported from lifecycle callbacks, getters, enumerators and write interceptors. Old-child detach notifications precede replacement-child attach notifications. The backing field has already changed when detach callbacks run.
- Nested writes normally settle ownership before returning; callbacks and Registry notifications may remain queued. A setter called from its own active seeding getter defers that property's reconciliation until the getter returns.
- Lifecycle work drains before each FIFO property-publication group, without inserting a lifecycle boundary between observers within one group. Events describe historical transitions; current graph queries can reflect later nested writes. Retained descendants may receive detach/attach transitions during reparenting, while executor claims survive queued detach delivery.
- Computed projections acquire no ownership. Stored derived properties remain structural edges. Selecting a projection for transport requires Registry metadata for its referenced payload, or a processor excluding it.
- Nested topology changes across different contexts still reject. Synchronously waiting for another thread's structural work while holding the topology gate remains unsafe. Timeout detection is heuristic: a 30-second blocked-holder window and a 300-second total bound cannot prove a first-use contract violation.
- Generated accessors do not synchronize the whole property operation. Old-value copies and changed-hook argument copies can tear under unsynchronized concurrent writes of non-atomic values, and reads with no read interceptors also remain unlocked. Consumers must coordinate relevant readers and writers, including background connector activity; an application lock only coordinates participants that acquire it.
- Stable authoritative getters, truthful commit metadata and valid stored graphs remain required. Arbitrary user side effects are not rolled back. Interrupted seeding can require later reconciliation or detach; raw in-place collection mutation remains untracked. Context configuration must precede concurrent use. Inline observer exceptions can still suppress later observers within their publication group.

Canonical details: [lifecycle design](docs/design/tracking-lifecycle.md), [generator](docs/generator.md), [hosting](docs/hosting.md), and [subject updates](docs/connectors-subject-updates.md).

## Breaking changes

These are additions and behavior changes relative to #494. No required interface members or API removals are added relative to the refreshed #494 target. The earlier draft-only LoadingCompleted proposal is replaced by master's existing RootLoaded task.

| Addition or behavior change | Consumer impact / migration |
|---|---|
| `HostedServiceStartupScope : IDisposable`, with `Complete()` and `Dispose()`; nullable `context.DeferHostedServiceStartup()` | Configure detached subjects before attachment, or complete a scope after configuration. Captured starts wait until all enclosing scopes complete and dispose. Disposal without completion cancels captured starts without detaching subjects. Do not await a captured start inside its open scope. `AddHostedSubject` and HomeBlaze initialization apply scopes automatically. |
| Eligible primary constructors receive context-taking generated overloads | Explicit matching constructors still win. Unsupported signatures remain skipped; handwritten constructors retain attachment responsibility. |
| Dictionary wire shapes and missing referenced payloads now report errors | Supply a supported keyed representation or custom factory, and register selected referenced subjects or filter the referencing property. Applying updates remains partial on error. |
| Same-instance singleton registration becomes idempotent; direct late lifecycle registration consistently rejects | Configure lifecycle before first attachment. A different reserved singleton instance still conflicts. |

## Diff composition

Measured at `d16b1a2c` against refreshed #494 `cb911c72`: production **1,730 added / 1,208 removed lines, net +522 across 50 files**. The latest review follow-up fixes startup scheduling and shutdown admission, repairs the getter witness, and consolidates tests. Its cleanup commit removes 216 lines net; together with the correctness fix the follow-up adds 132 lines net (86 production and 46 test lines). The optional wide-value synchronization expansion remains excluded.

|---|---:|---:|---:|---:|
| HomeBlaze.Services | 2 | 33 | 14 | +19 |
| HomeBlaze.Storage | 1 | 11 | 10 | +1 |
| Namotion.Interceptor | 10 | 108 | 74 | +34 |
| Namotion.Interceptor.Connectors | 8 | 119 | 21 | +98 |
| Namotion.Interceptor.Generator | 2 | 32 | 22 | +10 |
| Namotion.Interceptor.Hosting | 4 | 229 | 36 | +193 |
| Namotion.Interceptor.Registry | 1 | 78 | 66 | +12 |
| Namotion.Interceptor.Tracking | 22 | 1,120 | 965 | +155 |
| Tests and benchmarks | 75 | 6,597 | 659 | +5,938 |
| Documentation | 5 | 95 | 48 | +47 |
| **Total** | **130** | **8,422** | **1,915** | **+6,507** |

## Verification

Current candidate: `d16b1a2c844794a6e2d3f4df91be5b4b0bfcd416`. Target base: `cb911c728cc9ee4349897d1ef148750819f8c0e4`, containing master `f2a6efbd`. Separate startup and cleanup reviews plus a final integrated review covered every follow-up change. Findings about shutdown admission and graph-oracle error aggregation were fixed and re-reviewed with no actionable findings remaining.

- [x] Fresh full solution Release non-integration tests on the final source: **4,438 passed across 34 nonempty suites**, zero failures/skips.
- [x] Hosting: **30 passed**, including independent-flow startup/stop, scope order, detach/reattach, pending-start shutdown, and completion-hold cleanup. The blocked-start regression and four shutdown-admission/stop-ownership cases failed before their fixes and passed afterward.
- [x] Cleanup validation: Core **156**, Tracking **879**, Registry **197**, and Connectors **860** passed. The boxed getter's positive witness failed on its old trigger; the corrected owned-seeding trigger passes. All 12 callback graph-oracle theory cases remain and pass.
- [x] A deterministic isolated shutdown probe forced one scoped start into the pending list and another to remain buffered behind a blocked service. Both canceled, no scoped service started after scope release, and all three holds were released.
- [x] Public API snapshots passed unchanged; this follow-up adds no public API or generator changes. The broader PR's earlier draft-only LoadingCompleted proposal remains replaced by master's RootLoaded signal.
- [x] Hosted CI passed at `d16b1a2c`: unit tests plus MQTT, OPC UA, WebSocket and HomeBlaze integration jobs. [run 34333081439](https://github.com/RicoSuter/Namotion.Interceptor/actions/runs/34333081439).
- [ ] Benchmarks: final code not measured. Earlier support measurements remain historical and do not certify this candidate's performance.
- [ ] Connector Tester load tests: sustained throughput/allocation and memory validation remains outstanding.
- [ ] Connector Tester chaos tests: a fresh master/candidate matrix remains outstanding.

Historical validation at `a7d535d4`: 4,431 unit tests, HomeBlaze Services 236, browser startup E2E 1, and integration suites MQTT 21, OPC UA 43, WebSocket 55 and HomeBlaze browser 27 passed. Hosted CI also passed at that prior head ([run 34324939692](https://github.com/RicoSuter/Namotion.Interceptor/actions/runs/34324939692)). Those integration results are not presented as fresh local runs for this follow-up.

Final local validation used Release `dotnet test` with `--no-restore -m:1 -p:UseSharedCompilation=false`, filtering `Category!=Integration`, on the existing reviewed restore graph. The unchanged local NuGet plugin dependency graph emits visible NU1603 warnings. No repository package declarations changed. The previously fixed RootLoaded canceled-start path and browser readiness checks remain covered by their existing tests; sustained connector verification and final performance measurement remain separate merge gates.


RicoSuter:
Superseded and closing. Everything here has a home, recorded so nothing is lost.

## Where it went

| Workstream | Now in | Note |
|---|---|---|
| Lifecycle callback support, correctness fixes, reductions | **#538** | same base as this PR; production net +441 against +522 here |
| Subject update wire integrity | **#537** | against `master`, since every file it touches is byte-identical between master and this base |
| Performance recovery | **#578** | new, stacked on #538; the regression this PR carried is now measured and largely recovered |
| Hosted startup scope | already on `master` | #532 and #534, in a simpler form: `Complete()` removed and `Dispose()` unconditional, which also removes a startup wedge this PR still had |
| Storage accessor, RootLoaded cancellation, browser startup test | **#579** | two are defects live on master today, neither was rehomed |

## What the review changed

Six defects were confirmed and fixed in #538, three of which this PR introduced rather than inherited: `ResumeFailedSeed` was reachable from only one of three routes, leaving subjects permanently orphaned; the generated context constructor silently displaced a declared one; and `AggregateException.Flatten()` is breadth first, so the completion failure landed ahead of the write failure, the opposite of what the change claimed.

Two of the removal justifications were also wrong and are corrected in #579: master's #530 filters on `property.Type.IsInterface` and so does not cover `IBlobStorage Client`, and master never received the `RootManager.StartAsync` cancellation fix.

## On the deleted benchmark record

`SUPPORT-V3-REPORT.md` was removed by `2580e15f1` along with fifteen other spike reports. Deleting scaffolding is right, but it took the only measurement with it, and that measurement recorded structural writes 17.6 to 32.5 percent slower. It was also stale: its production baseline was `ffdaf2f4`, 44 commits before this PR's head, and the one row it showed improving is the row that had regressed by the time the head landed.

Re-measured properly on a pinned CPU, the cost was 24 to 52 percent on structural writes and 7.6 percent on scalar writes. #578 brings scalar to parity and single child replacement below base, and cuts the collection and subtree rows to roughly 15 percent. Those numbers are in #578 rather than in a file that can be deleted.

No code is lost by closing this.

PR #534, opened 2026-09-09
fix: simplify the hosted startup scope and close its shutdown gaps

RicoSuter:
Simplifies the startup scope introduced in #532: disposing it releases captured starts, including during exception unwinding. Services and consumers remain responsible for configuration validity. Also fixes older handler defects: detached services whose stops are queued now receive `StopAsync` during shutdown, and attach/detach continuations no longer occupy the action loop.

- Supersedes the startup scope API introduced in #532
- Follow-up: #440

## Contract

- Captured starts wait for disposal of their scope and all enclosing scopes. Do not await them inside an open scope.
- Dispose scopes in reverse creation order in the creating flow. Incorrect or repeated disposal does not throw, but ambient-scope cleanup is not guaranteed for misuse. Undisposed scopes hold starts until host shutdown; there is no timeout or automatic unwinding.
- A stop drained during shutdown receives a cancelled token. Successful completion remains successful; `OperationCanceledException` cancels the awaiting caller.
- `RootLoaded` is signalled before root loading releases deferred starts.
- Shutdown still stops attached services whose deferred start never ran. Changing that requires tracking which services started and belongs with #440.

## Breaking changes

None.

The startup scope has never shipped. The last release is v0.9.2 (2026-08-31), #532 merged after it, and `HostedServiceStartupScope` is absent from that tag and from its public API snapshot. `Complete()`, the cancel-on-failure behaviour and the out-of-order `InvalidOperationException` therefore only ever existed on `master`. This reshapes an unreleased API rather than breaking a released one, and consumers meet the scope for the first time in the shape that lands here.

For the release notes: this pull request and #532 fall between the same two tags, so the startup scope should be described once, in the shape that lands here. #532's description documents `Complete()`, a scope that cancels its captured starts when it is disposed without completing, and an `InvalidOperationException` on out-of-order disposal. None of those ship, and none of them need a migration step.

For anyone tracking `master` rather than a release: drop the `Complete()` call, since a `using` alone now does what `using` plus `Complete()` did. A scope disposed without completing no longer cancels its captured starts, so suppress a hosted service by not attaching the subject or by detaching it, rather than by abandoning the scope. `Dispose()` no longer throws `InvalidOperationException` on out-of-order disposal.

## Diff composition

|---|---:|---:|---:|---:|
| Production code | 6 | 66 | 63 | +3 |
| Tests | 3 | 109 | 53 | +56 |
| Documentation | 1 | 5 | 4 | +1 |
| **Total** | **10** | **180** | **120** | **+60** |

## Verification

- [x] Documentation and public API snapshot updated.
- [x] At `739dd81`: all CI test jobs passed; Hosting 30 and HomeBlaze Services 229 also passed locally. The prior review reports 3,967 passing non-integration tests.
- [x] Wording cleanup: executable content unchanged apart from the test name; all 30 Hosting tests passed again.
- [x] CI integration suites passed: HomeBlaze, MQTT, OPC UA and WebSocket.
- [ ] ~~Benchmarks~~ Not run; no performance claim.
- [ ] ~~Load and chaos tests~~ Not applicable.

Positive controls reported in the preceding review failed when reverting stop-delay cancellation handling, asynchronous attach continuations, or the action loop's ambient-scope reset. The irregular-disposal test checks that captured and later services start, not that the ambient scope is cleared. Cancellation-result handling, asynchronous detach continuations, stop invocation outside the lock, deferred-start identity checks and root-readiness ordering lack discriminating regression tests.

Deferred cancellation and disposal remain under the handler lock. Moving them outside it introduced a start-after-detach race and `ObjectDisposedException`; that change was reverted. The ownership redesign remains in #440.

### DIFF
--- a/docs/hosting.md
+++ b/docs/hosting.md
@@ -152,11 +152,11 @@ using (context.DeferHostedServiceStartup())
 }
 ```
 
-The scope follows the current execution flow for that context. Starts wait until the scope and all enclosing scopes are disposed, so the `using` is the whole contract. An exception unwinding through it releases the captured starts like any other exit: the scope decides when a start runs, not whether the subject is fit to run, so a subject whose configuration threw still starts and reports its own failure. The extension returns null when hosted services are not configured. Do not await an attaching service's startup inside its open scope, since that startup is waiting for scope disposal.
+The scope follows the current execution flow for that context. Captured starts wait until it and all enclosing scopes are disposed, including when configuration throws. Services and consumers remain responsible for configuration validity. The extension returns null when hosted services are not configured. Do not await a captured service's startup inside its open scope.
 
-Scopes should be disposed in reverse creation order in the execution flow that created them. Nothing enforces it: disposal out of order, from another flow, or more than once releases the scope's own starts and is otherwise ignored, so a misuse is silent rather than loud. A scope that is never disposed at all holds the starts it captured for the lifetime of the host.
+Dispose scopes in reverse creation order in the creating flow. Incorrect or repeated disposal does not throw, but ambient-scope cleanup is not guaranteed for misuse. An undisposed scope holds captured starts until host shutdown.
 
-`AddHostedSubject<T>()` holds a scope through its configuration callback. HomeBlaze deserialization uses this scope while populating configuration. Root loading holds an enclosing scope until the root is published, attached, registered as a context service and signalled through `RootLoaded`, so a service that waits on that task inside its own `StartAsync` is not racing it.
+`AddHostedSubject<T>()` defers startup through its configuration callback. HomeBlaze deserialization defers it through configuration population; root loading also waits until the root is published, attached, registered as a context service and signalled through `RootLoaded`.
 
 ## Deferred Starts and Startup Completion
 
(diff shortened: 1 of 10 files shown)


docs/hosting.md
```diff
@@ -145,17 +145,18 @@ await person.AttachHostedServiceAsync(
...
+The scope follows the current execution flow for that context. Captured starts wait until it and all enclosing scopes are disposed, including when configuration throws. Services and consumers remain responsible for configuration validity. The extension returns null when hosted services are not configured. Do not await a captured service's startup inside its open scope.
 
-`AddHostedSubject<T>()` holds a scope through its configuration callback. HomeBlaze deserialization uses this scope while populating configuration. Root loading holds an enclosing scope until the root is published, attached and registered as a context service.
+Dispose scopes in reverse creation order in the creating flow. Incorrect or repeated disposal does not throw, but ambient-scope cleanup is not guaranteed for misuse. An undisposed scope holds captured starts until host shutdown.
+
+`AddHostedSubject<T>()` defers startup through its configuration callback. HomeBlaze deserialization defers it through configuration population; root loading also waits until the root is published, attached, registered as a context service and signalled through `RootLoaded`.
```
RicoSuter: Remove homeblaze information from docs (whole paragraph?) also rewrite these docs (also what 532 added) in a way consumers understand what this is for (not exposing implementation details, just consumer contract)

PR #537, opened 2026-09-11
fix: complete subject update payloads and explicit update modes

RicoSuter:
Subject updates could reach a receiver incomplete or be applied inconsistently. A subject attached at runtime could arrive without its unchanged properties and attributes, a receiver kept stale collection members, and a subject named twice in one update became an empty second subject. This fixes the producer and the C# receiver and adds one optional field to the wire format.

- **Producer.** Every subject a change assigns or inserts, and every subject the update mentions for the first time while that payload is built, arrives with its complete property set and attributes in any arrival order within a batch. Every change states its path from the root. Repeated changes to one property are merged, a stale object id is cleared, a property referencing a subject without Registry metadata is omitted with one warning, computed subject projections are not published, and a change whose first-parent path crosses an excluded property is dropped.
- **Receiver.** Each id names one subject for the rest of an update, each existing instance takes the payload of at most one id, and instance creation is bounded per id and declared type. A collection or dictionary stating its whole membership removes stale members and creates missing ones. A property without a setter drops only its own value and reports structure it cannot store in one warning. An id with no payload fails that property while its siblings apply.
- **Wire.** `SubjectPropertyUpdate` gains `mode`, a second dimension beside `kind`: `Complete` on a collection or dictionary whose items state the whole membership, and `Replaced` on an object reference that now holds a different subject. `Incremental`, the default, is omitted, so existing messages are unchanged and an applier that ignores the field behaves as before.

Related to #527

## Why

Ids are minted per update, so an id that already exists only means something earlier in the same update referenced the subject, not that the receiver has it. A change inside a newly attached subject that was processed before the reference or insert attaching it, for example a write from a property initializer during attach, shipped the subject as a bare id plus its changed properties, and a receiver gating rendering on an attribute set at attach could never render it.

The receiver cannot tell two situations apart from the entries alone: a collection entry that states the whole membership versus one that only carries changed items, and an object entry whose reference was reassigned versus one that is only the path to a change below it. Reconciling membership or replacing a shared instance on a guess corrupts the receiver in the other case, so the producer now says which one it means.

### Wire

- `mode` is `Incremental` (omitted), `Complete` (Collection and Dictionary entries) or `Replaced` (Object entries). A mode that does not apply to the entry's kind is treated as `Incremental`.
- `Complete`: the items state the whole membership, with no operations, a `count` equal to the items, and every position or key named once with a payload. A `Complete` entry that does not meet that shape applies as `Incremental`.
- `Replaced`: the reference now holds a different subject, so its payload never applies to the instance the receiver holds there.

### Producer

- A subject a change assigns or inserts arrives complete, and so does every subject the update mentions for the first time while that payload is built. Items sent in full and inserted items are always completed; an object reference to a subject the update already carries is expanded only when it is that subject's first parent. Each subject is completed at most once per update.
- Every change carries its path from the root along first parents, and every id in an update has a payload.
- `Complete` is set on every collection or dictionary written in full, in complete updates and in the payloads of new subjects. `Replaced` is set when a reference change ends on a different, non-null subject, never in complete updates.
- Computed derived properties that hold subjects are not published. Exclusion follows the first-parent path.
- Move operations arrive as a complete permutation, one for every retained item whose position changed.

### Receiver

- An id is one subject for the duration of an update: later references resolve to the subject it first names, and no binding outlives the update. A position whose declared type the bound subject does not fit gets one instance per id and type.
- An existing instance takes the payload of at most one id per update. A position holding an instance that already took another id's payload is treated as empty, so a complete update repairs a receiver that still shares an instance the source has split.
- Membership is reconciled only for `Complete`: unlisted members are removed, missing ones created, and retained instances keep their unspecified properties. `count: 0` creates an empty container.
- A property the model cannot write drops only its own value and any new membership. Dropped structure is reported once per update; dropped values are not.
- An id with no payload fails that property without a replacement. Application is not transactional.

## Remaining limitations

- An insert and a move in one collection change can corrupt the receiver, because the move's `fromIndex` does not account for the insert.
- A change to an attribute of an object reference, arriving before a change inside the referenced subject, loses the child's change.
- When first parents form a loop, for example a child's back reference after its parent moved, a change below it cannot be stated from the root and does not arrive.
- Exclusion follows the first-parent path, so a subject also referenced through an included property is still published with its values.
- Complete membership matches by position or key rather than by subject identity. When a completed subject's collection is applied to an instance the receiver already holds, an item the source replaced has its payload written into the instance that held that position, and every other position sharing that instance sees the change. `Replaced` covers this for references; there is no equivalent marker for a collection position.
- A reference inside a complete payload can name a subject the update carries only a path entry for; when the receiver meets that reference first, it creates the subject from the path entry alone, so the instance lacks what the path does not state.
- There is no subject identity across updates.
- Inserting many items that each change in the same batch costs quadratic time.

## Breaking changes

API: additive only. `SubjectPropertyUpdate.Mode` and the `SubjectPropertyUpdateMode` enum are new, and the public API snapshot changes by exactly those.

Behavior:

- Partial updates carry complete payloads for subjects nested inside newly assigned or inserted subjects.
- JSON and update snapshots gain `"mode"` on complete collections and dictionaries and on replaced references.
- A property referencing a subject without Registry metadata is omitted with a warning, where it previously carried an id no payload described.
- Computed derived properties that hold subjects are no longer published, and updates naming one are ignored.
- A change reached only through an excluded first-parent edge is dropped entirely.
- The receiver removes stale members of a `Complete` collection or dictionary and creates an empty container for `count: 0`.
- A repeated id within one update resolves to one subject, where a second reference produced a new subject without properties, and a reference back to the root points at the receiver's root.
- A value for a property without a setter is no longer converted, and `transformValueBeforeApply` no longer runs for it.
- An id without a payload fails that property and keeps its value, where an insert was skipped and a reference set to null.
- A sparse item update at a position holding `null` creates the item, where it failed with a `NullReferenceException`.
- Id numbering and key order within an update can differ.

Contract: another applier, such as a TypeScript one, keeps its current behaviour while it ignores `mode`, and gains both fixes by adopting two rules: reconcile membership only for `Complete`, and never apply a `Replaced` reference's payload to the instance it holds.

## Diff composition

|---|---:|---:|---:|---:|
| Production code | 10 | 1'002 | 201 | +801 |
| Tests and benchmarks | 49 | 3'014 | 37 | +2'977 |
| Documentation | 1 | 101 | 12 | +89 |
| **Total** | **60** | **4'117** | **250** | **+3'867** |

36 of the test files are update snapshots whose only change is an added `Mode:` line.

## Verification

Every new rule has a test that fails when the rule is reverted. The tests for bounded instance creation use a factory that stops after a few creations, so a regression fails instead of exhausting the stack. Round trips through JSON cover the producer and receiver together, including shared subjects, reassigned references and re-attached subjects with shrinking membership.

- [x] Documentation updated
- [x] Unit tests (`dotnet test src/Namotion.Interceptor.slnx --filter "Category!=Integration"`): 34 suites, 4182 tests
- [x] Integration tests: `Namotion.Interceptor.WebSocket.Tests` including integration, 160 tests
- [ ] Benchmarks: not run yet. A rough in-process timing, not BenchmarkDotNet, put update creation and application about 20 to 25 percent higher in CPU than `master`, with identical allocations. The largest single cost is merging a batch again that `ChangeQueueProcessor` has already merged. `SubjectUpdateBenchmark` should run before merge.
- [ ] ~~Load tests~~ not run
- [ ] ~~Chaos tests~~ not run


src/Namotion.Interceptor.Connectors/SubjectFactoryExtensions.cs
```diff
@@ -31,19 +36,39 @@ internal static IInterceptorSubject CreateCollectionSubject(this ISubjectFactory
...
-            itemType = propertyType.GenericTypeArguments[0];
+            itemType = GetCollectionTypes(propertyType, propertyType.IsSubjectDictionaryType()).Element;
         }
 
         return subjectFactory.CreateSubject(
             itemType ?? throw new InvalidOperationException("Unknown collection element type"),
             serviceProvider);
     }
+
+    internal static (Type? Key, Type Element) GetCollectionTypes(Type propertyType, bool dictionary = false)
```
RicoSuter, line 49: Remove dictionary parameter and calc internally? Or not possible?

RicoSuter:
Correction to my previous comment: it described a push that was rejected, because your two commits had landed in the meantime. I reset onto them and re-applied, so what is on the branch now is different from what that comment said. Disregard it in favour of this one.

## What landed on top of your two commits

**The source no longer publishes a computed derived property that holds subjects.** Such a property is a projection: it owns nothing, the subjects behind it are often outside the graph and contribute no payload, and an id that describes nothing is worse on the wire than saying nothing at all.

The setter test asks whether the value is stored on this side, which is what separates a projection from a real edge, and not whether a receiver could write it. A stored `[Derived]` property carries an ordinary edge and is still published. Value typed derived properties are published too, because a receiver may display them or rely on this side to compute them, and they carry no reference that can dangle.

**The applier ignores any property this model cannot write**, for every update rather than only this shape. Every branch of the apply ends in a write, so a property with no setter could only fail. Whether a property is writable is a fact about this model rather than the producer's, and a producer may legitimately publish one for a receiver to display, so this is an expected shape rather than a malformed update. The applier previously had no writability check at all.

**Apply stays strict about a missing payload.** With projections no longer published, such a reference can only arrive from a peer that sent something it could not describe, which is a real protocol error rather than a shape this side produces. I had a tolerant version of this ready and dropped it: once the source stops emitting projections, the strict tests pass unchanged and the looser behaviour buys nothing.

### One shape worth knowing about

`IsDerived` is literally just "carries `[Derived]`", so it does not by itself distinguish stored from computed. A stored `[Derived]` property that holds subjects would carry a real edge, and skipping it would lose data a receiver could apply. There is no such property in the repository today (the two stored ones, `DerivedWithSetter` and `Nickname`, are both `string?`), but that is why the predicate tests the setter rather than `[Derived]` alone.

### Verification

Release build clean, 0 warnings. `Namotion.Interceptor.Connectors.Tests` **871 passed**, full non-integration solution **4,047 passed, 0 failed** across 34 assemblies. The omission pass is still load bearing and I confirmed it by mutation: disabling it fails 3 tests, because a reference can still dangle when a later change in the same batch detaches a subject an earlier change referenced, which is a writable property and so not covered by the source-side skip.

src/Namotion.Interceptor.Connectors/Updates/Internal/SubjectItemsUpdateApplier.cs
```diff
@@ -180,9 +209,9 @@ internal static void ApplyDictionaryUpdate(
             {
                 var key = ConvertDictionaryKey(collUpdate.Index, targetKeyType);
 
-                if (collUpdate.Id is not null &&
-                    context.Subjects.TryGetValue(collUpdate.Id, out var itemProps))
+                if (collUpdate.Id is not null)
                 {
+                    var itemProps = context.GetSubjectProperties(collUpdate.Id);
```
RicoSuter: avoid abbreviations in whole pr, here: rename to itemProperties

src/Namotion.Interceptor.Connectors/SubjectFactoryExtensions.cs
RicoSuter: Calculating it internally would be wrong at one call site: in `SubjectItemsUpdateApplier.ApplyDictionaryUpdate` the flag is what the caller requires, not what the type is. When the producer sends a dictionary update for a property this model declares as a positional collection, forcing the dictionary shape is what produces the documented `NotSupportedException`; auto-detection would return a null key type and the caller would null-deref instead. `CreateSubjectCollection` likewise asks for the collection shape deliberately, so a dictionary-typed property surfaces as a mismatch rather than being silently reinterpreted.

Removed the boolean from the API instead: `GetCollectionElementType` and `GetDictionaryKeyAndValueTypes` state the intent per call site, the flag stays private in `GetContainerTypes`, and the dictionary accessor is now the only place that resolves the nullable key, which also removed the three `Key!` assertions. In a4296d297.

src/Namotion.Interceptor.Connectors/Updates/Internal/SubjectItemsUpdateApplier.cs
RicoSuter: Done in d5867201d: itemProps is now itemProperties and collUpdate is now itemUpdate, and no abbreviated locals remain in the update code this pull request touches.

PR #538, opened 2026-09-11
fix: support lifecycle callback mutations with a minimal production delta

RicoSuter:
Replaces #527. Same lifecycle work, with the three bundled workstreams removed, the correctness holes found in review fixed, and the production surface cut.

Targets #494 (`rewrite/single-context-lifecycle`).

## Why this rather than #527

#527 carried five workstreams. Three of them no longer belong here:

| Removed | Where it belongs now |
|---|---|
| Hosted service startup scope | Already on master as #532 and #534, in a simpler form. `Complete()` is gone and `Dispose()` is unconditional, which removes a startup wedge #527 still had. |
| Subject update wire integrity | #537, against master. Every file it touches is byte-identical between master and this base, so it never needed this stack. |
| Storage client accessor rename | A workaround for the predicate #530 already fixed generically on master. Dropped. |

One narrow piece of the hosting work is kept: `RootManager`'s anchor guard. Master does not have it and this stack needs it, or HomeBlaze throws `InvalidOperationException: The subject is already explicitly attached to this context`.

## Production size

| | Files | Added | Removed | Net |
|---|---:|---:|---:|---:|
| #527 against this base | 50 | 1,730 | 1,208 | **+522** |
| This PR | 39 | 1,301 | 1,210 | **+91** |

The reduction is 306 lines from dropping the three workstreams and a further 125 from the simplifications below, so the callback support contract lands for roughly 90 net production lines.

## Correctness fixes found in review

**A failed seed is now resumed on every route that reaches its subject.** `ResumeFailedSeed` was reachable only from the reconciler. Two other routes reached a subject whose seed had failed and left it permanently orphaned: no ownership, no context attachment, zero references, no Registry entry, and a baseline disagreeing with storage forever. The resume moved into `AttachEdge`, so every edge-recording route is covered by construction rather than by each caller remembering. Four cases added to `FailedSeedRecoveryTests`, two of which fail without the fix.

**The generated context constructor no longer displaces a declared one.** Suppression keyed on arity 1 only, so a hand-written `Subject(IInterceptorSubjectContext, bool initialize = true)` was left in place while the generator emitted the arity-1 overload that wins overload resolution. The user's constructor was silently never called. Suppression now also covers a public constructor whose first parameter is exactly the context and whose remaining parameters are optional or params. Ten cases cover the displacing tails and the shapes that must not suppress: `ref`, `out`, `in`, `params`, a concrete context class, and a derived context interface. Suppressing for `in` would trade silent displacement for silent context loss, so it does not.

**Derived completion failures report one flat aggregate** instead of a nested one that changed the caller-visible exception type.

## Simplifications

| Change | Lines | Why it is safe |
|---|---:|---|
| Property edge journal backed by a plain occurrence list | -116 | The intrusive linked list, free list and per-subject chain bought O(1) `RemoveLast` only, and `SubjectOwnership.RemoveIncoming` already pays the same reverse scan on every removal. |
| Lifecycle callback reentrancy guard removed | -64 | Unreachable whenever a lifecycle interceptor is configured, because the queued path returns before the scope is entered. Its message also asserted that callbacks must not change topology, which is the opposite of this contract. |
| Redundant baseline value accessor dropped | -6 | No production callers remained. |
| Collection scan arm folded into the enumerable one | -16 | Behaviourally identical once guarded on keyed entries, and it double-evaluated that guard. |
| Single-caller seed wrapper inlined | -8 | |

## Documentation corrected

Three statements were false as written: when the Registry sees a subject relative to the descent, the derived ownership contract that this branch changed, and the absence of any enforced bound on reentrancy. Termination is now stated as the handler author's responsibility rather than implied by a guard that no longer exists.

## Verification

- Release build clean, 0 warnings, 0 errors.
- Full non-integration suite: 4,370 passed, 0 failed, 0 skipped, across 34 assemblies.
- The two orphaned-subject cases and the three constructor displacement cases were each confirmed to fail before their fix and pass after.
- Integration suites were not run locally and are left to CI.

## Open before this is finalized

Benchmarks. The only measurement that exists froze production 44 commits before #527's head and was deleted from that branch. It recorded structural writes 17.6 to 32.5 percent slower than #494 with a flat control row. Nothing has measured the current tree, and the one favourable row in that run predates the routing change that touches the same path. That measurement, not this review, is the remaining gate.


RicoSuter:
## Merge master before finalizing this stack

CI on the first push failed `BrowserStartupTests.WhenBrowserOpensBeforeRootLoadingCompletes_ThenRootPropertiesAppearAfterLoading`. That turned out to be a true positive about #494 rather than a problem with this PR, so recording it here.

Master's `RootManager` defers hosted service startup while the root configuration deserializes:

```csharp
using var startup = _context.DeferHostedServiceStartup();
```

That arrived with #532, which landed after the master merge in this stack's base. So #494 is three commits behind master and has neither master's deferral nor the equivalent #527 had built for itself. Without it a hosted service starts before its subject is configured, the layout never leaves its loading gate, and the page reports an unhandled error.

Confirmed by running the same test both ways locally:

| Tree | Result |
|---|---|
| This branch | fails, reproduces CI |
| This branch with `origin/master` merged | passes |

The test is dropped here because `DeferHostedServiceStartup` does not exist on this base, so the behaviour it guards is not this PR's to provide. It should come back once the stack has master merged.

### What the merge costs

`git merge origin/master` is clean, no conflicts. Three test call sites then fail to compile, because master's newer tests use `IInterceptorSubject.Context`, which this rewrite removes:

- `ConfigurableSubjectStartupTests.cs:91` and `SubscriptionTransactionOrderingTests.cs:31` are mechanical, `.Context` becomes `.GetContext()`.
- `ConfigurableSubjectStartupTests.cs:155` is not mechanical. `ConfigurableStartupSubject`'s constructor calls `AddFallbackContext`, and this rewrite removes fallback contexts entirely. The model needs rewriting onto single-context attachment. Using `AttachToContext(context, SubjectAttachmentAnchorKind.Provisional)` compiles and the suite passes, but whether that is the right shape for what the test intends is a call for whoever owns #532's test.

With those three adapted, the full build is clean and the browser startup test passes.

RicoSuter:
## Benchmarks: the open gate is now measured

Two independent runs, #538 (`90cfa48f4`) against this PR's base `cb911c728`, BenchmarkDotNet `LaunchCount 3`, `MemoryRandomization` on. CPU pinned and verified at the operating system rather than from the BenchmarkDotNet header: `performance` governor, `scaling_min_freq` == `scaling_max_freq` == 3.6 GHz on all 16 cores, `intel_pstate/no_turbo` = 1. Machine otherwise idle.

| Row | base | #538 | run 1 | run 2 | alloc |
|---|---:|---:|---:|---:|---:|
| `ServiceOrderResolver.LinearChain` (control) | 1,754 ns | 1,775 ns | +1.2% | -0.2% | 0% |
| `SetScalarAttached` | 163 ns | 180 ns | +10.1% | +7.1% | 0% |
| `ReplaceSingleChildReference` | 2,638 ns | 3,267 ns | +23.8% | +23.9% | 0% |
| `ReplaceCollectionUniqueChildren` | 8,268 ns | 12,128 ns | +46.7% | +47.4% | -1.7% |
| `ReplaceCollectionDuplicateChildren` | 5,128 ns | 7,570 ns | +47.6% | +46.9% | -4.3% |
| `AttachAndReleaseSubtree` | 27,542 ns | 41,626 ns | +51.1% | +52.0% | -1.8% |
| `AddLotsOfPreviousCars` | 47.4 ms | 54.6 ms | +15.3% | +16.9% | +2.4% |

The control row is the noise reference and sits within 1.4 percentage points across both runs, while the structural rows reproduce to under 1 point. The deltas are real, and allocations are flat to slightly better everywhere except `AddLotsOfPreviousCars`.

### Two conclusions

**Structural writes cost 24 to 52 percent more.** That is the price of the callback support contract itself: capture at commit, the installed occurrence journal, and queued notification delivery. It is not something this PR's trimming introduced, and reverting the property edge journal to its original O(1) form already removed a separate quadratic regression that the first draft of this branch had.

**The scalar write path regressed too, and that one looks recoverable.** `SetPropertyValue` now performs an interface dispatched `IReadOnlyDictionary.TryGetValue` plus a record struct copy where it previously read a static field, inlined into every generated setter. Generated setters always pass the exact declared property type, so only `TypedPropertyWriteFactory` and `PropertyReferenceExtensions` can narrow the generic argument. Gating the lookup to those two call sites keeps the routing fix and should return this row to parity. Not done here, since it is performance rather than correctness, but it is cheap and verifiable against two rows.

Earlier support measurements recorded on the predecessor branch are superseded: their production baseline was 44 commits stale, and the one row they showed improving is the row that regressed.

RicoSuter:
## Correction: the metadata lookup is not the cause of the scalar regression

My previous comment attributed the `SetScalarAttached` regression to the per-write `IReadOnlyDictionary.TryGetValue` in `InterceptorExecutor.SetPropertyValue`, and said gating it to the narrowing call sites should return the row to parity. I measured that directly and it is wrong.

Isolation run: this branch with **only** that hunk removed, against this branch, same filter, `LaunchCount 3`, same pinned machine.

| Row | lookup removed | branch as-is | delta |
|---|---:|---:|---:|
| `ServiceOrderResolver.LinearChain` (control) | 1,780.1 ns | 1,828.7 ns | -2.7% |
| `SetScalarAttached` | 176.9 ns | 179.2 ns | **-1.3%** |
| `ReplaceSingleChildReference` | 3,311.1 ns | 3,299.6 ns | +0.3% |

The lookup is worth about one percent, and the control row moved further than that in the same run, so the measurement cannot even separate it from noise. The mechanism was plausible from reading the code, and it was wrong.

Two consequences:

**No public API change is needed.** The proposed new `IInterceptorExecutor` member existed only to let generated setters skip this lookup. It would buy roughly nothing, so it is dropped.

**The scalar regression is still unexplained.** `SetScalarAttached` is genuinely 7 to 10 percent slower than the base across two runs, and the cause is somewhere else in the write path. Finding it needs a profile rather than another guess.

Issue #545, opened 2026-09-14
chore: adopt Sonar analysis across the whole solution


RicoSuter:
Enable SonarAnalyzer.CSharp across every project in `src/Namotion.Interceptor.slnx`, including tests and HomeBlaze, through a sequence of reviewable PRs. The current inventory contains 105 projects, including 35 dedicated `.Tests` projects; the Benchmark project also references test tooling. There are 63 projects under HomeBlaze.

## Shared rule policy

Pin `SonarAnalyzer.CSharp` to `10.33.0.1635`. Keep its default-enabled rules and explicitly enable these additional maintainability rules as warnings in `src/.editorconfig`, using the analyzer's default thresholds:

| Rule | Concern | Default threshold |
| --- | --- | --- |
| S3776 | Cognitive complexity | 15 per method, 3 per property |
| S138 | Method length | 80 lines |
| S134 | Nested control flow | 3 levels |
| S1067 | Expression complexity | 3 conditional operators |
| S107 | Parameter count | 7 |
| S1541 | Cyclomatic complexity | 10 |
| S1200 | Class coupling | 30 types |
| S104 | File length | 1,000 lines |

Disable S2094 globally because intentional empty marker and fixture classes are common. Also disable S3267, S2743, S2696, and S2326 for this repository's intentional performance and state-ownership patterns; their rationale and portability considerations are recorded below. Keep S108 (empty blocks) and S2699 (tests without recognized assertions). Review their actual locations: use meaningful assertions, document intentional empty blocks, and suppress only justified cases at a narrow scope. Add an emitted `AssertionMethodAttribute` in the shared testing library and annotate `AsyncTestHelpers.WaitUntilAsync` and `SourceStateRecorder.WaitForStatesAsync` so genuine assertion helpers are recognized across assemblies. Do not add fake assertions.

The existing warnings-as-errors policy remains in effect. Explicit warning severities therefore still fail the normal build. Shared configuration applies wherever the analyzer is installed; it does not activate Sonar in a project by itself. S2326 is disabled globally to preserve intentional type-identity and per-type cache patterns, and S1135 is disabled provisionally to allow inline TODOs. Tests additionally disable S1144 and S2325 under `.Tests` directories to reduce fixture noise; both remain enabled in production. Other exceptions require a targeted justification. HomeBlaze is included in the final scope.

## Delivery and follow-ups

[...]

PR #587, opened 2026-09-16
fix: preserve transaction replay outcomes

RicoSuter:
Transaction commit now distinguishes vetoed or suppressed writes from successful application and remembers assignments even when a later hook or observer throws. Rollback and BestEffort can therefore compensate the correct mutations and report failed compensation.

Related to #545 (the two transaction replay P1 findings) and #369 (the broader origin redesign remains separate).

## Contract

- Check the whole pending batch before source writes or local replay. Unsupported properties fail with the subject/property identified, retaining pending changes for retry.
- Revalidate setter metadata during replay after awaited source work. Equality acceptance is a successful no-op.
- Compensate recorded mutations in reverse order according to the failure policy. Compensation can fail; delivered notifications and arbitrary callback side effects cannot be undone.

## Breaking changes

Commit requires outcome-aware replay. Regenerate supported subjects and their generated ancestors. Nonvirtual partial setters on compatible root/derived subjects work, including separately compiled ancestors and unmarked intermediate classes. Virtual/override, derived `new`, manual/dynamic and older-generated setters require a compatible `ISubjectPropertyReplay` implementation. Custom replay implementations retain control; helper-name collisions cannot authorize automatic replay.

Core adds `ISubjectPropertyReplay`, `PropertyReplayOutcome`, an outcome-aware executor overload and `PropertyWriteContext.IsEqualityAccepted`. Custom interceptors must preserve truthful write status, including when throwing. Generated classes add protected replay dispatch helpers and a root executor accessor; the API snapshots include these additions.

## Performance

Ordinary generated setters are unchanged. Final ARM64/.NET 9 machine-code comparison matches master for the sampled setter, ordinary executor and transaction/change/lifecycle handlers. The equality handler adds two executed instructions to record acceptance on the equal-value path. Commit adds capability checks and mutation bookkeeping; larger batches reuse pooled storage.

Final Release allocation probes, with matching full-optimization settings and warmed pools:

| Commit size | Master | This branch | Delta |
|---|---:|---:|---:|
| 1 change | 352 B | 352 B | 0 B |
| 50 changes | 3,880 B | 3,880 B | 0 B |
| 512 changes | 37,144 B | 37,144 B | 0 B |

Earlier timing runs at the B2 checkpoint (`ecfa417d`, before the inheritance extension) were inconclusive on this unpinned M4 Max. They are not final-head timings; final verification used allocation and machine-code comparisons.

| Benchmark at B2 checkpoint | Master | B2 | Delta | Allocated |
|---|---:|---:|---:|---|
| Single commit, first pair | 437.304 ns | 461.699 ns | +5.58% | 352 B both |
| Unchanged noise reference, first pair | 2.575 ns | 2.900 ns | +12.63% | 0 B both |
| Single commit, reverse-order repeat | 461.361 ns | 471.587 ns | +2.22% | 352 B both |
| Unchanged noise reference, repeat | 2.797 ns | 2.858 ns | +2.18% | 0 B both |

These measurements do not establish zero timing overhead. Source-value transformation/origin redesign remains outside this fix.

## Diff composition

|---|---:|---:|---:|---:|
| Namotion.Interceptor | 4 | 64 | 2 | +62 |
| Namotion.Interceptor.Generator | 6 | 224 | 14 | +210 |
| Namotion.Interceptor.Tracking | 4 | 122 | 75 | +47 |
| Tests and benchmarks | 28 | 2'685 | 32 | +2'653 |
| Documentation | 1 | 39 | 36 | +3 |
| **Total** | **43** | **3'134** | **159** | **+2'975** |

Production: net +319 lines across 14 files. Most remaining lines are tests and generated/API snapshots.

## Verification

- [x] Documentation updated; independent scoped and whole-branch reviews completed
- [x] Unit tests: 4,110 local cases across 34 suites; no compiler/analyzer warnings or errors. The complete CI Release build and unit suite also passed.
- [x] Integration tests: CI OPC UA, MQTT, WebSocket and HomeBlaze jobs passed.
- [x] Benchmarks: final allocation probes and matching-workload machine-code comparison; earlier timing evidence qualified above
- ~~[ ] Load tests~~ No connector implementation changes
- ~~[ ] Chaos tests~~ No connector implementation changes

[Passing CI run](https://github.com/RicoSuter/Namotion.Interceptor/actions/runs/35159800373).


RicoSuter:
Review note: Instead of adding all those fixes and complexity, we can maybe also just drop the "cancel" support in the On* hooks instead

RicoSuter:
Superseded by #592, which fixes the same two defects with a different mechanism.

The approach here generates a per-class replay path keyed by property name, so it has to enumerate which C# property shapes it can serve. Two shapes that work on master stop working: properties added at runtime through `RegisteredSubject.AddProperty` (dynamic attributes, OPC UA browsed nodes, HomeBlaze method metadata), and `virtual`/`override` partial properties. Both were reproduced against this branch and against master. A runtime-added property cannot be named in a compile-time switch, so that one is not fixable within the design, and the `ISubjectPropertyReplay` escape hatch does not help because implementing any member suppresses generation for the whole class.

#592 reads the outcome from the write pipeline instead, which already computes it, so every property shape keeps working. It is 14 files and +1'055 lines against 43 files and +3'134 here, production net +194 against +319, no generator change, no regeneration requirement and no breaking change.

What is kept from this PR: the defect analysis, the compensation rework in `SubjectPropertyChangeOperations` including the three-state outcome encoding, and `SubjectTransactionReplayOutcomeTests` as the behaviour spec.

Leaving this open until #592 merges, then closing it as superseded.

